PR #31, opened 2026-05-20
fix: stop background flashing in low-memory mode

saeedkolivand:
## Summary

Selecting **Low Memory** performance mode caused the cinematic background to flash rapidly. Two root causes, two fixes.

**Root cause:** `animation-duration: 0.01ms !important` doesn't stop a CSS keyframe animation — it just makes it complete and loop in microseconds. The aurora ribbons, nebulae, and streak animations all looped thousands of times per second, producing a visible flash.

**Fix 1 — `globals.css`:** Replace `animation-duration`/`transition-duration` overrides with `animation: none !important` and `transition: none !important`. This removes keyframes entirely instead of forcing a near-zero looping duration.

**Fix 2 — `CinematicBackground.tsx`:** Return `null` when `performanceMode === 'low-memory'`. The body is already a flat `#07060f` in that mode, so all aurora/nebulae/blob/orb/grain GPU layers are pure waste. The RAF cursor-blob loop is also skipped inside `useEffect`. All hooks are called unconditionally before the null return to satisfy `rules-of-hooks`.

## Test plan

- [ ] Settings → Performance → Low Memory: background stops animating, no flash
- [ ] Settings → Performance → Balanced / Performance: cinematic background renders normally
- [ ] Switching modes live (Low Memory → Balanced) restores the background
- [ ] `pnpm typecheck && pnpm lint` pass

### DIFF
--- a/.cursor/rules/architecture.mdc
+++ b/.cursor/rules/architecture.mdc
@@ -6,6 +6,10 @@ alwaysApply: true
 
 # Architecture Rules
 
+## Shell — always prefix commands with `rtk`
+`rtk pnpm build` · `rtk git status` · `rtk rg foo` — 60-90% token savings.
+Meta: `rtk gain` · `rtk discover`. Always Bash, never PowerShell.
+
 ## Monorepo packages
 - `packages/shared` — IPC contracts + types. No React, no Node APIs.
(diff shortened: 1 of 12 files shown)


saeedkolivand:
🎉 This is included in version 1.3.1.

PR #66, opened 2026-05-21
fix: scraping — cancel, sidecar errors, linkedin session, disconnect, auto-launch

saeedkolivand:
## Summary

**Cancel button does nothing** — reset scraping state synchronously before calling sidecar; don't wait for `job.cancelled` event that may never arrive.

**Sidecar not running → spinner forever** — check `res.error` from `scrape_board` and stop the spinner + show error toast immediately.

**"LinkedIn session not found" after connecting** — `login.ts` never called `context.storageState()` before closing the browser. The HTTP scraper looks for `state.json` to get cookies. Fixed by exporting storage state after successful auth.

**Disconnect shows still connected** — `rmSync` on the Playwright profile directory silently fails on Windows due to file locks. Now writes explicit `{ connected: false }` to `auth-status.json` + removes only `state.json`. Also added optimistic UI update so the status flips instantly on confirm.

**Sidecar auto-launch (dev + production)**:
- Dev: `sidecar.rs` runs `npx tsx src/index.ts` — changes take effect on Tauri restart, no manual build step needed
- Production: `build:binary` script bundles via `ncc` + `@yao-pkg/pkg` into a platform binary; CI injects binary path into `bundle.resources` before `tauri-action`
- `externalBin` removed — it requires the binary at cargo build time which breaks local dev

## Test plan

- [ ] `pnpm dev` → sidecar starts automatically, scraping works
- [ ] Cancel during scrape → spinner stops immediately  
- [ ] Start scrape with sidecar not running → error toast shown
- [ ] Connect LinkedIn → scrape → no "session not found" error
- [ ] Disconnect LinkedIn → status flips to disconnected instantly


saeedkolivand:

PR #70, opened 2026-05-21
fix: disconnect boards reliably and show toast notifications

saeedkolivand:
## Summary

- **Toast not showing**: `z-[var(--z-toast)]` is a Tailwind class defined only in `packages/ui/src/components/Toast.tsx`. When the UI package is compiled to `dist/` before Tailwind scans it, the class is never generated — the toast container gets `z-index: auto` and sits behind modal overlays. Fixed by replacing with `style={{ zIndex: 700 }}` inline.

- **Disconnect doesn't persist**: `sidecar_board_disconnect` only called the sidecar HTTP endpoint. If the sidecar wasn't running, the disconnect was a silent no-op and boards re-appeared connected on next launch. Now also writes `auth-status.json` directly to disk (`~/.ajh/browser-state/<boardId>/auth-status.json`) before reaching the sidecar — matching what the sidecar's own `disconnect()` does.

- **Sign out all incomplete**: `privacy_sign_out_all` only cleared PostingsCache/InteractionStore. Now iterates all known boards (linkedin, indeed, xing, glassdoor) and runs the full disconnect flow for each.

## Test plan

- [ ] Open settings → board shows Connected → click Disconnect → confirm → toast appears + button switches to Connect
- [ ] Restart the app → board stays Disconnected
- [ ] Privacy tab → Sign out all → toast appears → all boards show Disconnected
- [ ] Verify toast appears for other actions (AI generation, support actions, etc.)
- [ ] `cargo check` passes ✓

### DIFF
--- a/apps/tauri/src-tauri/src/commands.rs
+++ b/apps/tauri/src-tauri/src/commands.rs
@@ -1506,14 +1506,52 @@ async fn sidecar_board_status(app: &AppHandle, board_id: &str) -> Value {
 }
 
 async fn sidecar_board_disconnect(app: &AppHandle, board_id: &str) -> Value {
-    let Some(port) = sidecar_port(app) else {
-        return json!(null);
-    };
-    let job_id = uuid_v4();
-    let cmd = json!({ "kind": "board.disconnect", "boardId": board_id });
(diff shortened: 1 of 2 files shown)


saeedkolivand:

PR #74, opened 2026-05-22
fix: add tauri signing config and fallback for unsigned builds

saeedkolivand:
Add Tauri signing configuration to enable .sig file generation for secure auto-updates. Also adds fallback support in the release workflow to work without signatures for unsigned builds.


saeedkolivand:

PR #75, opened 2026-05-22
fix: remove invalid signing property from tauri bundle config

saeedkolivand:
Remove the 'signing' property from bundle config as it's not allowed in Tauri v2 schema. This fixes CI build failures on both macOS and Windows.


saeedkolivand:

PR #76, opened 2026-05-22
fix: pass updater manifest env to node

saeedkolivand:
## Summary
- pass inline environment variables before Node invocations in the updater manifest release job
- fixes latest.json generation failing when process.env.PLATFORMS is undefined

## Verification
- rtk git diff --check


saeedkolivand:

PR #77, opened 2026-05-22
fix: launch bundled scraper sidecar in release

saeedkolivand:
## Summary
- resolve the bundled scraper-runtime path from Tauri resources/binaries in release builds
- handle Windows .exe sidecars and macOS universal builds by trying the runtime architecture sidecar
- retry sidecar startup briefly before scrape/login/apply/autopilot commands return not-ready errors
- clear sidecar running state if the process exits before or after readiness

## Verification
- rtk cargo check
- rtk git diff --check -- apps/tauri/src-tauri/src/sidecar.rs apps/tauri/src-tauri/src/commands.rs
- rtk git diff --cached --check -- apps/tauri/src-tauri/src/sidecar.rs apps/tauri/src-tauri/src/commands.rs

### DIFF
--- a/.codexrules
+++ b/.codexrules
@@ -1,0 +1,138 @@
+# OpenAI Codex — Usage Rules
+
+OpenAI Codex API integration for code generation and completion. Follow these rules for safe, efficient, and cost-effective usage.
+
+---
+
+## API Key Management
+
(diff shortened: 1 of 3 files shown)


saeedkolivand:

PR #78, opened 2026-05-22
fix: updater signature field and dropdown styling improvements

saeedkolivand:
This PR fixes two issues:

1. **Updater signature field error**: Fixed the "missing signature" error when checking for updates on Windows and macOS. The release workflow now always includes the `signature` field in `latest.json` (empty string when signature files are missing).

2. **SelectDropdown styling**: Made dropdown background darker and more glassy, with support for low-memory performance mode:
   - Changed background from `bg-secondary` to `bg-black/60`
   - Increased backdrop-blur from `backdrop-blur-md` to `backdrop-blur-xl`
   - Added CSS rule to reduce blur to 8px when Low Memory mode is selected in performance settings


saeedkolivand:

PR #80, opened 2026-05-22
fix: remove obsolete scraper-runtime build steps from release workflow

saeedkolivand:
Removes sidecar scraper build steps from Windows and macOS builds in release orchestration workflow, as the scraper-runtime app has been removed and replaced with in-process Rust scrapers.


saeedkolivand:

PR #81, opened 2026-05-22
fix: strip newlines from minisign signatures in latest.json generation

saeedkolivand:
Fixes 'invalid encoding in minisign data' error in the updater by stripping newlines from .sig signature files before embedding them in latest.json. The signature files contain newlines that break JSON when embedded directly.


saeedkolivand:

PR #94, opened 2026-05-23
fix: use gh_token for semantic-release api calls

saeedkolivand:
## Changes

- Use `GH_TOKEN` secret for semantic-release API calls instead of `GITHUB_TOKEN`
- Add `id-token: write` permission to workflow for better authentication

## Issues Fixed

- Semantic-release failed with HTTP 401 "Bad credentials" when trying to create releases
- `GITHUB_TOKEN` has insufficient permissions for release creation

## Testing

- Workflow changes validated by GitHub Actions syntax check
- Uses existing `GH_TOKEN` secret which has proper release permissions

### DIFF
--- a/.github/workflows/04-release-orchestration.yml
+++ b/.github/workflows/04-release-orchestration.yml
@@ -9,6 +9,7 @@ permissions:
   contents: write
   issues: write
   pull-requests: write
+  id-token: write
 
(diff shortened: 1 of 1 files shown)


saeedkolivand:
🎉 This is included in version 0.2.3.

PR #97, opened 2026-05-24
Add cancellation and language detection to AI generate page

saeedkolivand:
# Add cancellation and language detection to AI generate page

## Changes
- Added client-side language detection to AI generate page using @ajh/shared/language-detection
- Integrated AbortController for cancellation in AI generate page UI
- Added AbortSignal support to streamGenerate, generateResume, and generateCoverLetter functions
- Added cancellation checks to all AI streaming functions in Rust backend
- Modified streaming loops to check job cancellation status before reading each chunk
- Forces HTTP connection closure when job is cancelled

## Behavior
- UI shows cancellation immediately when reset is pressed during generation
- App stops processing and emitting tokens
- Backend marks job as cancelled
- LLM generation continues in background until provider detects disconnection (API limitation)
- Applies to all providers: Ollama, OpenAI, Anthropic, Gemini

## Related
- Similar to cancellation implementation in analyze page
- Language detection moved to shared package for reuse


saeedkolivand:
🎉 This is included in version 0.2.6.

PR #101, opened 2026-05-24
fix: build platform-specific targets in release workflow

saeedkolivand:
## Summary
Fixes the release workflow build step that was attempting to build macOS targets on Linux and Windows, causing all builds to fail.
 
## Problem
The workflow had separate 'Build Tauri Application (Intel)' and 'Build Tauri Application (Apple Silicon)' steps that used hardcoded macOS targets (--target x86_64-apple-darwin) on all platforms. This caused Linux and Windows builds to fail with 'target x86_64-apple-darwin is not installed'.
 
## Solution
Combined the build steps into a single step that:
- On macOS: builds both Intel and Apple Silicon targets
- On Linux/Windows: builds for the current platform (no --target flag)
 
## Changes
- Combined Intel and ARM build steps into one conditional build step
- Updated failure check to use new combined step ID
- Updated job summary to use new step ID
- Removed ARM64 target addition (now handled in build step)


saeedkolivand:
🎉 This is included in version 0.2.15.

PR #102, opened 2026-05-24
fix: add shell bash to windows build step

saeedkolivand:
## Summary
Fixes Windows build failure by explicitly setting shell to bash for the build step.

## Problem
The Windows runner was using PowerShell to execute bash syntax (\if [ ... ]\), causing a parser error:
\`
ParserError: Missing '(' after 'if' in if statement.
\`

## Solution
Added \shell: bash\ to the build step to ensure bash is used on all platforms, including Windows.

## Changes
- Added \shell: bash\ to the Build Tauri Application step
- This ensures bash syntax works correctly on Windows runners

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,7 +173,7 @@ jobs:
           cache: pnpm
 
       - name: 💾 Restore Turbo Cache
-        uses: actions/cache/restore@v4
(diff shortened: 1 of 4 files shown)


saeedkolivand:
🎉 This is included in version 0.2.16.

PR #107, opened 2026-05-25
fix: enable auto-update detection with createUpdaterArtifacts

saeedkolivand:
## Problem
The auto-updater was not detecting new releases even when a newer version was available on GitHub. Users checking for updates would see "latest version" regardless of actual release status.

## Root Cause
The `createUpdaterArtifacts` configuration was missing from \[tauri.conf.json\](cci:9://file:///apps/tauri/src-tauri/tauri.conf.json:0:0-0:0). This setting is required for Tauri v2 to automatically generate:
- `.sig` signature files for each platform installer
- Proper update metadata for version comparison

Without this, the build process didn't create the required update artifacts, so the updater couldn't detect new releases.

### 1. apps/tauri/src-tauri/tauri.conf.json
- Added `"createUpdaterArtifacts": true` to the bundle section
- Enables automatic generation of signature files during build

### 2. .github/workflows/release.yml
- Added step to generate `latest.json` after build
- Collects signatures from generated `.sig` files
- Constructs proper JSON with version info, URLs, and signatures for all platforms
- Updated artifact upload paths to explicitly include `.sig` files:
  - Windows: nsis/*.sig, msi/*.sig
  - macOS: x86_64-apple-darwin/dmg/*.sig, aarch64-apple-darwin/dmg/*.sig
  - Linux: appimage/*.sig

### 3. docs/UPDATER_SETUP.md
- Updated documentation to reflect the new automated workflow
- Removed manual signing steps (now handled by createUpdaterArtifacts)
- Added troubleshooting section for common errors
- Added GitHub secrets configuration instructions

## Testing
After this merge, the next release will automatically generate proper update artifacts. Users will be able to detect and install updates correctly via Settings → Check for Updates.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -297,20 +297,96 @@ jobs:
           fi
           echo "::endgroup::"
 
-      - name: 📤 Upload Release Assets
+      - name: � Generate latest.json
+        if: success()
+        shell: bash
+        run: |
+          echo "::group::Generating latest.json"
+          cd apps/tauri/src-tauri/target/release/bundle
+
+          # Collect signature files and URLs
+          VERSION="${{ needs.sync-version-files.outputs.version }}"
+          PUB_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
+
+          # Start JSON structure
+          cat > latest.json << EOF
+          {
+            "version": "$VERSION",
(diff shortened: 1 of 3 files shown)


saeedkolivand:
🎉 This is included in version 0.5.1.

PR #110, opened 2026-05-25
fix: enable tauri devtools feature for release builds

saeedkolivand:
## Summary

- Adds `"devtools"` to the Tauri feature list in `Cargo.toml`
- Without this feature, `WebviewWindow::open_devtools()` does not exist as a method in release builds, causing a compile error on macOS (`no method named open_devtools found`)
- The Open DevTools button in Settings → Developer now works in both debug and release builds

## Root cause

`open_devtools()` is gated behind the `devtools` Cargo feature in Tauri v2. Debug builds implicitly enable it; release builds do not unless it is explicitly listed in features.

### DIFF
--- a/apps/tauri/src-tauri/Cargo.toml
+++ b/apps/tauri/src-tauri/Cargo.toml
@@ -19,7 +19,7 @@ strip = true
 tauri-build = { version = "2", features = [] }
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


saeedkolivand:
🎉 This is included in version 0.7.1.

PR #124, opened 2026-05-27
chore(deps): batch dep upgrades — keyring v4, parking_lot, tauri-plugin-log, rusqlite/scraper/lopdf bumps

saeedkolivand:
## Summary

- **rusqlite 0.39→0.40, scraper 0.26→0.27, lopdf 0.34→0.40**: Bump three crates to latest minor/major. Fixes lopdf `as_name_str()` → `as_name()` API break in `extraction/pdf.rs`.
- **keyring v3 → keyring-core v1 + platform store crates**: Migrate from monolithic `keyring = "3"` to the v4 split model (`keyring-core` + `windows-native-keyring-store` / `apple-native-keyring-store` / `dbus-secret-service-keyring-store`). Adds `credentials::init_keyring()` call at startup to register the platform backend.
- **parking_lot**: Replace all `std::sync::Mutex` usage (16 files) with `parking_lot::Mutex`. Eliminates 67 `.lock().unwrap()` calls — `parking_lot` guards are infallible so there is no poisoning risk, and the lock never blocks the Tokio executor.
- **tauri-plugin-log**: Register file-based log rotation (5 MB, keep 3 files) at `Warn` level. Wire `support_export_logs` command to return the actual log directory path instead of returning `{ success: false }`. Add `"log:default"` capability. Replace all `eprintln!` with `log::warn!`.

## Notes

- `rusqlite_migration` was evaluated but no published version is compatible with rusqlite 0.40 (`v1.x` uses the removed `NO_PARAMS` constant; `v2.x` requires `rusqlite = "^0.34"` which conflicts). Skipping migration framework — existing `CREATE TABLE IF NOT EXISTS` + manual `pragma_table_info` guard in `documents/mod.rs` continue to handle schema evolution correctly.
- `reqwest`, `generic-array`, `toml` transitive bumps are blocked by `chromiumoxide` and `tauri-plugin-updater` pinning; left as-is.
- `Cargo.lock` is now tracked (removed from `.gitignore`).

## Test plan

- [ ] `cargo build --release` succeeds with zero warnings
- [ ] `cargo test` — 437 tests pass
- [ ] Launch app on Windows: verify credential read/write via Credential Manager still works after keyring v4 migration
- [ ] Verify log file is created at `%APPDATA%\com.ajh.desktop\logs\`
- [ ] Open Settings → Support → Export Logs returns a valid path


saeedkolivand:
🎉 This is included in version 0.16.0.

PR #138, opened 2026-05-28
fix(deps): scope tracing-subscriber to explicit features

saeedkolivand:
## Summary

- Set `default-features = false` on `tracing-subscriber` to avoid pulling in unused default features
- Explicitly list required features: `env-filter`, `fmt`, `std`, `ansi`


saeedkolivand:
Superseded by #139 which removes tracing-subscriber entirely in favour of tracing/log feature bridging to tauri-plugin-log.

PR #146, opened 2026-05-28
fix: repair auto-updater key mismatch and macos release upload

saeedkolivand:
## Summary

Fixes two independent bugs in the release/updater pipeline that left the auto-updater broken for a long time.

**1. Updater verified against the wrong signing key** — the app shipped with public key `CB8D1EEABC860572`, but every release (v0.17.0, v0.18.0, …) is signed with `41C986C44BFBD4F9`, so `download()` always failed with `invalid encoding in minisign data`. The build also overwrote the committed key from a wrong/malformed `TAURI_SIGNING_PUBLIC_KEY` secret on every run, so fixing `tauri.conf.json` alone never stuck.

- Commit the correct public key (`41C986…`) and make `tauri.conf.json` the single source of truth.
- Stop injecting the pubkey from a secret in `sync-tauri-version.cjs` (a public key isn't secret).
- New `scripts/verify-updater-key.cjs` CI guard fails the build if a freshly-signed artifact's key id ≠ the committed key — so this can't silently regress.

**2. v0.19.0 release failed (macOS)** — the macOS leg builds two bundles (Intel + Apple Silicon) that both emit the identical filename `AI-Job-Hunter-Assistant.app.tar.gz`. Uploading two same-named assets to one release races in `softprops/action-gh-release`; v0.18.0 won the race, v0.19.0 lost → macOS job failed → manifest job skipped → no `latest.json`.

- Suffix each macOS artifact + `.sig` with `-intel` / `-apple-silicon`.
- Map each `darwin-*` entry in `latest.json` to its own tarball + signature (also fixes serving the wrong-arch binary to one Mac architecture).

## Verification

- [x] Decoded key ids: committed pubkey now `41C986…`, matches published release signatures.
- [x] `verify-updater-key.cjs` passes against the real v0.18.0 signature; exits non-zero on mismatch.
- [x] `release.yml` parses as valid YAML; macOS rename logic dry-run produces unique `-intel` / `-apple-silicon` names.
- [x] `node --check` + ESLint clean on changed scripts.

## After merge (manual)

- A `fix:` → patch release will build with the corrected key and unique macOS names. Re-running the failed v0.19.0 job won't help (that tag predates this fix).
- Existing installs (≤ v0.19.0) must be **reinstalled once** — their baked-in key can't verify the corrected updates.
- The `TAURI_SIGNING_PUBLIC_KEY` GitHub secret is now unused and can be deleted. Leave `TAURI_SIGNING_PRIVATE_KEY` (+ password) as-is.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,7 +72,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GH_TOKEN: ${{ secrets.GH_TOKEN }}
           HUSKY: 0
-          TAURI_SIGNING_PUBLIC_KEY: ${{ secrets.TAURI_SIGNING_PUBLIC_KEY }}
 
       - name: 📝 Release Summary
         if: always()
@@ -133,8 +132,6 @@ jobs:
           cache: pnpm
 
       - name: 🔖 Sync Version Files
-        env:
-          TAURI_SIGNING_PUBLIC_KEY: ${{ secrets.TAURI_SIGNING_PUBLIC_KEY }}
         run: |
           echo "::group::Syncing tracked version files"
           node scripts/sync-tauri-version.cjs ${{ needs.release.outputs.version }}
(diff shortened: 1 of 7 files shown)


saeedkolivand:
🎉 This is included in version 0.19.1.

PR #147, opened 2026-05-28
feat: autopilot tailoring, scraping reliability, provider routing & UX fixes

saeedkolivand:
## Summary

A batch of reported-bug fixes plus several user-facing features across **Autopilot**, the **resume input**, **job scraping**, and **AI provider** handling.

## Autopilot
- **Edit mode** — the creation wizard now also edits existing autopilots (pre-filled, "Save Changes").
- **Found jobs on the card** — each run persists the jobs it found (with match scores); a "Found N" toggle reveals an expandable, collapsible list.
- **Per-job "Apply"** — an inline modal (no page redirect) to generate a **tailored resume/cover letter from the job's description**. Reuses the AI Generate primitives: model picker, generation progress (analyzing → writing), streaming output, copy, and an export dropdown. Falls back to "open posting" when a description can't be fetched.
- Schedule options reordered (Once a day before Twice a day).

## Resume input
- Extracted the resume picker/uploader into a shared **`components/resume/*`** bundle, reused in AI Generate, Analyze, and the autopilot **Filter step** (so autopilot can select a saved resume, not just paste text).
- Fixed the **saved-resume dropdown** (portal menu vs. outside-click handler dropped clicks), added a per-row **set-as-default** star, and removed the noisy switch toast.

## Scraping reliability
- Fixed the **intermittent "infinite scraping"** state: the job-events listener now subscribes **once** (stable handler) instead of re-subscribing every render (which raced the async Tauri listener and dropped `job.completed`). Added a **JobTracker polling watchdog** and made completion **idempotent** so the UI can never hang.
- **New search query clears** the previously scraped jobs.
- Scraped descriptions are **beautified** (paragraphs/bullets preserved), LinkedIn "Show more/less" stripped, and a job's description can be **fetched on demand** via the URL resolver.

## AI providers
- **Analyze** now routes to the **active provider** (it always hit Ollama before, even on cloud).
- **OpenAI reasoning models (o1/o3/o4)**: omit `temperature`, use `max_completion_tokens` (their endpoints reject the standard shape).
- Provider HTTP errors (**401/429/400/5xx**) mapped to clear, actionable messages instead of raw bodies.

## Other
- **macOS**: re-enabled the cinematic background with WebKit-safe animations (static blur, transform-only keyframes).
- **LocationInput**: free-text entry, not just dropdown results.
- **Onboarding**: Skip action now shown on the cloud AI step.
- **README**: macOS released-build quarantine (`xattr -cr ...`) instructions.
- **chore**: exclude untracked `.claude/` tooling from lint.

## Known follow-up (not in this PR)
- "Route embeddings to cloud" is intentionally **deferred** — it needs a real migration (Ollama 768-dim vs. cloud 1536-dim vectors mean existing indexed docs must be re-embedded, plus per-vector provider tagging). Tracked for a dedicated change.

## Test plan
- [ ] Autopilot: create, edit, run; verify found-jobs list, per-job Apply generates + exports; schedule order
- [ ] Resume input: select saved resume, set default, upload, paste — in AI Generate / Analyze / autopilot Filter
- [ ] Scraping: run repeated/changing searches; confirm state always resolves and old results clear on new query
- [ ] AI providers: cloud generation + analyze hit the right endpoint; trigger a 429 / o-series model and confirm clear errors
- [ ] macOS: background animates; LinkedIn description fetch/beautify
- [ ] Onboarding cloud step shows Skip


saeedkolivand:
🎉 This is included in version 0.20.0.

PR #170, opened 2026-05-29
ci: make readme release badge static and sync it on release

saeedkolivand:
## Why

The README **Release** badge used the dynamic shields.io endpoint:

```
https://img.shields.io/github/v/release/saeedkolivand/ai-job-hunter-assistant-app?color=e24b4a
```

That calls the GitHub API via shields' shared token pool, which intermittently renders **"Unable to select next GitHub token from pool"** instead of the version when the pool is rate-limited. It's the only badge in the README that hits an API.

## What

Make it a **static** badge and keep it accurate automatically — no API call, so it never flakes:

- **`README.md`** — badge is now `img.shields.io/badge/release-v0.23.1-e24b4a`.
- **`scripts/sync-tauri-version.cjs`** — already syncs the release version into `package.json`, `Cargo.toml`, `tauri.conf.json`; now also rewrites the README badge version (regex `release-v<version>-<hex>`, idempotent, warns+skips if not found).
- **`.github/workflows/release.yml`** — the existing `sync-version-files` job adds `README.md` to the diff-check and `git add`, so the badge is bumped and committed to `main` on every release (same `chore: sync version files to vX [skip ci]` commit).

Verified the regex against the current README (matches `v0.23.1`, rewrites cleanly). `ci:` type → no release triggered.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -146,7 +146,8 @@ jobs:
             package.json \
             apps/tauri/package.json \
             apps/tauri/src-tauri/Cargo.toml \
-            apps/tauri/src-tauri/tauri.conf.json; then
+            apps/tauri/src-tauri/tauri.conf.json \
+            README.md; then
             echo "Version files are already synced."
             exit 0
           fi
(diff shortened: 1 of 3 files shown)


saeedkolivand:

PR #175, opened 2026-05-29
refactor: add resume text-to-model adapter (phase 2a)

saeedkolivand:
## Phase 2a — model adapter

The brick the layout engine needs. Introduces `model::adapter::model_from_resume_text`, the strangler-fig bridge that builds a canonical `DocumentModel` from the existing `parser::parse_resume` output — **without touching extraction yet**.

It regroups the flat `Vec<ParsedLine>` into the structured model shape:
- **Header** — name + contact as `RichText` runs (links are first-class, recovered via `tokenize_rich`)
- **Sections** — classified by `SectionId::from_header`, heading text preserved
- **Blocks** — paragraphs, standalone bullets, and **entries** (a title with an optional subtitle/date and its own bullets, e.g. a job role)

### Design notes / decisions
- **Faithful, not enriching.** Mirrors exactly what the parser already recognizes — no re-classification. Bold survives wherever the parser preserved it; links survive everywhere (the parser only strips `**`, never `[label](url)`).
- **Never drops content.** Leading content under no heading becomes an untitled `Summary` section (empty heading → engine renders no title line); stray/out-of-place lines fall back to paragraphs. A dedicated test asserts every source line surfaces in the model.
- **Entries.** A `JobEntry` opens an entry; a following `JobTitle` is its subtitle; following bullets are its bullets; the next entry/paragraph/section closes it. The date is `right_text`; the title uses the parser's left-hand part.
- **Resumes only.** Cover letters have a fundamentally different shape (letterhead/date/recipient/salutation/closing/signature) and keep the legacy `export::pdf` path until modeled explicitly later.

### Why this slicing
Phase 2 (PDF onto a `layout/` engine) is the largest phase. Building the layout *geometry* in isolation risks rework, because its output can't be parity-checked until the PDF actually consumes it. So:
- **2a (this PR):** the adapter — pure, fully testable, zero-regression.
- **2b (next):** the `layout/` engine **+** PDF consumer **together**, behind the `layout_pdf` flag, where golden-snapshot parity is validated where it matters.

This mirrors the Phase 1a → 1b split precedent.

### Risk
**Zero behavior change.** Nothing consumes the adapter yet — covered by the `model` module's existing `#![allow(dead_code)]` (same as the other Phase 1 foundations).

### Gate
- `cargo test --all-features` → **533 passed** (was 524; +9 adapter tests)
- `adapter.rs` rustfmt-clean
- No IPC contract change

### DIFF
--- a/apps/tauri/src-tauri/src/model/adapter.rs
+++ b/apps/tauri/src-tauri/src/model/adapter.rs
@@ -1,0 +1,401 @@
+//! Adapter: build a canonical [`DocumentModel`] from already-parsed resume text.
+//!
+//! Temporary strangler-fig bridge. It reuses the existing line-based
+//! [`parser::parse_resume`](crate::export::parser::parse_resume) so the new model
+//! (and the layout engine that will consume it in Phase 2b) can be populated
+//! WITHOUT yet touching extraction. The flat `Vec<ParsedLine>` is regrouped into
+//! the structured shape the model expects: a header plus titled sections whose
+//! content is paragraphs, standalone bullets, and entries (a title line with an
+//! optional subtitle / date and its own bullets).
+//!
+//! Faithfulness: this mirrors what `parse_resume` already recognizes — it does
+//! not re-classify or enrich. Inline formatting is recovered with
+//! [`tokenize_rich`], so links survive as first-class runs (and bold survives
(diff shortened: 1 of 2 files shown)


saeedkolivand:

PR #176, opened 2026-05-29
refactor: exact pdf links and centering via font metrics (phase 2b)

saeedkolivand:
## Phase 2b — exact links & centering (kills the `× 0.52` hack)

Wires `crate::measure::FontMetrics` (the font-metrics layer from Phase 1b) into the PDF renderer, replacing **every** character-count width estimate with real glyph advances from the bundled fonts. This is the first real consumer of `measure/`.

### What it fixes — for all 9 templates at once
- **Link rects** now sit exactly over the rendered contact-line labels. The old `pt_to_mm(size) * 0.52 * char_count` estimate also **mis-measured multi-byte text** (it counted *bytes* via `.len()`), so accented names / unicode separators (`·`) drifted the clickable rects off the label.
- **Centered names & contact lines** (Executive, Refined Executive) center on the true visible glyph width instead of a `× 0.3` per-char guess.
- **Right-aligned job dates** align to their real width instead of a `× 0.3` estimate that under-measured and left them floating short of the margin.

### Design
- `FontMetrics` is a zero-sized stateless type, so it drops in with **no plumbing** — render functions already know each role's `FontFamily` from the template.
- The link-rect math is extracted into a **pure, injectable** function `contact_link_rects(text, x_start, family, size, &dyn MeasureText) -> Vec<ContactLinkRect>`, unit-tested directly. `ContactSpanCtx` now carries the `FontFamily` so rects use the body face.

### Scope (deliberately bounded — low regression risk)
- **No structural / reading-order change.** Text still renders via `ShowText` (PDF positions glyphs); only the x-coordinates of centering and the invisible link rects change — strictly more accurate.
- `wrap_segments`' line-break heuristic is **left as-is** (changing it alters line breaks → belongs to the layout-engine phase).
- The cover-letter inline date estimate in `pdf/mod.rs` stays until cover letters are modeled.

### Why no feature flag
This is a strict accuracy improvement to coordinates only (not a new layout path), so it ships directly rather than behind `layout_pdf`. The big structural engine (true multi-page two-column) is the next PR and *will* be flag-gated with golden snapshots.

### Tests
- **5 unit tests** on the extracted geometry: centering math, advance scaling, exact link-rect positions (LinkedIn + `mailto:`), empty-when-no-links, and a regression-direction test that the exact advance differs from the legacy char-count guess.
- **2 end-to-end PDF tests**: one parses the generated PDF with `lopdf` and asserts the `https://linkedin.com/in/jane` and `mailto:jane@example.com` annotations are present; one exercises the centered-template (Executive) path.

### Gate
- `cargo test --all-features` → **540 passed** (was 533; +7)
- edited files rustfmt-clean · no IPC contract change

### DIFF
--- a/apps/tauri/src-tauri/src/export/pdf/test.rs
+++ b/apps/tauri/src-tauri/src/export/pdf/test.rs
@@ -136,7 +136,8 @@ fn test_generate_pdf_resume_with_section_markers() {
 
 #[test]
 fn test_generate_pdf_cover_letter_with_section_markers() {
-    let text = "Some header\n### COMPLETE COVER LETTER ###\nDear Hiring Manager,\n\nI am writing...";
+    let text =
+        "Some header\n### COMPLETE COVER LETTER ###\nDear Hiring Manager,\n\nI am writing...";
     let request = ExportRequest {
         text: text.to_string(),
         format: super::super::types::ExportFormat::Pdf,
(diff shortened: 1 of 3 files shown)


saeedkolivand:

PR #177, opened 2026-05-29
refactor: add single-column layout engine (phase 2c)

saeedkolivand:
## Phase 2c — the layout engine

Introduces `layout/` — the format-agnostic layout engine the spec calls for. `layout_document(model, template, geom, &dyn MeasureText)` turns a `DocumentModel` into a backend-agnostic **`LaidOutDoc`** display list:

```
LaidOutDoc { page_width_mm, page_height_mm, pages: [
  Page { fills, rules, texts: [PlacedText], links: [LinkRect] }
] }
```

Coordinates are millimetres from each page's **top-left**. Backends (PDF now, DOCX later) translate this list and do **no layout themselves**.

### What it does
- **Real-metrics geometry everywhere** (via `measure/`, no char-count guesses): greedy word **wrapping** (`wrap_runs`), name/contact **centering**, right-aligned entry **dates**, and **link rects** placed directly from the model's `RichText` link runs — `header.contact` already carries links as first-class runs, so there's no regex re-parsing.
- **Pagination**: vertical rhythm from the template's point sizes + line spacing, page breaks at the bottom margin, and a section-heading **orphan guard** (keeps a heading with ≥2 body lines).
- **Blocks**: header (name + contact + rule), sections (caps/rule headings), paragraphs, standalone bullets, entries (title + right-aligned date + subtitle + bullets).

### Scope (single-column for now)
True multi-page **two-column** (independent column cursors + per-page sidebar band) is layered on next — `LaidOutDoc` already has `FillRect` for the band. **Wiring the PDF backend** onto this (behind a `layout_pdf` flag, with golden-snapshot parity) follows two-column.

### Risk
**Zero behavior change** — nothing consumes the engine yet (module `#![allow(dead_code)]`, like the other foundations). Pure geometry → verified entirely by unit tests.

### Tests (13)
Driven with the **real bundled-font metrics** so geometry is exact: wrapping stays within width, Classic (left) vs Executive (centered) name alignment, contact link rects + URLs, link rects in-page, all-caps headings, right-aligned dates, multi-page pagination, a no-content-dropped guard, and all 9 templates lay out without panicking.

### Gate
- `cargo test --all-features` → **552 passed** (was 540; +12)
- `cargo clippy --all-targets --all-features -- -D warnings` → clean (bundled `place_line` args into a `LineStyle` struct rather than allowing `too_many_arguments`)
- new files rustfmt-clean · no IPC change · full local pre-push gate green

### DIFF
--- a/apps/tauri/src-tauri/src/layout/mod.rs
+++ b/apps/tauri/src-tauri/src/layout/mod.rs
@@ -1,0 +1,571 @@
+//! Layout engine — turns a [`DocumentModel`] into a backend-agnostic
+//! [`LaidOutDoc`] *display list*: pages of positioned text, link rectangles,
+//! background fills, and rules, all in millimetres from each page's top-left.
+//!
+//! This is the format-agnostic core the spec calls for: the PDF backend (and,
+//! later, DOCX) translates a `LaidOutDoc` to its own primitives and does **no**
+//! layout itself. All horizontal geometry — wrapping, centering, right-aligned
+//! dates, link-rect placement — uses real glyph advances from [`MeasureText`]
+//! (the `measure/` layer), never a character-count estimate.
+//!
+//! Phase 2c introduces the engine for the **single-column** path with no
+//! consumer yet (additive, behind the module's `dead_code` allow, like the other
+//! migration foundations). True multi-page two-column layout is layered on in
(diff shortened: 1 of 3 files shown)


saeedkolivand:

PR #178, opened 2026-05-29
refactor: true multi-page two-column layout in the engine (phase 2d)

saeedkolivand:
## Phase 2d — true multi-page two-column

Extends the layout engine to two-column documents, fixing the legacy renderer's **single-page-only** two-column limitation.

### How it works
The single cursor is refactored into a reusable **`Flow`** (a paginating column):
- **Single-column** = one full-width flow (unchanged behavior — same `Flow`, full-width frame).
- **Two-column** = header laid full width on page 1, then an **independent sidebar flow + main flow**, merged page-for-page with the **sidebar background band painted behind each page's columns**.

Because each column paginates on its own and pages are merged page-for-page, two-column resumes now flow across **as many pages as needed** — page N still carries the band even if one column has run out of content.

### Decisions
- **Section → column** uses the canonical **`theme::placement_for`** (skills / education / languages / certifications → sidebar), not heading string-matching.
- **Header + contact link rects** stay full-width on page 1.
- Sidebar/main geometry mirrors the legacy split (ratio + gap), with a little padding inside the band.

### Risk
**Zero behavior change** — still no consumer (module `#![allow(dead_code)]`). Single-column output is byte-for-byte the same path (one full-width flow); the existing single-column tests still pass.

### Tests (+4, 17 total in `layout`)
Column splitting (EXPERIENCE → main, SKILLS → sidebar), sidebar band at the left margin, full-width header + links on page 1, and a band on **every** page when the main column paginates.

### Gate
- `cargo test --all-features` → **556 passed** (was 552; +4)
- files rustfmt-clean · no IPC change · full local pre-push gate green

### Next
**Phase 2e:** wire the PDF backend onto `LaidOutDoc` behind a `layout_pdf` feature flag, with golden-snapshot parity per template before flipping the default. That's where the engine finally renders real output.

### DIFF
--- a/apps/tauri/src-tauri/src/layout/mod.rs
+++ b/apps/tauri/src-tauri/src/layout/mod.rs
@@ -8,19 +8,25 @@
 //! dates, link-rect placement — uses real glyph advances from [`MeasureText`]
 //! (the `measure/` layer), never a character-count estimate.
 //!
-//! Phase 2c introduces the engine for the **single-column** path with no
-//! consumer yet (additive, behind the module's `dead_code` allow, like the other
-//! migration foundations). True multi-page two-column layout is layered on in
-//! the next phase; the PDF backend is wired onto this — behind a feature flag,
-//! with golden-snapshot parity — after that.
+//! Both single-column and **true multi-page two-column** layouts are built from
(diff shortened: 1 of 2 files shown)


saeedkolivand:

PR #179, opened 2026-05-29
refactor: render resumes through the layout engine behind a feature flag (phase 2e)

saeedkolivand:
## Phase 2e — the payoff: PDF backend on the layout engine

The engine finally renders real output. `export::layout_pdf` builds a `DocumentModel` from the resume text → lays it out with `layout_document` into a `LaidOutDoc` → translates that display list to `printpdf` ops (top-left mm → printpdf's bottom-left), reusing `pdf_renderer`'s font loading / colors / line helpers. **It does no layout itself** — wrapping, centering, columns, pagination, and link rects all come from the engine.

### Feature-flag dispatch (no dead code, CI-validated)
Wired into `pdf::generate_pdf` behind the **`layout_pdf` Cargo feature** via `cfg!(feature = "layout_pdf")`:
- **Default build** → legacy renderer (**zero behavior change**).
- **`--all-features`** (CI + pre-push) → resumes route through the engine.

Both arms are compiled and *referenced*, so neither path rots and there's no dead code under either config (clippy `-D warnings` verified both ways).

### Behavior
- **ATS mode** linearizes the model (`transform::linearize`) and collapses two-column → single-column before layout.
- **Generation-meta** `candidate_name` overrides the header name.
- Page geometry is **A4**, matching the legacy renderer (locale-driven geometry is a later phase).
- **Cover letters stay on the legacy path** (modeled later).

### Parity tests (6, run regardless of the feature)
- **`pdf-extract` (text survival):** name, section headings, and bullets survive ✅ (confirms printpdf output is cleanly extractable)
- **`lopdf` (annotations):** LinkedIn, website, and `mailto:` links are embedded
- Two-column renders **both** columns; **ATS mode** keeps all content single-column; meta name overrides; **all 9 templates** produce a valid, parseable PDF

### Gate
- `cargo test` → **562 passed** · `cargo test --all-features` → **562 passed**
- `cargo clippy --all-targets [--all-features] -- -D warnings` → clean under **both** configs
- new files rustfmt-clean · no IPC change · full local pre-push gate green

---

### Also in this branch (per request, no separate branch): landing `user-select` fix
`ui:` commit — disables text selection on the `landing/index.html` doodle guys (`.doodle`/`.bubble`/`.dc-text`) so rapidly poking them (each click pops a speech bubble) never highlights the SVG or bubble text. `+ -webkit-touch-callout:none` to suppress the iOS long-press menu. Landing-only (deploys via `pages.yml`), `ui:` so it doesn't trigger an app release. `landing/` is in `.prettierignore`, so the format gate is unaffected.

### Next
**Phase 2f (flip):** once you've eyeballed a few engine-rendered PDFs, flip `layout_pdf` to default (or drop the legacy resume path). Then Phase 3 (cross-viewer page-size/font parity).

### DIFF
--- a/apps/tauri/src-tauri/Cargo.toml
+++ b/apps/tauri/src-tauri/Cargo.toml
@@ -5,6 +5,13 @@ edition = "2021"
 description = "AI Job Hunter — Tauri desktop shell"
 authors = ["Saeed Kolivand"]
 
+[features]
+# Render resumes through the canonical layout engine (model → LaidOutDoc → PDF)
+# instead of the legacy renderer. Off by default during the strangler-fig
+# migration; enabled by `--all-features` so CI and the pre-push gate exercise the
+# new path. Flips to default once golden-snapshot parity is locked.
+layout_pdf = []
+
 [profile.dev]
 incremental = true
 
--- a/apps/tauri/src-tauri/src/export/layout_pdf.rs
+++ b/apps/tauri/src-tauri/src/export/layout_pdf.rs
@@ -1,0 +1,203 @@
+//! PDF backend for the canonical layout engine.
+//!
+//! Strangler-fig step: this renders a resume by building the [`DocumentModel`],
+//! laying it out with [`crate::layout::layout_document`] into a backend-agnostic
+//! [`LaidOutDoc`], and translating that display list to `printpdf` ops. It does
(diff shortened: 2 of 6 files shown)


saeedkolivand:

PR #180, opened 2026-05-29
test: add ignored sample-pdf dump for legacy-vs-engine review

saeedkolivand:
## Sample-PDF dump for visual review

Adds `dump_sample_resume_pdfs`, an `#[ignore]`d dev test that writes a **legacy** and an **engine** PDF for every template (plus an **ATS** pair for the two-column template) to `target/sample_pdfs/`, from one realistic synthetic resume.

### Why
The automated Phase 2 gates prove **text survival, link annotations, and geometry** — not **visual polish**. Before flipping `layout_pdf` to default (making the engine the user-facing resume renderer), the output should be eyeballed against the legacy renderer. This makes that a one-liner.

### Run it
```text
cargo test -p ajh-tauri -- --ignored dump_sample_resume_pdfs --nocapture
```
Output: `<template>_legacy.pdf` / `<template>_engine.pdf` (+ `twocolumn_{legacy,engine}_ats.pdf`) under `apps/tauri/src-tauri/target/sample_pdfs/` — 20 files. Open the pairs side by side.

### Notes
- Both backends are compiled regardless of the `layout_pdf` feature, so no special build is needed.
- `#[ignore]`d → never runs in normal CI; output lands in the gitignored `target/` dir (no binaries committed).
- `cargo clippy --all-targets --all-features -- -D warnings` clean; rustfmt-clean.

### DIFF
--- a/apps/tauri/src-tauri/src/export/pdf/test.rs
+++ b/apps/tauri/src-tauri/src/export/pdf/test.rs
@@ -226,3 +226,92 @@ fn centered_template_resume_pdf_is_generated() {
         bytes.len()
     );
 }
+
+/// Dev tool (ignored): write legacy-vs-engine sample resume PDFs for every
+/// template to `target/sample_pdfs/`, so the canonical layout engine's output
+/// can be eyeballed against the legacy renderer before the `layout_pdf` flag is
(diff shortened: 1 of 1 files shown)


saeedkolivand:

PR #181, opened 2026-05-30
fix: match legacy vertical spacing in single-column layout engine

saeedkolivand:
## Fix: single-column layout engine compressed vertical spacing

Fixes the Phase-2 parity regression. Single-column resumes rendered through the layout engine were crammed into the top of the page — the section-header before-gaps had collapsed to roughly a bare line (reported −25%..−38%), pulling total content span ~18–22% short of legacy. `first_ink` was correct; the drift accumulated downward. **This was the blocker for flipping `layout_pdf` to default.**

### Root cause
During the model→layout refactor the single-column path **lost the legacy spacing wiring**. Legacy `pdf_renderer` derived vertical rhythm from `templates::calculate_spacing` (before/after gaps keyed on line kind — 12/3pt around section headers, plus entry/bullet gaps) **plus** the render functions' internal offsets (4pt lead above a heading, 10pt below its rule, the 2pt entry-title nudge). The engine instead used `section_spacing_before` with uniform trailing constants and **dropped every before-gap**, collapsing each template's rhythm. (Per-template airiness comes from `line_spacing`, already flowing through the line height — so the fix honors each template's own scale.)

### Fix
The single-column `Flow` now reproduces `calculate_spacing` (the shared source of truth) **plus** those exact internal offsets, tracking the previous line kind so context-dependent gaps (entry-after-bullet, first-bullet-after-title) match legacy. **Two-column is untouched** — gated behind a `single_col` flag on `Flow`; its reviewed spacing was already correct and is out of scope.

### Measured result (engine vs legacy, same resume)

| template | section-before gaps (mm) | total span Δ |
|---|---|---|
| classic | 13.7 / 17.8 / 17.8 | −5.0% |
| modern | 14.0 / 18.2 / 18.2 | −4.7% |
| executive | 14.1 / 18.6 / 18.6 | −4.4% |
| editorial-serif | 14.0 / 18.2 / 18.2 | −4.7% |
| swiss-minimal | **16.3 / 21.4 / 21.4** | −6.8% |
| mono-technical | 14.0 / 18.2 / 18.2 | −4.7% |
| refined-executive | 14.1 / 18.6 / 18.6 | −4.4% |
| academic | **13.4 / 17.3 / 17.3** | −4.0% |

Section gaps restored and **scale per template** (Swiss airiest, Academic tightest). The residual **−4.0%..−6.8% total span is intentional**: the canonical model doesn't carry the blank source lines the legacy renderer turns into ~3pt spacers, so the engine is uniformly a hair more compact.

### Parity gate (CI, can't regress) — component-level, not a raw span band
- **`single_column_inserts_real_section_spacing`** — the precise regression guard: every section-before gap (engine `LaidOutDoc`, no rasterization) exceeds a bare body line by the restored 12+4pt insertion. Exactly what the bug zeroed.
- **`single_column_honors_per_template_spacing_scale`** — Swiss Minimal (1.3) yields a taller block than Academic (1.1) for the same resume.
- **`single_column_first_ink_matches_legacy`** — name baseline identical to legacy (both real backends via `lopdf` `Td` baselines).
- **`single_column_span_backstop_vs_legacy`** — loose **±12%** span backstop only.

`pdf::generate_resume_pdf` is now `pub(crate)` so the gate can call the legacy backend regardless of the `layout_pdf` feature.

### Visual confirmation
Regenerated all 20 legacy-vs-engine sample PDFs (`target/sample_pdfs/`) via the #180 dump tool with this fix applied.

### Out of scope (unchanged, per the report)
Two-column output; `janedoe.dev` vs `https://janedoe.dev` website label; `•Led` bullet-glyph spacing (pre-existing in both backends).

### Gate
- `cargo test` & `cargo test --all-features` → green
- `cargo clippy --all-targets [--all-features] -- -D warnings` → clean both configs
- rustfmt clean · no IPC change · full pre-push gate green

### After merge
**Phase 2f flips `layout_pdf` to default.**

### DIFF
--- a/apps/tauri/src-tauri/src/export/pdf/mod.rs
+++ b/apps/tauri/src-tauri/src/export/pdf/mod.rs
@@ -13,7 +13,10 @@ use super::pdf_renderer::PageState;
 
 // ─── Resume PDF (single-column) ───────────────────────────────────────────────
 
-fn generate_resume_pdf(
+/// Legacy resume renderer. `pub(crate)` so the layout-engine parity gate
+/// (`layout::tests`) can compare its output against the new backend regardless
+/// of the `layout_pdf` feature; the public entry point stays [`generate_pdf`].
+pub(crate) fn generate_resume_pdf(
     text: &str,
     meta: Option<&GenerationMeta>,
     template: &Template,
@@ -56,7 +59,13 @@ fn generate_resume_pdf(
 
             LineKind::Name => {
                 let (line_ops, new_y) = render_name_line(
-                    &line.text, meta, &effective_template, &layout, &colors, &fonts, page_state.y,
+                    &line.text,
+                    meta,
+                    &effective_template,
+                    &layout,
+                    &colors,
+                    &fonts,
+                    page_state.y,
                 );
                 page_state.current_ops.extend(line_ops);
                 page_state.y = new_y;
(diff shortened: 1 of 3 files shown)


saeedkolivand:

PR #182, opened 2026-05-30
refactor: render resumes through the layout engine by default (phase 2f)

saeedkolivand:
## Phase 2f — flip the PDF default to the canonical layout engine

Completes the strangler-fig PDF migration (phases 2a–2e, #175–#181) by making
`layout_pdf` a **default Cargo feature**. Production builds (`tauri build`, which
uses default features) now render resumes through the canonical layout engine —
`model → LaidOutDoc → printpdf` — instead of the legacy renderer.

### Why this is safe now
Gap-by-gap snapshot parity with the legacy renderer is locked by `layout::tests`:
both backends render the same fixtures and every page-1 baseline gap must sit
within ±5% across all single-column templates, plus a total-span backstop and
two-column structural checks. That gate (added in #178–#181) is the precondition
the plan required before flipping.

### What stays
The legacy renderer remains **fully compiled** — the `cfg!(feature = "layout_pdf")`
switch keeps both arms alive, so it:
- stays the **parity reference** (`layout::tests` calls it directly),
- still renders **cover letters** (not yet migrated to the engine),
- is reachable via `--no-default-features`.

CI already exercised the engine under `--all-features`; that path is now identical
to the shipped default, so `--all-features` runs test exactly what users get.

### Verification
- `cargo clippy --all-targets -- -D warnings` clean under **both** `--all-features`
  and `--no-default-features`.
- Parity + layout + adapter + transform tests green under the **default** feature
  set (engine path): `71 passed, 1 ignored`.

This is the last item in Phase 2 — it unblocks Phase 3 (locale-driven page geometry
+ DOCX font fallback), which builds on the engine being the live PDF path.

### DIFF
--- a/apps/tauri/src-tauri/Cargo.lock
+++ b/apps/tauri/src-tauri/Cargo.lock
@@ -61,7 +61,7 @@ dependencies = [
 
 [[package]]
 name = "ajh-tauri"
-version = "0.23.2"
+version = "0.23.3"
 dependencies = [
  "anyhow",
  "apple-native-keyring-store",
--- a/apps/tauri/src-tauri/Cargo.toml
+++ b/apps/tauri/src-tauri/Cargo.toml
@@ -6,10 +6,13 @@ description = "AI Job Hunter — Tauri desktop shell"
 authors = ["Saeed Kolivand"]
 
 [features]
+default = ["layout_pdf"]
+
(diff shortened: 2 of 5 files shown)


saeedkolivand:
🎉 This is included in version 0.23.4.

PR #183, opened 2026-05-30
fix: give docx an explicit page size and predictable font fallback (phase 3)

saeedkolivand:
## Phase 3 — cross-viewer consistency (page size + DOCX fonts)

Low-risk consistency wins between the PDF and DOCX backends, with a single source
of truth for page geometry. No full model migration required.

### Page size — one source of truth
- The legacy PDF renderer (`pdf_renderer::setup_layout`) and the layout engine
  (`layout_pdf`) now read geometry from `LocaleProfile::default().page_geometry()`
  (A4) instead of hardcoded `210.0 × 297.0` literals.
- DOCX now sets its section page size **explicitly** (A4 in `dxa`, via the new
  `mm_to_dxa`) instead of relying on the docx-rs default.

> 📝 The plan assumed docx-rs defaulted to **Letter**. After the `0.4.20` bump
> (#174) it actually defaults to **A4** — so this isn't fixing a visible Letter
> bug, but setting the size explicitly makes `LocaleProfile`/`PageGeometry` the
> source of truth and guards against future default drift. Default stays
> consistent PDF↔DOCX.

### DOCX font fallback
The bundled TTFs (Inter, Source Serif 4, Manrope, JetBrains Mono, Playfair
Display) are **not embedded** yet (true embedding is Phase 5), so a reader
without them silently substitutes an arbitrary face. `docx_fallback_font` now
maps each bundled family to a common system font:

| Bundled | Fallback |
|---|---|
| Inter, Manrope, Calibri | Calibri |
| Source Serif 4 | Georgia |
| Playfair Display | Cambria |
| JetBrains Mono | Consolas |

Runs reference the fallback on **both** the ASCII and high-ANSI ranges, so
accented Latin (common in DACH names) renders in the same face. Output is now
predictable everywhere instead of depending on the reader's silent substitution.

### Tests
- **DOCX unzip invariants** (`word/document.xml`): explicit A4 `pgSz`
  (`11906 × 16838` dxa) for resume **and** cover letter; fallback fonts present
  (`w:ascii`/`w:hAnsi`), no bundled font names leak.
- `mm_to_dxa` matches Word's A4/Letter dxa; `docx_fallback_font` mapping locked.
- Legacy PDF `setup_layout` geometry asserted equal to the locale profile.

### Verification
- PDF A4 values unchanged → the Phase 2 layout-engine **parity gate stays green**.
- `clippy --all-targets --all-features -- -D warnings`: clean.
- `cargo test --all-features` (export/layout/locale/pdf): **86 passed, 1 ignored**.

True OOXML font embedding and locale-driven per-request page sizing are
intentionally deferred (Phases 5 and 7 respectively).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -215,3 +215,7 @@ CHANGELOG.md
 .cargo
 target/llvm-cov/
 .mcp.json
+graphify-out/
+
+post-checkout
+post-commit
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -211,3 +211,14 @@ Violations **fail the commit**. See `commitlint.config.mjs`.
 - **Subject ≤ 100 chars**; **body lines ≤ 200 chars**; blank line between subject and body.
 - **Type** is one of: `feat`, `fix`, `perf`, `refactor`, `ui`, `style`, `test`, `docs`, `build`, `ci`, `chore`, `revert` (`type-enum`). Only the types in the table above affect releases.
 - Imperative mood, no trailing period in the subject.
+
+## graphify
+
+This project has a knowledge graph at graphify-out/ with god nodes, community structure, and cross-file relationships.
+
+Rules:
+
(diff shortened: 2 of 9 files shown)


saeedkolivand:
🎉 This is included in version 0.23.4.

PR #184, opened 2026-05-30
feat: validate exports and auto-fix unreadable two-column layouts (phase 4)

saeedkolivand:
## Phase 4 — pre-export validation + ATS round-trip gate

After a resume/cover letter is rendered, the bytes are **re-extracted** the way a
parser (or a human pasting into a form) would read them, and checked for content
survival and sane reading order. The export is auto-fixed where possible and
blocked only as a last resort.

### New `validate` module
- `validate_and_fix(request, generate)` → renders, re-extracts (`pdf-extract` for
  PDF, unzipped `document.xml` for DOCX), evaluates, and returns the (possibly
  re-rendered) bytes plus an `ExportReport { ok, atsMode, issues, fixed }`.
- **Auto-fix:** a two-column layout whose sections interleave when read back is
  re-exported single-column (ATS-safe via `ats_mode`), re-checked, and the fix
  is recorded in `report.fixed`.
- **Block:** only when a critical issue *survives* auto-fix → returns a
  plain-language error, no file (the behavior chosen for this phase).

### Severity model — tuned to never false-block a valid document
| Issue | Severity |
|---|---|
| Two-column sections interleave on read-back | **Critical** (auto-fixable → linearize) |
| No extractable text at all | **Critical** (unfixable → blocks) |
| Missing name / email / a section, single-column order quirks | Warning (surfaced, non-blocking) |

Because the only auto-fixable critical is two-column interleave (always fixed by
linearizing) and the only unfixable critical is an empty-text document (never
true for a real render), valid exports are never blocked — verified by
end-to-end tests.

### Wiring
- `documents_export_document` routes PDF/DOCX through the gate, so **both** the
  download and the save-to-disk flow receive auto-fixed bytes.
- `ExportResult` gains an optional `report` and now serializes camelCase
  (incidentally fixing the long-latent `mime_type`/`mimeType` mismatch — no real
  consumer read the old field). TXT has no layout and carries no report.
- IPC contract mirrors `ExportReport`/`ExportIssue` as **optional** on the
  `exportDocument` result (backward-compatible). The renderer surfacing UI is
  Phase 6.

### Tests / gates
- `evaluate` unit-tested on synthetic extracted text (interleave→critical,
  single-column reorder→warning, empty→critical, missing section→warning).
- `validate_and_fix` exercised end-to-end: real single-column PDF, DOCX, and
  two-column PDF exports are never falsely blocked; TXT returns unvalidated.
- `cargo test --all-features` green incl. the Phase 2 **layout parity gate**;
  `clippy -D warnings` clean; `pnpm typecheck` + `gen:ipc:check` clean.

### DIFF
--- a/apps/tauri/src-tauri/Cargo.lock
+++ b/apps/tauri/src-tauri/Cargo.lock
@@ -61,7 +61,7 @@ dependencies = [
 
 [[package]]
 name = "ajh-tauri"
-version = "0.23.3"
+version = "0.23.4"
 dependencies = [
  "anyhow",
  "apple-native-keyring-store",
--- a/apps/tauri/src-tauri/src/export/commands/mod.rs
+++ b/apps/tauri/src-tauri/src/export/commands/mod.rs
@@ -7,6 +7,7 @@ use super::{
     types::{ExportFormat, ExportRequest, ExportResult},
 };
 use crate::error::{AppError, AppResult};
+use crate::validate::{validate_and_fix, ExportReport, Severity};
 
 /// Tauri command to export resume or cover letter
 #[command]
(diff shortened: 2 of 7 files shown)


saeedkolivand:
🎉 This is included in version 0.24.0.

PR #185, opened 2026-05-30
feat: render resume docx from the canonical model with real two-column (phase 5)

saeedkolivand:
## Phase 5 — DOCX onto the canonical model

Migrates the resume DOCX backend onto the `DocumentModel`, mirroring the PDF
layout engine. New `export::model_docx`, gated behind the `model_docx` Cargo
feature (**on by default**; `--no-default-features` falls back to the legacy
line-based renderer, which still renders cover letters). Both arms compile via
`cfg!` so neither path rots — the same strangler-fig pattern as `layout_pdf`.

### What the model backend does differently
- **Builds a `DocumentModel`** from resume text via the shared adapter instead of
  re-parsing flat lines.
- **Real two-column** = borderless, single-row two-cell table: a shaded sidebar
  cell (fill = template tint) + a main cell, fixed layout, borders cleared,
  letting Word flow/paginate. Section → column assignment uses the canonical
  `theme::placement_for` decision — the **same source the PDF engine uses** — so
  PDF and DOCX agree on the sidebar.
- **Native `ats_mode`**: `transform::linearize` reorders the model and the table
  is dropped for a single column — removing the old *"DOCX ignores ats_mode"* gap
  (it previously always force-collapsed two-column to one column).
- **First-class links**: contact/body links become real hyperlinks
  (`Hyperlink::new` + relationship targets), showing the label, not the raw URL.
- Reuses the Phase-3 **font fallback** + explicit **A4** page size, and adds
  `keepNext` on headings/entry titles and `keepLines` on bullets (no orphaned
  headings, no split bullets).

### Tests — golden OOXML invariants (`export::model_docx::test`)
- two-column emits a shaded borderless `<w:tbl>` and splits sections between columns;
- ATS mode emits **no** table and keeps every section;
- single-column templates use no table;
- contact links become `<w:hyperlink>` with correct `.rels` targets, no raw URLs as text;
- A4 page size + font fallback hold.

The existing `export::docx` and Phase-4 `validate` suites now run **through** the
model path and pass; the PDF parity gate is unaffected.

### Verification
- `cargo test --all-features` (model_docx + docx + validate + layout parity) green.
- `clippy -D warnings` clean under **both** `--all-features` and `--no-default-features`.
- DOCX-only Rust change — no IPC/TS surface touched.

Cover-letter modeling and true OOXML font embedding remain deferred (a stretch
goal even in the plan).

### DIFF
--- a/apps/tauri/src-tauri/Cargo.lock
+++ b/apps/tauri/src-tauri/Cargo.lock
@@ -61,7 +61,7 @@ dependencies = [
 
 [[package]]
 name = "ajh-tauri"
-version = "0.23.4"
+version = "0.24.0"
 dependencies = [
  "anyhow",
  "apple-native-keyring-store",
--- a/apps/tauri/src-tauri/Cargo.toml
+++ b/apps/tauri/src-tauri/Cargo.toml
@@ -6,7 +6,7 @@ description = "AI Job Hunter — Tauri desktop shell"
 authors = ["Saeed Kolivand"]
 
 [features]
-default = ["layout_pdf"]
+default = ["layout_pdf", "model_docx"]
 
 # Render resumes through the canonical layout engine (model → LaidOutDoc → PDF)
 # instead of the legacy renderer. On by default now that gap-by-gap snapshot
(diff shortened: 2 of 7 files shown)


saeedkolivand:
🎉 This is included in version 0.25.0.

PR #186, opened 2026-05-30
feat: structured resume extraction with html/rtf formats and a review step (phase 6)

saeedkolivand:
## Phase 6 — structured extraction + new formats + review

The deterministic, fully-testable backbone of Phase 6. **OCR stays in the
renderer** (Tesseract.js — already wired in `ocr.ts` + `use-import-with-ocr`; the
plan's "unwired" note was stale), so Rust extraction is text-only.

> **Why not leptess?** Evaluated and declined: it binds native
> Leptonica/Tesseract via bindgen, which aren't present here or on the bare
> `ubuntu-latest` cargo runner — it wouldn't compile, blocking *all* Rust gates —
> and it duplicates the working TS OCR.

### Extraction registry
`match ext{}` → `trait Extractor` + registry. Adding a format is one entry plus a
module. pdf/docx/plain refactored into impls; image/legacy-doc keep returning the
errors the renderer OCR fallback relies on.

### New formats
- **HTML** — drop script/style/head, `<a href>` → `[label](url)` (+ link
  collection), block tags → line breaks, entity decode.
- **RTF** — control-word state machine: skips non-content destinations
  (font/color/style tables, pictures, info), maps par/line/tab/`\uN`/`\'xx`,
  recovers `HYPERLINK` field targets.

### Layout-aware cleanup
Strip Unicode **Private Use Area** + icon-font glyphs, the replacement char, and
control chars from extracted PDF/DOCX text (`extraction::clean`); `normalize_unicode`
extended to drop PUA on the render side too.

### Structured extraction
`extraction::structured` turns flat text into a typed **`StructuredResume`**:
name / email / phone / location / links + a section inventory, each with a
`Confidence` and the **byte span** it was found at, plus `reviewRequired` +
warnings. Never drops content. `build_model` reconciles the typed contact fields
into a `DocumentModel` (deterministic contact wins) — implemented + tested, wired
into export as a follow-up.

### Surfacing (IPC + UI)
- `documents_import` returns the structured `review`; shared contract mirrors
  `StructuredResume` / `ResumeField` / `SectionSummary` (optional, backward-compatible).
- New **`ResumeReviewPanel`** — detected fields with per-field confidence badges,
  links, sections, warnings — shown from `ResumeInputCard` when `reviewRequired`.
  HTML/RTF added to the upload accept list; en/de strings added.

### Verification
- **44** extraction tests (registry, HTML, RTF, PUA, structured pre-pass);
  broader Rust suite **204** passed; **128** renderer tests passed.
- `clippy -D warnings` clean (both feature configs), `typecheck`, `gen:ipc:check`,
  `lint:strict` all clean.

### Deferred (follow-ups)
Live AI JSON structuring stage (not CI-verifiable), exporting from `build_model`,
and adopting the review panel in the onboarding/settings import surfaces.

### DIFF
--- a/apps/tauri/src-tauri/Cargo.lock
+++ b/apps/tauri/src-tauri/Cargo.lock
@@ -61,7 +61,7 @@ dependencies = [
 
 [[package]]
 name = "ajh-tauri"
-version = "0.24.0"
+version = "0.25.0"
 dependencies = [
  "anyhow",
  "apple-native-keyring-store",
--- a/apps/tauri/src-tauri/src/commands/documents.rs
+++ b/apps/tauri/src-tauri/src/commands/documents.rs
@@ -44,6 +44,12 @@ pub async fn documents_import(app: AppHandle, req: DocumentsImportRequest) -> Va
         tracing::warn!(warning = %w, file = %req.name, "extraction warning");
     }
 
+    // Structured review: per-field confidence + missing/low-confidence flags the
+    // renderer surfaces before generation. Computed before `extraction.text` is
+    // moved into the record.
+    let review = serde_json::to_value(crate::extraction::structured::structure(&extraction))
+        .unwrap_or(json!(null));
+
     let store = app.state::<crate::documents::DocumentStore>();
     let doc_id = crate::documents::make_doc_id();
     let record = crate::documents::DocumentRecord {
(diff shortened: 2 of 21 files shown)


saeedkolivand:
🎉 This is included in version 0.26.0.

PR #188, opened 2026-05-30
feat: drive export page size from the recommended locale

saeedkolivand:
## Locale-driven export page size

Completes the deferral from phases 3 and 7: the recommended locale now flows all
the way to the page geometry. A **US** export renders **US Letter**; every other
market renders **A4**.

### Rust
- `ExportRequest` gains an optional `locale` (`#[serde(default)]`) and a
  `page_geometry()` that resolves it through `LocaleProfile::get` (US → Letter,
  else → A4; omitted → A4).
- The default production resume backends take the geometry as a parameter:
  `layout_pdf` and `model_docx` gain `…_in(geom)` variants the export command
  calls with `request.page_geometry()`. The old default-geometry helpers become
  test-only conveniences (`#[cfg(test)]`).

### Renderer (end-to-end)
- The session store carries the active `locale`; **applying a template
  recommendation also applies its locale**, and `exportPDF` / `exportDOCX` pass it
  through `BaseExportRequest`. So a US recommendation → Letter export, no extra
  clicks.

### Tests
- The layout engine honors Letter geometry; a `locale = "us"` DOCX declares the
  Letter `pgSz` (`12240 × 15840` dxa) while the default stays A4.
- The PDF↔legacy **parity gate is unaffected** (A4 unchanged).
- clippy `-D warnings` clean under **both** feature configs; typecheck,
  gen:ipc:check, lint:strict, **132** Rust and **128** renderer tests all pass.

This is the first item of the deferred "afterwards" bucket. Next: `docs/EXPORT_TEMPLATES.md`.
The `build_model` wiring stays forward-looking (it belongs to a "render the
*imported* resume" feature that doesn't exist yet — export renders AI-*generated*
text), and live AI structuring stays deferred (not CI-verifiable).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -188,6 +188,7 @@ No `// eslint-disable`, no `@ts-ignore`. Add scoped overrides to `eslint.config.
 | Architecture principles | `docs/PATTERNS.md` §13                                                          |
 | Architecture status     | `docs/ARCHITECTURE_STATUS.md`                                                   |
 | Architecture (general)  | `docs/ARCHITECTURE.md`                                                          |
+| Export templates        | `docs/EXPORT_TEMPLATES.md` (9-template + backend contract)                      |
 | Patterns                | `docs/PATTERNS.md`                                                              |
 | Design system           | `docs/DESIGN_SYSTEM.md`                                                         |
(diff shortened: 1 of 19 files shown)


saeedkolivand:
🎉 This is included in version 0.28.0.

PR #190, opened 2026-05-30
feat: make autopilot inputs shape fetched results and gate apply/work-type as coming soon

saeedkolivand:
@
## Summary

Audited the autopilot creation form and found several inputs were **dead config** — saved but never applied to the jobs a run fetches. This wires the genuinely-supported inputs to the results and defers the not-ready pieces behind "Coming soon".

### Inputs now actually affect results (Rust `autopilot_run`)
- **Must-include keywords** (all must be present) and **exclude keywords** (any drops the job) now filter the scraped postings — previously stored but never read.
- **Match score scale fixed (0–1 → 0–100):** `simple_similarity` returned a Jaccard fraction compared against a 0–100 threshold, so nothing ever matched and found-job scores rendered as `0%`. Every saved posting is now scored against the resume and results are **sorted highest-first**; the `%` badges show real values.
- **A run only fetches & saves results** — the apply loop is removed. This also fixes a latent bug where a "Save" autopilot still launched applications (the `action` field was never honored). The dead threshold-ranker is removed; `simple_similarity` is now called inline.
- Added unit tests: keyword filter (ALL include / ANY exclude / empty no-op / case-insensitive) and 0–100 scaling.

### Coming soon + disabled (renderer)
- **Apply & review**, **Auto-apply**, and the **Work Type** selector show a "Coming soon" badge and are disabled/unselectable (work type only ever affected LinkedIn; applying is deferred). **Save** stays the default and only selectable action.
- Shared `ComingSoonBadge`, an optional `badge` slot on `WizardField`, and en/de strings.

Per-input verdicts and the trace are in the plan. No schema/IPC changes — `keywords`, `excludeKeywords`, `workType`, and the `action` enum already existed.

## Testing
- Rust: `cargo clippy` clean; 12 autopilot tests pass (5 new).
- TS: `pnpm typecheck` clean; `eslint --max-warnings 0` clean; autopilot + locale/i18n tests pass.
- Manual: create an autopilot with must-include + exclude keywords and a resume -> Run -> found jobs reflect the filters, show real 0–100 scores sorted high->low, applied = 0, no browser/apply launched. Apply actions + Work Type show "Coming soon" and are unclickable.

@

### DIFF
--- a/apps/tauri/src-tauri/src/commands/autopilot.rs
+++ b/apps/tauri/src-tauri/src/commands/autopilot.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use parking_lot::Mutex;
 
-use crate::autopilot::{AutopilotStatus, AutopilotStore, FoundJob};
+use crate::autopilot::{AutopilotFilter, AutopilotStatus, AutopilotStore, FoundJob};
 use crate::autopilot_helpers::autopilot_scrape;
 use crate::scraping::{JobPosting, ScraperEngine};
 use serde_json::{json, Value};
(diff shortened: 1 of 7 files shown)


saeedkolivand:
🎉 This is included in version 0.30.0.

PR #191, opened 2026-05-30
feat: job-url + linkedin imports and macos cli-agent detection

saeedkolivand:
## Summary

A batch of related fixes/features around job-ad/profile import and CLI-agent detection (4 commits).

### fix: detect CLI agents on macOS by augmenting the GUI PATH
macOS/Linux apps launched from Finder/Dock inherit only a minimal PATH, so `Command::new("claude")` couldn't find a CLI installed via npm/nvm/Homebrew or the native installer (`~/.claude/local`) — Claude Code (and the other CLI agents) showed as not installed even when a terminal finds them. Added `platform::cli_path()`: a cached PATH that merges the login-shell PATH (probed via `$SHELL -lic`, time-bounded) with the process PATH and common install dirs, applied to CLI-agent detection + generation. No-op on Windows.

### feat: import job description from a URL (AI Generate + Resume Analyzer)
A "Link" button next to Upload in the job-ad field: paste a job posting URL (LinkedIn, Greenhouse, Lever, Workday, …) and pull its full description into the field, reusing the existing `scrape.resolveUrl` backend. Added a `headerAction` slot to `CollapsibleFileInput` so the importer is a compact toggle rather than an always-visible row.

### feat: enable LinkedIn profile import in onboarding + Settings
The profile-URL import was live everywhere `ResumeInputCard` is used, but onboarding and Settings → Resume still rendered a "coming soon" stub. Turned it into a real component that imports a profile and **saves it as a resume document** (onboarding sets it as default). Public profiles import with no login; profile import now also uses the connected LinkedIn session when available, so a private profile becomes importable after login — otherwise the UI shows a "this profile may be private, connect LinkedIn" hint. Shared the validator + auth-error helper across both import paths.

## Testing
- Rust: `cargo check` + `cargo clippy` clean; cli_agent + updater + profile tests pass.
- TS: `pnpm typecheck` clean; `eslint --max-warnings 0` clean; affected vitest suites pass.
- The macOS PATH path is `cfg(not(windows))` and was hand-audited (Windows toolchain only); worth a smoke test on macOS.


saeedkolivand:
🎉 This is included in version 0.31.0.

PR #193, opened 2026-05-30
fix: use linkedin time datetime attribute for accurate posted dates

saeedkolivand:
## Summary

Scraped LinkedIn jobs showed the wrong age ("1h ago" for months-old posts). The
scraper read the `<time>` element's **visible text**, which LinkedIn renders
relative to the repost/refresh time, not the original posting.

- **`scraping/linkedin/api_client/mod.rs`** — prefer the `<time datetime="…">`
  attribute (RFC 3339, then `YYYY-MM-DD`); fall back to the relative text only when
  the attribute is absent. Also fixed `parse_relative_time` so `"minutes"` is no
  longer matched as `"months"` via the bare `"m"` stem.
- **Tests** — strengthened to assert actual elapsed values (not just `is_some`) and
  cover the ISO-date parser + attribute-wins behaviour.

> Note: an earlier revision of this branch also changed PDF résumé link extraction
> and the cover-letter DOCX renderer. Those were reverted after they regressed the
> résumé contact line, so this PR is now scoped to the LinkedIn date fix only.

## Test plan

- `cargo test` — LinkedIn `api_client` suite (22) passes; clippy clean.
- Manual: scrape LinkedIn, pick a job LinkedIn shows as months old, confirm the
  app's "… ago" now matches.

### DIFF
--- a/apps/tauri/src-tauri/src/export/docx/mod.rs
+++ b/apps/tauri/src-tauri/src/export/docx/mod.rs
@@ -9,7 +9,6 @@ use super::{
 use crate::locale::LocaleProfile;
 
 use super::docx_renderer::*;
-use super::links::{split_urls, Span};
 
 /// Page size (in DOCX `dxa`) for the active locale. Defaults to the `en` profile
 /// (A4), keeping DOCX on the same page geometry as the PDF backends. Set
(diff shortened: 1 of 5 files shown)


saeedkolivand:
🎉 This is included in version 0.33.0.

PR #194, opened 2026-05-30
feat: section-aware resume context and skill grounding for tailored generation

saeedkolivand:
## Summary

Improves the accuracy of tailored résumé and cover-letter generation by fixing **how the résumé is fed to the model**, not by rewording the (already thorough) prompts.

### The bottleneck

`buildResumePrompt` / `buildCoverLetterPrompt` fed the résumé with a blind prefix slice:

```ts
resumeBody.slice(0, resumeChars)   // 2,500 chars for local tiers
```

A typical 2-page résumé is 4–6k chars, so the **tail (Skills, Education, older-but-relevant roles) was silently dropped** and the model tailored against a partial résumé. The package already shipped a section-aware `truncateResume()` + a per-tier `TruncationStrategy` (via `resolveProfile().truncation`) — but the builders never called it.

### Changes (prompts package only)

- **Section-aware context** — `resume.ts` / `cover-letter.ts` now feed the résumé through `truncateResume(body, truncation)`:
  - keeps the **whole résumé** when it fits the model's token budget (the common case);
  - otherwise preserves high-value sections (Experience, Skills) instead of cutting the tail.
  - This also raises the effective budget well past the old 2,500-char cap.
- **Skill grounding** — new `buildGroundingBlock()` splits the job-ad top requirements into **PRESENT** (résumé-backed → safe to emphasize) and **ABSENT** (never claim/imply/bold). A verifiable anti-hallucination signal, wired into both prompts next to the existing emphasis block. Matching is heuristic (`resumeMentions`: word-boundary for tokens, substring for `Node.js`/`REST API`).
- **Job-ad budgets** — raised `jobAdChars` (provider) and the metadata job-ad slice so late requirements aren't truncated before extraction/generation read them.

**Résumé extraction is intentionally untouched** (separate subsystem; left exactly as-is).

## Notes / follow-ups

- Generation passes a tier string, so cloud models currently use the local `large` strategy rather than their full context window. Plumbing a `ProviderProfile` (with `contextWindow`) would unlock the larger cloud budget — left as a follow-up.
- `truncateResume` processes sections in document order; a pathologically huge early section can still crowd out later ones. Real résumés don't hit this; noted for future hardening.

## Test plan

- `pnpm --filter @ajh/prompts test` — 87 pass, incl. new `resumeMentions`, `buildGroundingBlock`, and wiring tests (grounding present in both prompts; résumé tail past 2,500 chars now survives on local tiers). Typecheck + ESLint clean.
- Manual: generate a tailored résumé/cover letter from a 2-page résumé on a local model and confirm later sections (Skills/Education) are now reflected, and that job-ad skills the résumé lacks are no longer asserted.

### DIFF
--- a/packages/prompts/src/generate/cover-letter.ts
+++ b/packages/prompts/src/generate/cover-letter.ts
@@ -1,7 +1,8 @@
 /** Cover-letter generation — system prompt (brief / task / full) + user prompt. */
 
+import { truncateResume } from '../context-manager/index.js';
 import { type PromptTarget, resolveProfile } from '../provider/index.js';
-import { buildEmphasisBlock } from './emphasis.js';
+import { buildEmphasisBlock, buildGroundingBlock } from './emphasis.js';
 import { parseLinksFromResume, stripLinkBlock } from './links.js';
 import { type GenerationMeta, type GenerationMode, MODES } from './modes.js';
 
@@ -109,7 +110,7 @@ export function buildCoverLetterPrompt(
   _mode: GenerationMode,
   target: PromptTarget = 'large'
 ): string {
-  const { resumeChars, jobAdChars } = resolveProfile(target);
+  const { jobAdChars, truncation } = resolveProfile(target);
   // Date in the target language's convention, following the job-ad locale.
   const today = new Date().toLocaleDateString(meta.targetLanguage || 'en', {
     day: 'numeric',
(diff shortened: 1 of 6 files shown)


saeedkolivand:
🎉 This is included in version 0.33.0.

PR #198, opened 2026-05-31
fix: anchor header links, contact-profile header source of truth, export typography + validation

saeedkolivand:
Fixes the five PDF/DOCX export defects surfaced by PyMuPDF inspection, generalized to any résumé (not the one data set), plus a validation gate so they can't regress. Two commits: `28ea867f` (Symptoms 1/3/4/5) + `abec14a3` (Symptom 2 + contact model + Step 3 validation).

## Root causes & fixes

**Symptom 1 — cover-letter header links dumped at page bottom (y≈752/842).** The legacy contact renderer's `y` is already bottom-up, but the link rect re-subtracted it from page height → a second flip, so all five header link rects landed at the bottom with no text under them. Extracted one shared `link_annotation_op` (top-down mm → single correct flip) used by **both** the modern layout engine and the legacy cover-letter/contact path, anchored to the text's own baseline. No fixed coordinates, one annotation code path. (The résumé was fine because it renders through the modern engine, which flips correctly.)

**Symptom 2 — résumé header URLs swapped (a company LinkedIn / employer site displacing the candidate's own).** The header was assembled from links scavenged out of the résumé by a domain heuristic + document position, so `linkedin.com/company/jibit` and `rabobank.com` could win header slots. Replaced with a **`ContactProfile`** model (single-row SQLite store, mirrors `job_preferences`) as the single source of truth: the header is built from **named fields only**, localized per language, identical across résumé / cover letter / DOCX. A name-based classifier seeds it on import (personal `/in/` LinkedIn, GitHub, personal site; rejects `/company/`, `/school/`, job-board hosts) as an **editable suggestion**, never silently trusted. Full IPC (`contact_profile_get/set`), shared contract, tauri-client namespace, React Query hook, and a Contact Profile settings tab (en/de).

**Symptom 3 — stray asterisks (`*React`, `AWS*`).** Added `sanitize_markdown` to drop lone `*`/`_`/backtick that aren't part of a valid `**bold**` pair; stopped `normalize_unicode` mapping daggers (U+2020/2021) → `*`.

**Symptom 4 — sentence-break dash glued to the previous word (`zu Hause- die`).** `normalize_unicode` was collapsing en/em-dash to a bare `-`. Now preserves U+2013/U+2014 (a font-glyph test asserts every bundled face carries them) and a `typography` pass normalizes clause dashes to a spaced en-dash and rewrites the `word- word` break pattern — excluding German suspended hyphens (`Backend- und …`) and tight compounds/ranges (`2020–2023`, `state-of-the-art`).

**Symptom 5 — orphaned role title / role split onto a near-empty page.** Keep-with-next in `layout::lay_entry` binds title + subtitle + first bullet, sized at/below the legacy reservation so the gap-by-gap parity gate still holds.

## Validation (can't regress)
`validate/mod.rs` now hard-fails **every** generated document on:
- `empty_anchor_link` — a link rect with no text under it (the cover-letter bottom bug),
- `header_url_mismatch` — a header link URL not in the contact profile (the URL-swap bug),
- `stray_markdown` — `*` / backtick that leaked into rendered text.

Wired into the existing export gate (blocks on `Critical`) and covered by `validate` tests. `tests/architecture.rs` updated: `contact_profile` classified L1 + R3 rusqlite allowlist.

## Verification
- `cargo test` green (layout parity, architecture rules, new validate gates).
- `@ajh/shared` 80 + `@ajh/prompts` 87 vitest passing; tauri typecheck + `lint:strict` clean.
- Full pre-push CI gate passed (cargo check/test/clippy/fmt + JS build/lint/typecheck/test).

### DIFF
--- a/apps/tauri/src-tauri/Cargo.lock
+++ b/apps/tauri/src-tauri/Cargo.lock
@@ -61,7 +61,7 @@ dependencies = [
 
 [[package]]
 name = "ajh-tauri"
-version = "0.33.0"
+version = "0.33.2"
 dependencies = [
  "anyhow",
  "apple-native-keyring-store",
--- a/apps/tauri/src-tauri/src/commands/contact_profile.rs
+++ b/apps/tauri/src-tauri/src/commands/contact_profile.rs
@@ -1,0 +1,25 @@
+use serde_json::json;
+use serde_json::Value;
+use tauri::AppHandle;
+use tauri::Manager;
+
+use crate::contact_profile::{ContactProfile, ContactProfileStore};
+
+#[tauri::command]
+pub async fn contact_profile_get(app: AppHandle) -> Value {
+    let store = app.state::<ContactProfileStore>();
+    json!(store.get())
+}
+
+#[tauri::command]
+pub async fn contact_profile_set(app: AppHandle, profile: Value) -> Value {
+    let store = app.state::<ContactProfileStore>();
+    let parsed: ContactProfile = match serde_json::from_value(profile) {
+        Ok(p) => p,
+        Err(e) => return json!({ "error": format!("invalid contact profile: {e}") }),
+    };
+    match store.set(&parsed) {
+        Ok(()) => json!({ "success": true }),
+        Err(e) => json!({ "error": e.to_string() }),
+    }
+}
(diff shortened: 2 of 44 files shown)


saeedkolivand:
🎉 This is included in version 0.33.3.

PR #202, opened 2026-05-31
feat: contact-profile header autofill, wrapping, link editor, and first-run prompt

saeedkolivand:
## Summary

Fixes a chain of contact-profile header problems (exports were fully blocked, then produced broken headers once unblocked) and adds the missing editing + first-run UX. Backend fix + frontend feature, split into two commits.

## Background

The pre-export validator read PDF link annotations via lopdf's `get_page_annotations`, which only resolves *indirect-reference* `/Annots`. Our renderer (printpdf) writes them as **inline dictionaries**, so the validator saw **zero** header links and reported every contact-profile URL as "missing from the rendered header" → **every export with a contact profile was hard-blocked**. Removing that false block then exposed pre-existing header bugs from the contact-profile feature.

## Backend (`fix:`)

- **Read `/Annots` ourselves** (inline dictionaries *and* references) so the header-link validation works against our own renderer; downgrade the geometry-fragile "missing header link" check from blocking to advisory (the leak/URL-swap direction stays blocking).
- **Autofill the contact profile on import** — `classify_contact_links` keeps every other personal link as a labelled extra (Dribbble, Behance, …); email/phone/location come from the structuring pass; merged via `fill_empty_from`, filling only empty fields so user edits are never clobbered.
- **Stop the cover-letter contact-line leak** — drop any pre-salutation line that looks like a contact line (an `@` or ≥2 separators), so the generated letter's own contact line never leaks into the body as raw markdown.
- **Wrap long header contact lines** instead of overflowing the page margins, in both the résumé layout engine and the legacy letterhead; the single-line path stays byte-identical when it fits (golden parity). Contact-line geometry extracted into a `contact_layout` submodule to stay under the R8 LOC cap.

## Frontend (`feat:`)

- **"Other links" editor** (label/URL rows, add/remove) in the contact profile so autofilled extras are editable, not just visible. Fields extracted into a shared `ContactProfileForm` reused by Settings and the new prompt.
- **First-run prompt** — before the first résumé/cover-letter generation, a one-time modal surfaces that same form so the header is complete. Gated by a persisted `contactPromptSeen` flag (shows once, never again).

## Recovery

An already-sparse stored profile completes itself on the next résumé **re-import** (the merge fills email/phone/location/extras and keeps the user's edits).

## Testing

- New Rust tests: inline-dict annotation reading, no false block with a contact profile, advisory "missing" warning, classify extras, `fill_empty_from`, cover-letter no-leak, résumé + cover-letter long-line wrapping (links on ≥2 baselines, within page).
- Full backend suite green; clippy `-D warnings` clean; `cargo fmt` clean; R8 architecture cap satisfied.
- Frontend: ESLint, prettier, `@ajh/tauri` typecheck, preferences-store tests all green.

### DIFF
--- a/apps/tauri/src-tauri/Cargo.lock
+++ b/apps/tauri/src-tauri/Cargo.lock
@@ -61,7 +61,7 @@ dependencies = [
 
 [[package]]
 name = "ajh-tauri"
-version = "0.33.2"
+version = "0.34.1"
 dependencies = [
  "anyhow",
  "apple-native-keyring-store",
--- a/apps/tauri/src-tauri/src/commands/documents.rs
+++ b/apps/tauri/src-tauri/src/commands/documents.rs
@@ -50,20 +50,37 @@ pub async fn documents_import(app: AppHandle, req: DocumentsImportRequest) -> Va
     let structured = crate::extraction::structured::structure(&extraction);
     let review = serde_json::to_value(&structured).unwrap_or(json!(null));
 
-    // Seed the contact profile from the résumé's links — but ONLY when the user
-    // has no profile yet, so an existing (edited) profile is never clobbered. The
-    // classifier picks the personal LinkedIn/GitHub/Website by NAME and rejects
-    // company / job-board pages; the user reviews & edits it in settings. This is
-    // what stops a company link from ever reaching the document header.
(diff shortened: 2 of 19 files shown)


saeedkolivand:
🎉 This is included in version 0.35.0.

PR #203, opened 2026-06-01
feat: autopilot apply modal — reasoning box, cancel, default to both

saeedkolivand:
## What & why
Part C of the AI/desktop plan. The autopilot **Apply Job** modal now mirrors the AI Generate flow so users see what the model is doing and control it.

## Changes
- **Default target → Both** (was Cover letter).
- **Button order → Resume · Cover letter · Both**.
- **Cancel button** while generating — aborts the in-flight job (`gen.abort()`); previously only a disabled "Generating…" state showed.
- **Model reasoning box** — reuses `ThinkingBubble` (same component AI Generate + Analyze already use), fed by a new `onThinking` buffer in `useTailorGeneration` that clears at the start of each phase. Self-hides when the model emits no reasoning. (Once provider-side emission parity lands, every provider/model fills it.)
- i18n: new `autopilot.apply.cancel` key (en, de).

## Tests
- New `useTailorGeneration.test.ts`: idle defaults, reasoning threads into `thinking` and clears between résumé/cover phases, guarded when AI unavailable.
- `pnpm exec vitest run …` → 3 passed · `turbo typecheck --filter=@ajh/tauri` → clean · `eslint --max-warnings 0` on the changed files → clean.

## Notes
Result-loss on tab switch / modal close is a separate, larger change (background generation store) tracked as **D-3** in the plan.

### DIFF
--- a/apps/tauri/src/renderer/features/autopilot/components/ApplyJobModal/index.tsx
+++ b/apps/tauri/src/renderer/features/autopilot/components/ApplyJobModal/index.tsx
@@ -6,6 +6,7 @@ import { Button, cn, ModalShell } from '@ajh/ui';
 
 import { ResumeInputCard } from '@/components/resume/ResumeInputCard';
 import { ModelSelector, useCanUseAI, useSelectedModel } from '@/components/ui/ModelSelector';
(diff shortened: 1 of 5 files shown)


saeedkolivand:
🎉 This is included in version 0.36.0.

PR #206, opened 2026-06-01
perf: glyph-subset embedded fonts to shrink exported pdf size

saeedkolivand:
## Problem

Résumé PDFs export at **~3 MB** while the equivalent DOCX is ~32 KB.

**Root cause:** font embedding. printpdf 0.9.1 has its own font subsetting **hard-disabled** at serialize time (`serialize.rs`: `if false && do_subset …`) and it already prunes any face that no glyph references — so every export embeds the **full** remaining face. Calibri regular+bold alone is ~3.2 MB, which is essentially the whole file.

> The plan's other proposed lever — "template-aware font loading" — turned out to be a **no-op for output size**: printpdf already skips faces with no used glyphs, so only the selected template's face was ever embedded. Dropped it.

## Fix

Subset each face to the document's **actual codepoints** before `add_font`, using printpdf's own allsorts subsetter — already compiled in via the default `html` → `text_layout` feature, so **no new dependency**.

- `collect_codepoints` gathers the glyphs a document draws: from the **laid-out runs** on the engine path (the authoritative rendered text), and from the text/name/contact inputs on the legacy path — seeded with a **baseline** of the punctuation/bullet/quote/umlaut glyphs the renderers inject directly.
- `parse_font` maps those codepoints → glyph ids, subsets, re-parses, and embeds the subset.
- **Graceful fallback:** any failure (parse, subset, or re-parse) embeds the full font, so the worst case is exactly today's output — never a missing glyph.
- The font loading/subsetting code moves to a new `pdf_renderer/fonts.rs` submodule (re-exported, no API change) to keep `pdf_renderer/mod.rs` under the **R8** module-size cap.

## Result (sample résumé, engine path)

| Template | Before | After |
|---|---|---|
| ATS Classic (Calibri) | ~3 MB | **~120 KB** |
| Modern (Inter) | — | ~120 KB |
| Mono Technical | — | ~55 KB |
| Two-Column | — | ~56 KB |

## Tests

- **Size guardrail**: a Calibri résumé must stay **< 800 KB** (catches a full-font regression by an order of magnitude).
- `collect_codepoints` unit tests: baseline seeding + dedup.
- All 688 Rust tests pass; clippy `-D warnings`, fmt, and the R8 architecture cap all green.

## Review routing
Primary: **pdf-docx-generator** / **resume-export-expert** (export domain) · Secondary: **performance-profiler**. Part D-2 of the multi-PR plan.

### DIFF
--- a/apps/tauri/src-tauri/src/export/layout_pdf.rs
+++ b/apps/tauri/src-tauri/src/export/layout_pdf.rs
@@ -16,7 +16,8 @@ use anyhow::Result;
 use printpdf::*;
 
 use crate::export::pdf_renderer::{
-    build_line, link_annotation_op, load_all_fonts, resolve_fonts, rgb_to_color, LoadedFontSet,
+    build_line, collect_codepoints, link_annotation_op, load_all_fonts, resolve_fonts,
+    rgb_to_color, LoadedFontSet,
 };
 use crate::export::templates::Template;
 use crate::export::types::GenerationMeta;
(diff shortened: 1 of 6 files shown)


saeedkolivand:
🎉 This is included in version 0.36.3.

PR #207, opened 2026-06-01
feat: surface model reasoning across openai, gemini and ollama

saeedkolivand:
## Problem

Only Anthropic and the Claude-Code CLI agent surfaced their reasoning, so the existing `ThinkingBubble` never lit up for OpenAI, Gemini, or Ollama. The renderer plumbing was **already universal** (the normalized `thinking` stream flag) — the gap was **backend emission parity**.

## Fix — normalize reasoning at the provider boundary

Every adapter now emits reasoning as the same `ai:stream { delta, thinking: true }` shape, so the UI keeps one contract with **zero per-provider branching** and self-gates (the bubble only shows when reasoning text exists).

| Provider | Change |
|---|---|
| **OpenAI / compatible** | Emit `delta.reasoning_content` (DeepSeek-R1, vLLM, LM Studio, Ollama OpenAI shim) and `delta.reasoning` (OpenRouter). *o-series hide reasoning over Chat Completions — documented, not faked.* |
| **Gemini** | Request `thinkingConfig.includeThoughts` for thinking-capable models (**gated** — 1.5 / GA-2.0 reject it with a 400) and split `thought: true` parts. |
| **Ollama** | Emit structured `message.thinking` when present. We **do not** force the `think` flag (it 400s on non-thinking models) — capability-gated enablement rides with the `/api/show` work (Part A); tag-embedding models keep working via the splitter. |
| **Codex CLI** | Map `agent_reasoning*` events to the existing thinking channel. |
| **Gemini CLI** | No headless reasoning channel — left as-is (noted). |

## Renderer — centralize the inline `<think>` splitter

The inline `<think>…</think>` parser moves out of `generation.ts` into a shared `lib/generate/think-split.ts`, consumed by **both** the generation and analyze stream readers. This also **fixes analyze**, which previously only handled the structured flag and would leak raw `<think>` tags into the analysis JSON for tag-embedding local models.

## Tests

- Pure-parser unit tests: OpenAI reasoning split, Gemini `thought` split + thinking gate, Codex reasoning mapping.
- Shared-splitter tests: partial tags split across deltas, unclosed blocks (discarded), multi-block streams.
- Updated the analyze stream test to assert reassembled content (chunking-agnostic).
- Rust 695 tests + clippy + fmt; TS 26 tests + typecheck + lint:strict — all green.

## Review routing
Primary: **ai-provider-expert** · Secondary: **tauri-security-reviewer** (reasoning leakage), **frontend-reviewer** (splitter). Part B of the multi-PR plan.

### DIFF
--- a/apps/tauri/src-tauri/src/commands/ai_provider/cli_agent/codex.rs
+++ b/apps/tauri/src-tauri/src/commands/ai_provider/cli_agent/codex.rs
@@ -73,6 +73,12 @@ impl CliAgentBackend for CodexAgent {
             // Emit at message granularity (Codex's canonical assistant output).
             return text_of(m).map(CliEvent::Delta);
         }
+        if ty.contains("reasoning") {
+            // Codex streams chain-of-thought as `agent_reasoning*` events; surface
+            // it as thinking, like the cloud providers. Section-break markers carry
+            // no text and are dropped by `text_of`.
+            return text_of(m).map(CliEvent::Thinking);
+        }
         if ty.contains("task_complete") || ty.contains("turn_complete") {
             return Some(CliEvent::Done);
         }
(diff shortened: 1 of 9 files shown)


saeedkolivand:
🎉 This is included in version 0.37.0.

PR #209, opened 2026-06-01
feat: keep autopilot generation in a background store that survives navigation

saeedkolivand:
## Problem

The apply-job modal held its generation state (résumé / cover / reasoning) in **component state** and **aborted on close**, so closing the modal or navigating away threw away an in-flight or finished generation — website behaviour, not the desktop behaviour the app should have. The Rust job kept running; only the renderer dropped it.

## Fix — an app-wide generation store

A `useGenerationStore` (Zustand) keyed by a per-surface **context id** (e.g. `autopilot:<jobUrl>`). The analyze → résumé → cover flow runs as a **store action, decoupled from any component lifecycle**, so it finishes in the background and is restored when the modal reopens or the page is revisited.

- Session = the single source of truth: `{ phase, generating, resumeOut, coverOut, thinking, activeOut, meta, error }`.
- AbortControllers live in a module map (non-serializable, lifecycle-independent).
- `useTailorGeneration` becomes a **thin adapter** that selects the session and keeps only modal-local UI (copy/export).
- The modal's **close no longer aborts** — the Cancel button is the explicit abort.

## Scope

This PR lands the **store + the autopilot modal** (the reported "apply modal state loss" bug). Adopting the same store for the **AI Generate page** is a focused fast-follow (the page's generation state is spread across several hooks; the store is built generically so it drops in cleanly).

## Tests

- Store: runs analyze→resume→cover into a session that **survives a simulated remount**, isolates sessions per context id, guards concurrent runs, resets.
- The modal-hook tests now drive the store (same public API, background-safe).
- Full renderer suite: 151 tests pass; typecheck (`--force`) + lint:strict green.

## Review routing
Primary: **frontend-reviewer** · Secondary: **rust-backend-architect** (JobTracker reconcile — noted as follow-up), **performance-profiler** (per-token store writes). Part D-3 of the multi-PR plan.

### DIFF
--- a/apps/tauri/src/renderer/features/autopilot/components/ApplyJobModal/index.tsx
+++ b/apps/tauri/src/renderer/features/autopilot/components/ApplyJobModal/index.tsx
@@ -43,12 +43,19 @@ export function ApplyJobModal({ job, resumeText, onClose }: Props) {
   const hasDesc = jobDesc.length > 0;
   const fetchingDesc = !initialDesc && resolved.isLoading;
 
-  const gen = useTailorGeneration({ jobDesc, model, canUse, hasDesc });
+  // Per-job session key — generation lives in the store under this id, so closing
+  // and reopening the modal (or leaving the page) preserves the result.
+  const gen = useTailorGeneration({
+    contextId: `autopilot:${job.url}`,
+    jobDesc,
+    model,
+    canUse,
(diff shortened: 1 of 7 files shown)


saeedkolivand:
🎉 This is included in version 0.39.0.

PR #214, opened 2026-06-01
feat: persist application answers as a per-job application aggregate (D-1b)

saeedkolivand:
## D-1b — persist answers + brief on the application record

Completes the application-questions assistant: the drafted answers (and the company-research brief used) are now **stored on the `ai_generations` record** and shown in the Applications/History view, realizing the plan's "single application aggregate — one row per job".

### What changed
- **Schema** — additive `add_application_answers` migration: `application_answers` (JSON) + `company_brief` (TEXT). New `ApplicationAnswer { id, question, answer }`. Legacy rows default empty.
- **Per-job merge-upsert** — `save_application` merges an incoming save into the existing row for its `jobUrl` via the pure, unit-tested `merge_application` (takes each field only when non-empty). So the résumé/cover (tailor flow) and the answers/brief (questions assistant) — separate user actions — land on **one** row. Unlinked manual generations (no `jobUrl`) still insert separately. Backed by `find_by_job_url` + `update`.
- **Renderer** — the questions assistant persists its answers + brief through the save (merge-upsert by `jobUrl`) and invalidates the history + autopilot queries. The history card gains a **Company research** section and an expandable **Application answers** Q/A list.
- **Contract/IPC** — shared contract + Zod schema += `applicationAnswers` + `companyBrief`; `ai.rs` regenerated. en/de i18n.

### Architecture
- **Single application aggregate** — the record is now the full per-job application (résumé, cover, answers, brief, applied-state) instead of one row per generation. `applied_job_urls` (distinct `jobUrl`) is unaffected.
- **`merge_application` is pure** — re-running layers fields without clobbering; no field is lost when a lighter save (answers-only) arrives after a heavier one (cover).

### Tooling fix (bundled)
The new nested array pushed a generated `#[serde(default)]` fn signature past rustfmt's 100-col width, which `gen:ipc:check` (wants generator output) and `cargo fmt --check` (a pre-push gate) disagreed on. Fixed the **generator** (`gen-ipc-rust.ts`) to wrap long zero-arg signatures exactly as rustfmt does, so generated files stay clean under both gates.

### Tests
- Rust: record round-trips answers/brief; `merge_application` layers answers onto an existing cover without clobbering; `save_application` upserts one aggregate per `jobUrl` and inserts separately when unlinked; legacy rows default empty. (708 Rust tests pass.)
- Renderer: the hook drafts answers and saves them linked to the job url; no-op when nothing selected.

### Gates
`pnpm build:packages` · `lint:strict` · `format` · `typecheck --force` · `gen:ipc:check` · `cargo check/test/clippy` · `cargo fmt --check` — all ✅ (pre-push enforced).

### DIFF
--- a/apps/tauri/src-tauri/src/ai_generations/mod.rs
+++ b/apps/tauri/src-tauri/src/ai_generations/mod.rs
@@ -10,6 +10,14 @@ use crate::data_store::DataStore;
 use crate::db::{run_migrations, Migration};
 use crate::error::AppResult;
 
+/// One answered application question, stored on the application record.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ApplicationAnswer {
+    pub id: String,
+    pub question: String,
+    pub answer: String,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct AiGenerationRecord {
     pub id: String,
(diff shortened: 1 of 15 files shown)


saeedkolivand:
🎉 This is included in version 0.44.0.

PR #205, opened 2026-06-01
fix: factory reset now wipes every persistent store

saeedkolivand:
## What

Full factory reset (`privacy_reset_app`) previously left several persistent stores behind, so a "Reset app" did not return the user to a fresh-install state.

This change makes the reset clear **every** managed persistent store, keeping it in lockstep with `commands/data.rs::build_bundle` (the backup set) plus the stores intentionally excluded from backups (secrets, caches, the job log).

## Stores now wiped

- `JobTracker` (the job-execution log) — new `clear()`
- `AutopilotStore` — new `clear_all()`
- `JobPreferencesStore` — new `clear()`
- `ContactProfileStore` — new `clear()`
- `KvCache` (company-research / OCR cache) — new `clear()`
- All stored secrets (every AI/provider key incl. Brave, and board passwords) — driven off the credential **metadata index** via `CredentialStore::clear_all()` (no hardcoded provider list)

Already-cleared stores (documents, ai_generations, conversations, postings cache, interactions, board sessions) are unchanged.

## Future-proofing

- `CredentialStore::clear_all()` iterates the metadata index, so a newly-stored secret type is wiped automatically — no list to hand-maintain.
- A comment on `privacy_reset_app` ties the wipe-set to `build_bundle` so adding a new persistent store forces handling in both places.

## Tests

New unit tests (all `tempfile`-backed):
- `autopilot::test_clear_all_removes_every_autopilot`
- `job_preferences::test_clear_resets_to_empty`
- `pipeline::cache::clear_drops_every_entry`

`cargo test` for the touched modules: green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #217, opened 2026-06-01
refactor: drive full factory reset from a Resettable registry

saeedkolivand:
## Why

D-5 (#205) fixed the factory-reset bug but shipped a **hand-maintained `clear_all()` list** in `privacy_reset_app`, guarded only by a comment. That list silently rots: add a new persistent store, forget the clear line, and a "factory reset" leaves its data behind — the exact data-retention/GDPR bug reset exists to prevent. ADR-009 documented this and flagged the registry as the intended design. This PR implements it.

## What

- **`data_store.rs` (L0)** — a pure `Resettable { fn reset(&self) }` trait. Tauri-free so it stays below the shell layer (R2); a superset of `DataStore` that also covers secrets, caches, and the job log (which are excluded from backups).
- **`commands/privacy.rs` (L3)** — `Resettable` impls for each managed wrapper type (`Mutex<PostingsCache>`, `Arc<Mutex<AutopilotStore>>`, bare `DocumentStore`/`KvCache`, …), a `ResetRegistry` of type-erased reset closures, and **`manage_resettable(app, &mut reg, label, store)`** which manages a store *and* registers its reset in one call. `privacy_reset_app` now just iterates the registry — **zero per-store code**.
- **`main.rs::setup`** — persistent stores wired via `manage_resettable`; non-user state (updater, scraper engine) stays plain `manage`.

## The guarantee (and its honest limit)

- Registration is **type-checked**: `register::<T>()` requires `T: Resettable`, so any registered store provably clears. Adding a store via `manage_resettable` covers it on reset with **no edit to the reset command**.
- **Residual gap:** Tauri state can't be enumerated, so there's no compile-time proof that *every managed store* was registered — only that *registered* ones are `Resettable`. Mitigation: the `manage_resettable` convention (use it instead of bare `app.manage` for persistent stores) + tests + review. Documented in ADR-009.
- **Backups unchanged** — `data.rs::build_bundle` (the `DataStore` export set) remains a separate explicit list; this PR changes *reset* only.

## Tests
- The job-log impl clears via `.clear()` (not `.clear_all()` — guards the mapping); the cache impl clears; the registry records registrations in order. (711 Rust tests pass.)

## Docs
- **ADR-009 rewritten** (and renamed `…-lockstep-list` → `…-resettable-reset-registry`) to document the registry decision; `security-rules.md` + the knowledge README index updated to match.

## Gates
Full pre-push: build:packages · lint:strict · format · typecheck · `cargo check/test/clippy -D warnings`/`fmt --check` · architecture (R2/R7) — all ✅. No new top-level module (trait in L0, registry in L3).


saeedkolivand:
🎉 This is included in version 0.45.0.

PR #220, opened 2026-06-01
feat: native app menu via predefined roles with a working about dialog

saeedkolivand:


saeedkolivand:
🎉 This is included in version 0.46.0.

PR #221, opened 2026-06-01
feat: window-state, single-instance and notification plugins

saeedkolivand:


saeedkolivand:
🎉 This is included in version 0.47.0.

PR #224, opened 2026-06-02
feat: provider-native company research + Ollama Cloud provider

saeedkolivand:
## What & why

The cover-letter / application-answer **company research** feature was hardwired to a paid **Brave Search** key. This replaces it with **per-provider native web search** — every AI provider researches the company with its *own* engine — and removes Brave from the entire codebase. It also adds a first-class **Ollama Cloud** provider, unified by a single free Ollama account key.

## How it works now

| Active provider | How it searches |
| --- | --- |
| **Ollama (local) + Ollama Cloud** | Ollama **Web Search API** (`ollama.com/api/web_search`, free with an Ollama key) → synthesize |
| **OpenAI** | native `web_search` tool (Responses API) |
| **Anthropic** | native `web_search_20250305` server tool (Messages API) |
| **Gemini** | Grounding with Google Search (`google_search`) |
| **CLI agents** | their own built-in web tools (prompt-driven) |

- **Architecture:** a new `AiProvider::research()` capability (default no-op) + `Completer::research()`; one shared brief spec with two prompt shapes (`research.rs`). Ollama Cloud is a **composed** `OllamaCloudClient` (wraps the OpenAI client for chat, overrides `research()`) — no provider-id branching in generic clients.
- **Resilience/cost:** opt-in checkbox (default off), 7-day cache, **bounded timeout** in the enricher (never stalls generation), graceful degrade to an empty brief on any failure. The brief stays untrusted/fenced (ADR-010); research uses only the extracted company/role, never the raw job ad.
- **Application answers** now benefit too: the apply-modal toggle is **ungated** (works for application-questions-only flows), and the brief is broadened (mission/values/culture/news).
- **Checkbox** is always shown for every provider; a non-blocking hint nudges Ollama users to add the free key.
- **Ollama Cloud** appears in Settings → AI (registry-driven) and onboarding; the same key powers cloud chat + web search.
- New **Company Research** settings section + a dedicated, skippable **onboarding step**; en/de i18n.

## Tests / verification

- Rust: per-provider response parsers (Ollama/OpenAI-Responses/Anthropic/Gemini), `ProviderId` round-trip + `resolve(OllamaCloud)`, prompt builders. `cargo check` + `clippy -D warnings` + **722 tests** green.
- Frontend: `isOllamaFamily` + Ollama-Cloud registration tests; `typecheck` + `lint:strict` + **499 tests** green.
- Brave fully removed (`rg -i brave` → zero code hits; CHANGELOG history only).

[...]

PR #225, opened 2026-06-02
feat: render changelog and generated output as markdown

saeedkolivand:
## What & why

Two spots dumped raw text that should be styled. Both now reuse the existing **`MarkdownMessage`** renderer from `@ajh/ui` — **no new dependency**.

### 1. Changelog — Settings → Updates
The release body and "what's new" notes were rendered as raw `whitespace-pre-wrap`, so GitHub/semantic-release markdown (`### Features`, `* fix: …`) showed as unstyled text. Now rendered as styled headings/bold/lists.

### 2. Generated résumé / cover letter — AI Generate output
Adds a small **Edit / Preview** toggle:
- **Preview** (default) renders the output as prettified markdown — bold ATS keywords, bullets, headings.
- **Edit** keeps the raw textarea.

**Display-only and export-safe:** the raw string stays canonical — copy, export, and the `**keyword**` emphasis markers (which the export pipeline reads) are untouched, so switching views can never change the exported file. The streaming view stays raw live tokens; prettify applies only to the completed output.

## Tests
- New `OutputPanelDone` test: Preview hides raw `**` markers (renders `<strong>`); Edit shows the raw textarea with markers intact; switching views never calls `onOutputChange`.
- `typecheck` + `lint:strict` + targeted vitest green.

Follow-up to #224 (independent — UI-only).

### DIFF
--- a/apps/tauri/src/renderer/features/ai-generate/components/OutputPanelDone/OutputPanelDone.test.tsx
+++ b/apps/tauri/src/renderer/features/ai-generate/components/OutputPanelDone/OutputPanelDone.test.tsx
@@ -1,0 +1,55 @@
+import { describe, expect, it, vi } from 'vitest';
+import { fireEvent, render, screen } from '@testing-library/react';
+
+import type { TemplateId } from '@/lib/generate';
+
+import { OutputPanelDone } from './index';
+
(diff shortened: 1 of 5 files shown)


saeedkolivand:
🎉 This is included in version 0.49.0.

PR #227, opened 2026-06-02
fix: list full ollama cloud model catalog

saeedkolivand:
## Problem
In **Settings → AI**, the connected **Ollama Cloud** provider only offered `gpt-oss:20b` and `gpt-oss:120b` in its model dropdown (same in the AI-generate picker), even though Ollama Cloud serves many more hosted models (qwen3-coder, deepseek-v3.1, kimi-k2, glm-4.6, …).

## Root cause
Both pickers fetch via `ai_list_provider_models` → `OllamaCloudClient::list_models` → `OpenAiClient::list_models`. That filter passed everything **only** for `OpenAiCompatible`; for any other provider id it kept only `gpt-*`/`o1`/`o3`/`o4`/`chatgpt`. Ollama Cloud is `ProviderId::OllamaCloud`, so all non-`gpt-` models were silently dropped.

## Fix
Extract the predicate into a pure, tested helper `should_list_model(provider, id)` and invert the gate so **only native OpenAI** is restricted to chat families. Every other OpenAI-compatible backend (incl. Ollama Cloud, and any future composed provider) lists its full curated catalog — no per-provider coupling. Native OpenAI is still kept from flooding the picker with embeddings/audio/image models.

## Tests
- New unit test `list_filter_only_restricts_native_openai` covers both directions (native OpenAI drops non-chat ids; Ollama Cloud / OpenAI-compatible keep arbitrary names). Passes; crate compiles with 0 warnings.

## Verify
Settings → AI with an Ollama Cloud key → expand the row: the dropdown lists the full catalog; the same list appears in the AI-generate picker. Native OpenAI's picker is unchanged.

### DIFF
--- a/apps/tauri/src-tauri/src/commands/ai_provider/openai.rs
+++ b/apps/tauri/src-tauri/src/commands/ai_provider/openai.rs
@@ -39,6 +39,22 @@ fn join_responses_text(data: &Value) -> String {
         .unwrap_or_default()
 }
 
+/// Whether a model id returned by `/v1/models` should be offered in the picker.
+/// Native OpenAI exposes a large non-chat catalog (embeddings, audio, image,
+/// moderation…), so restrict it to chat-capable families. Every *other*
+/// OpenAI-compatible backend (custom gateways, Ollama Cloud, …) returns a curated
+/// catalog of its own models under arbitrary names, so pass those through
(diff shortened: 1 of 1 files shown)


saeedkolivand:
🎉 This is included in version 0.50.1.

PR #229, opened 2026-06-02
feat: inline option previews on ai generate with slide-fade animation

saeedkolivand:


saeedkolivand:
🎉 This is included in version 0.52.0.

PR #232, opened 2026-06-03
refactor: reactivate design-system eslint rules and clear all violations

saeedkolivand:
## Why

The design-system ESLint ruleset **and** the type-checked rules (`no-floating-promises`, `no-misused-promises`, `no-unnecessary-type-assertion`) were silently **inert**: their file globs targeted `apps/desktop` — a directory that no longer exists — so every rule matched zero files. Raw `<button>`/`<input>`, inline motion objects, `window.api`/react-query boundary leaks, and floating promises were never caught.

## What

**Config repair + hardening** (`eslint.config.mjs`)
- Retarget all globs `apps/desktop/src/renderer` → `apps/tauri/src/renderer`; fix the i18n adapter path (`lib/i18n/i18n.ts`).
- Fix a latent schema bug: the `@tanstack/react-query` restriction lived under `patterns` (which only accepts `group`) — moved to `paths`.
- Teach `RAW_INPUT` to honor its own documented exceptions (`range|file|checkbox|radio|hidden`). **12 of 13** flagged inputs were false positives the message already said were allowed.

**Cleared all 92 surfaced violations**
- **11** floating promises → `void`.
- **2** restricted imports → import the i18n instance from `@/i18n`; add a `useOpenDevtools` service hook (Ports & Adapters).
- **25** inline `transition={{…}}` → motion tokens. Added a `withDelay()` helper + `spinSlow`/`breathe`/`ping` ambient tokens so delayed/continuous animations stay in the token layer.
- **41** raw `<button>` → `<Button>` and **1** raw `<input>` → `<Input>`, via a new **`unstyled`** variant. It routes custom surfaces (segmented controls, icon toggles, inline links, clickable cards) through the primitive for consistent focus-visible + disabled handling **without injecting chrome** — so each call site keeps its exact look (className preserved verbatim).

**Tooling-discrepancy fix**
- Reverted spurious `no-unnecessary-type-assertion` autofixes that removed load-bearing assertions (e.g. `match.data as (MatchScore & { error? })`, `'' as string | undefined`, `as HTMLTextAreaElement`) and broke `tsc`. Replaced with explicit type annotations / generics so ESLint and `tsc` agree.

## Verification

- `pnpm lint:strict` → **0** (was 92).
- `pnpm typecheck` → **7/7 tasks**.
- `@ajh/ui` tests **124 + 3 new** (unstyled Button/Input + `withDelay`); affected `apps/tauri` suites green.
- Button/input conversions are **className-faithful by construction** (same element, same classes; `unstyled` injects no visual chrome — covered by a test). A visual spot-check of high-conversion screens (onboarding, AI settings) is a good optional follow-up.

### DIFF
--- a/apps/tauri/src-tauri/Cargo.lock
+++ b/apps/tauri/src-tauri/Cargo.lock
@@ -61,7 +61,7 @@ dependencies = [
 
 [[package]]
 name = "ajh-tauri"
-version = "0.51.0"
+version = "0.52.0"
 dependencies = [
  "anyhow",
  "apple-native-keyring-store",
--- a/apps/tauri/src/renderer/components/layout/Sidebar/index.tsx
+++ b/apps/tauri/src/renderer/components/layout/Sidebar/index.tsx
@@ -17,7 +17,7 @@ import { AnimatePresence, motion } from 'motion/react';
 import { useRef, useState } from 'react';
 import { Link, useRouterState } from '@tanstack/react-router';
 
-import { cn, transition, variants } from '@ajh/ui';
+import { Button, cn, transition, variants } from '@ajh/ui';
 
 import { ROUTES } from '@/constants/routes';
 import { getTimeGreeting } from '@/lib/greeting';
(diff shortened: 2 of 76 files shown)


saeedkolivand:
🎉 This is included in version 0.53.0.

PR #233, opened 2026-06-03
fix: correct invalid bg-brand/08 opacity and guard it in eslint

saeedkolivand:
## Why
`bg-brand/08` is an **invalid Tailwind opacity** (leading zero) — it's dropped at build, so the element gets **no background**. The selected Autopilot **schedule** and **action** wizard cards therefore rendered with no fill (invisible selection). No lint rule caught it. _(UX audit §3.1 — the one concrete consistency defect.)_

## What
- `bg-brand/08` → `bg-brand/10` in `wizard-steps/StepSchedule` + `StepAction`.
- New ESLint `no-restricted-syntax` guard (`/0<digit>` in `className`) registered in both the base (warn) and UI-layer (error) blocks, so this class of typo can't silently ship again.
- gitignore `docs/assets/ux-audit/` — UX-audit review screenshots are large binaries and should never be committed.

## Verify
- Guard fires on a planted `bg-brand/08` (confirmed) and matches **nothing** valid across the repo.
- `pnpm lint:strict` → 0.

First of the UX-audit §12 backlog (full program).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,9 @@ playwright/.cache/
 test-results/
 vitest.config.*.timestamp-*
 
+# UX-audit screenshots — captured locally for review only; large binaries, never committed
+docs/assets/ux-audit/
+
 # ── OS files ───────────────────────────────────────────────────────────────────
 # macOS
 .DS_Store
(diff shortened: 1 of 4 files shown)


saeedkolivand:
🎉 This is included in version 0.53.1.

PR #234, opened 2026-06-03
fix: restore visible focus rings and add dialog + toggle aria semantics

saeedkolivand:
## Why
`:focus-visible { outline: none }` was set **globally** with no compensating ring, so any element that doesn't bring its own ring (links, `<label>`s, `div[role=button]` rows) focused **invisibly** — a WCAG 2.4.7 failure. Inputs/textareas had their focus indicator explicitly zeroed. Dialogs had no accessible name; the ATS/debug toggles had no switch semantics. _(UX audit §4.)_

## What
- **Global focus ring:** `:focus-visible` now draws a visible `2px solid var(--color-brand)` outline. Components that own a ring already pair `focus-visible:outline-none` (higher specificity than a bare `:focus-visible`), so they **opt out automatically — no double ring** (verified visually).
- **Inputs + textareas:** `.input-field:focus-visible` now shows a brand ring (was `box-shadow: none`), covering `<Input>` and `<TextArea>` centrally.
- **Dialogs:** `ModalShell` gains `ariaLabelledby`/`ariaLabel` props; `ConfirmModal` wires its title via `useId()` → `aria-labelledby`.
- **Toggles:** `role="switch"` + `aria-checked` on the ATS-mode and debug-mode toggles.

## Verify
- `pnpm lint:strict` → 0 · `typecheck` → 7/7 · `@ajh/ui` tests → 129 pass.
- Keyboard-focus capture confirms the ring renders and primitives don't double-ring.

UX-audit §12 backlog, PR 2/13. Independent of #233.

### DIFF
--- a/apps/tauri/src/renderer/features/ai-generate/components/GenerationConfig/index.tsx
+++ b/apps/tauri/src/renderer/features/ai-generate/components/GenerationConfig/index.tsx
@@ -208,6 +208,8 @@ export function GenerationConfig({
         <Button
           variant="unstyled"
           type="button"
+          role="switch"
+          aria-checked={atsMode}
           onClick={() => onAtsModeChange(!atsMode)}
           className={cn(
             'w-full flex items-center justify-between rounded-lg border px-3 py-2.5 transition-all text-left',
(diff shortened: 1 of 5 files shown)


saeedkolivand:
🎉 This is included in version 0.53.2.

PR #238, opened 2026-06-03
perf: optimistically delete generations and autopilots with rollback

saeedkolivand:
## What
PR 5 — optimistic updates. Both delete hooks used `onSuccess → invalidateQueries`, so the deleted card sat on screen (with a spinner) until the SQLite round-trip returned. Now they update the cache **immediately**:

- `useRemoveAiGeneration` and `useRemoveAutopilot` gain `onMutate` (cancel in-flight, **snapshot** the list, optimistically filter the item out) + `onError` (restore the snapshot) + `onSettled` (invalidate to reconcile with the backend).

## Why
The card disappears the instant you confirm the delete; if the backend genuinely fails, it reappears and the error surfaces — no perceived lag for the common (successful) case.

## Tests
Each hook gains a test that drives a *deferred* backend call and asserts:
1. the item leaves the cache **before** the backend resolves (optimism), and
2. it is **restored** when the backend rejects (rollback).

## Scope note
The audit's PR 5 also mentioned "bookmark/applied counters" and "board move":
- **PostingRow** already updates its applied/viewed/saved badges via local `useState` (instant) — no cache change needed.
- There is **no drag-and-drop job board** (DnD is explicitly out of scope per the audit), so "board move" doesn't apply.

So the genuinely clean, high-value optimistic targets were the two deletes.

## Verification
`pnpm typecheck` ✅ · `pnpm lint:strict` ✅ · both service-hook suites ✅ (smoke + new optimistic/rollback tests).

### DIFF
--- a/apps/tauri/src/renderer/services/use-ai-generations/use-ai-generations.test.ts
+++ b/apps/tauri/src/renderer/services/use-ai-generations/use-ai-generations.test.ts
@@ -1,11 +1,53 @@
-import { describe, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
+import { QueryClient } from '@tanstack/react-query';
+import { act, waitFor } from '@testing-library/react';
 
-import { exerciseServiceHooks } from '@/test-support';
+import { createMockClient, exerciseServiceHooks, renderHookWithClient } from '@/test-support';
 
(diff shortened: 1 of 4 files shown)


saeedkolivand:
🎉 This is included in version 0.55.1.

PR #241, opened 2026-06-03
feat: drop auto-apply engine; reposition autopilot as an apply assistant

saeedkolivand:
## Why

Auto-apply (the browser-automation apply engine) was costly to maintain — selector drift, captcha, per-board form logic — and was already **non-functional / "Coming Soon"** in the UI, so **nothing user-facing regresses**. This reframes the product as an apply **assistant**: Autopilot **finds → ranks → notifies**; the user tailors documents and applies themselves.

## Removed (engine)

- **Rust** `applying/` in full (6 board appliers, `captcha_handler`, `form_filler`, `selectors`, the `APPLIERS` registry), `apply_helpers/`, `commands/apply.rs`, the apply IPC contract, and the `From<ApplyError>` error impl. `chromiumoxide` stays (scraping uses it). The two apply commands are dropped from the invoke handler (IPC attack surface ↓).
- **FE** `ApplyDrawer/**`, the `use-apply` service + `apply` query keys + client `apply` namespace + mock entry. `useJobEvents`/`useCancelJob` are shared and kept.
- **Autopilot auto-apply surface** — the `action` field + `auto_apply`/`review` + `autoSubmit` across the Zod schema, generated IPC, Rust store, shared types, and the wizard. Saved autopilots **migrate silently** on load (serde drops the dead keys; behaviour is uniformly find-&-save).

## Changed (assistant)

- **`StepAction`** → an informational step (find → notify → tailor) + an optional **base cover letter** the assistant tailors per job. The kept `coverLetter` field threads through `ApplyJobModal` → `generateCoverLetter` as the starting point.
- **PostingRow "Apply" → "Tailor"** — seeds the AI Generate workspace with the job, opens the posting, and marks it applied; available for **every board** (tailoring is board-agnostic), not just the old appliable set.
- **AutopilotCard** drops the `action` pill, the permanent `Applied {N}` counter, and the apply step-icons; the run state machine drops the `applying` state.

## Kept

Scraping, find/score/`minMatchScore`, found-jobs list, the `ApplyJobModal` tailor assistant, AI Generate, Analyze, cover-letter generation, derived "Applied" tracking, the "Autopilot" name (copy clarified).

## Docs

Scrubbed auto-apply from **README, API.md, ARCHITECTURE.md, PATTERNS.md, architecture-rules.md, and the knowledge base**; `tests/architecture.rs` updated (`applying` + `apply_helpers` removed from the layer model + allowlists). Landing page left untouched (owner decision).

**Follow-up:** the two interactive HTML architecture-visualization docs (`docs/architecture-map.html`, `docs/architecture/how-it-works.html`) still depict the applier subsystem in their graph data — best regenerated/rewritten in a focused docs pass rather than risk dangling node/edge refs here.

## Verification

`pnpm typecheck` ✓ · `pnpm lint:strict` ✓ · 221 renderer vitest ✓ · 80 shared vitest ✓ · `cargo test` (756) ✓ incl. the updated architecture suite · `cargo clippy -D warnings` ✓.


saeedkolivand:
🎉 This is included in version 0.58.0.

PR #242, opened 2026-06-03
feat: autopilot new-jobs notifications + tray surface + deep-link focus guard

saeedkolivand:
## Why

Autopilot is a background discovery agent — but a scheduled run that surfaced new jobs while you were away was **invisible** (no notification was ever sent, the tray was Show/Quit only). This makes the "notify" half real: find → rank → **notify → click → jump to the finds**.

Per owner decision, the **OS-registered `ajh://` scheme is deferred** to a follow-up (it adds `tauri-plugin-deep-link` + platform-specific dev/prod behavior). This PR lands everything else **plus** the argv security guard, so the deep-link follow-up is a thin, pre-hardened add.

## What

**Notifications** — `record_run` now returns the count of newly-surfaced postings (never-seen URLs). After a run, that drives a **permission-gated** OS notification (request once when not granted; skip on deny).

**Tray** (new L3 `tray` module) — dynamic **"New jobs: N"** (click → focus window + emit `autopilot.focus` to that run's autopilot, reset counter) + **"Pause all autopilots"**, alongside Show/Quit.

**Renderer focus flow** — new `autopilot.focus` event (contract + client + mock + service hook) → an app-global hook (`useAutopilotFocusNavigation`, mounted in `__root`) that invalidates the autopilot list and, for a non-empty id, routes to `/autopilot` and auto-expands + scrolls that card's found-jobs. Empty id = pure refresh (used by tray Pause-All so the renderer reflects it).

## Security (tauri-security-reviewer lens)

The single-instance handler previously **ignored argv**. It now validates argv via the new L3 `deeplink` module against a **strict route allowlist** — only `ajh://autopilot/<id>` with a syntactically valid id (`[A-Za-z0-9_-]{1,64}`) navigates. Unknown schemes/actions, path traversal (`ajh://autopilot/../../settings`), query/fragment, separators, and oversized ids are **denied** (navigate nowhere). Unit-tested with injection cases. The OS scheme is not registered yet, so no external URL can reach this path today — the guard lands first by design.

## Tests

- Rust: `deeplink::parse_focus_target` (valid / unknown scheme+action / path-traversal+injection / normal-launch / found-among-args); `record_run` new-count (first run, re-run dedup, none-new, unknown id).
- Arch test + `architecture-rules.md` classify `tray` + `deeplink` as L3.

## Verification

`pnpm typecheck` ✓ · `pnpm lint:strict` ✓ · `pnpm gen:ipc:check` ✓ · 80 shared + 223 renderer vitest ✓ · `cargo clippy -D warnings` ✓ · `cargo test` (763) ✓.

## Follow-ups (tracked)

- OS `ajh://` scheme via `tauri-plugin-deep-link` (scheme config + capabilities) — guard already in place.
- Notification/tray strings are English (matches the existing English-only tray menu); localize later.
- The two interactive HTML arch-viz docs still depict the removed applier subsystem (from #241).

### DIFF
--- a/apps/tauri/src-tauri/src/autopilot/mod.rs
+++ b/apps/tauri/src-tauri/src/autopilot/mod.rs
@@ -238,23 +238,29 @@ impl AutopilotStore {
     /// list **merged** with prior runs by URL — so re-running keeps history
     /// (first-seen + any state) instead of replacing it, and genuinely new
     /// postings are flagged `is_new`.
+    /// Returns the number of **newly surfaced** jobs in this run (postings whose
+    /// URL was never seen before) — drives the "N new jobs" notification + tray.
     pub fn record_run(
         &self,
         id: &str,
         total_found: u32,
         total_applied: u32,
         found_jobs: Vec<FoundJob>,
-    ) {
+    ) -> u32 {
         let mut map = self.load();
+        let mut new_count = 0u32;
         if let Some(ap) = map.get_mut(id) {
             let now = now_ms();
             ap.total_found = total_found;
             ap.total_applied = total_applied;
             ap.found_jobs = merge_found_jobs(&ap.found_jobs, found_jobs);
+            // `merge_found_jobs` flags only never-before-seen URLs as `is_new`.
+            new_count = ap.found_jobs.iter().filter(|j| j.is_new).count() as u32;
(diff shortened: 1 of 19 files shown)


saeedkolivand:
🎉 This is included in version 0.59.0.

PR #243, opened 2026-06-03
feat: register the ajh:// deep-link scheme through the validated guard

saeedkolivand:
## Why

Completes the deep-link feature flagged as a follow-up in #242. External `ajh://autopilot/<id>` URLs now open/focus the app and jump to that autopilot's found-jobs. #242 landed the **security guard** first; this PR only registers the OS scheme and routes URLs through that already-validated entry point.

## What

- Add `tauri-plugin-deep-link`; register the `ajh` scheme in `tauri.conf.json` (installer-time registration on Windows/macOS) + `register_all()` at startup (runtime registration for Linux + `pnpm dev`).
- `on_open_url` (macOS + runtime-registered schemes) funnels every incoming URL through the **same** strict allowlist as the single-instance argv path (`deeplink::parse_focus_target`) before any navigation, then focuses the window + emits `autopilot.focus`. Windows/Linux relaunch already routes via the single-instance guard from #242, so both delivery paths share one validated entry point.

## Security

No new capability — deep links are handled entirely in Rust (no JS plugin API), so the IPC surface is unchanged. Only `ajh://autopilot/<id>` with a valid id navigates; everything else is denied (guard + tests from #242). `cargo deny` clean (advisories / licenses / bans / sources).

## Verification

`cargo check` · `cargo clippy -D warnings` · `cargo fmt --check` · `cargo test` (763) · `cargo deny check` — all green. No FE/TS changes.

## Note

This wires **external URL → app**. Clicking the OS *notification* itself is separate (see PR discussion).

### DIFF
--- a/apps/tauri/src-tauri/Cargo.lock
+++ b/apps/tauri/src-tauri/Cargo.lock
@@ -61,7 +61,7 @@ dependencies = [
 
 [[package]]
 name = "ajh-tauri"
-version = "0.52.0"
+version = "0.59.0"
 dependencies = [
  "anyhow",
  "apple-native-keyring-store",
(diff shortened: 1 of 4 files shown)


saeedkolivand:
🎉 This is included in version 0.60.0.

PR #244, opened 2026-06-03
feat: open the autopilot page when the new-jobs notification is clicked

saeedkolivand:
## Why

The "new jobs" OS notification shipped in #242 was **alert-only** — clicking it didn't take you anywhere. This makes the natural gesture work: **click the notification → open the autopilot page** to act on the finds.

## What

- Add the **`@tauri-apps/plugin-notification`** JS package and route its `onAction` (notification body-click) through the **AppClient adapter** as `autopilot.onNotificationClick` — keeping the renderer off direct plugin APIs (ports-and-adapters rule).
- An app-global listener (folded into the existing `useAutopilotFocusNavigation` hook, mounted in `__root`) invalidates the autopilot list and navigates to `/autopilot` on click. Autopilot is the only notification source today, so any click maps there; the run's card already shows its **New** badge.
- Capability: the **minimal** `notification:allow-register-listener` (the command `onAction` registers behind) — not the broad `notification:default`.

## Platform reality (as discussed)

Body-click → `onAction` is reliable on **macOS** and **packaged Windows**; best-effort on Linux (depends on the notification daemon), and toast activation needs the installer-registered AUMID so it's not testable in `pnpm dev`. The **tray "New jobs: N"** item and the **`ajh://` deep link** (#243) remain the fully-reliable jump paths on every platform.

## Verification

`pnpm typecheck` · `pnpm lint:strict` · 80 shared + 223 renderer vitest — all green. No Rust changes (the notification is already sent Rust-side; this is the click handler). Independent of #243 (no file overlap).

### DIFF
--- a/apps/tauri/package.json
+++ b/apps/tauri/package.json
@@ -23,6 +23,7 @@
     "@tanstack/react-query": "^5.100.14",
     "@tanstack/react-router": "^1.170.8",
     "@tauri-apps/api": "^2.11.0",
+    "@tauri-apps/plugin-notification": "^2.3.1",
     "clsx": "^2.1.1",
     "i18next": "^26.3.0",
     "i18next-browser-languagedetector": "^8.2.1",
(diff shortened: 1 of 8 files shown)


saeedkolivand:
🎉 This is included in version 0.60.0.

PR #245, opened 2026-06-03
fix: make autopilot scheduling, scoring, and cancellation reliable

saeedkolivand:
First slice of the plan's **PR 9 (Autopilot catch-up + reliability)**. PR 9 bundled six items spanning Rust + IPC + FE + two new dependencies; splitting so each lands as a reviewable, correctly-typed PR (this is the pure `fix:` core — no new deps, no IPC/FE).

## What

Three reliability fixes to the autopilot run path:

1. **Startup catch-up.** The scheduler consumed its immediate first tick and only checked due autopilots after a full `TICK_INTERVAL_SECS` (60s). It now runs one catch-up sweep ~5s after launch (grace period for startup to settle), then enters the minute loop — so autopilots that fell overdue while the app was closed fire promptly.

2. **`minMatchScore` is now enforced.** It was dead config: every scraped posting was recorded regardless of score. Postings that *can* be scored against the resume are now dropped below the threshold before `record_run`. Unscored postings (no resume set, or no description to compare) are always kept — the gate only filters jobs we could actually rank. The `rank_done` step reports kept/dropped counts, and the "new jobs" notification now reflects real matches.

3. **Cancellation spans the whole run.** `autopilot_run` registered a token, but `scrape_board` overwrote the slot with its own, orphaning it. The engine now *reuses* a caller-registered token (`entry().or_insert_with`) instead of overwriting it, so a tray/UI cancel reaches the run's own token; the run checks it and bails before recording or notifying for an aborted run.

## Tests

- scheduler due-logic: `interval_secs` mapping; `is_due` for manual / paused / archived / never-run / interval-elapsed.
- score gate: at/above/below threshold (inclusive boundary) + unscored-always-kept.
- engine: a cancel reaches a pre-registered token (the invariant the run relies on).

Green locally: clippy `-D warnings` clean, full bin suite 759 passed / 2 ignored, architecture tests 11 passed.

## Follow-ups (rest of PR 9)

- **9b** (`feat:`) — `run_status` + amber "interrupted" badge (IPC + FE).
- **9c** (`feat:`) — battery-aware pause + optional launch-at-login (new deps → security review).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -149,6 +149,7 @@ jobs:
             package.json \
             apps/tauri/package.json \
             apps/tauri/src-tauri/Cargo.toml \
+            apps/tauri/src-tauri/Cargo.lock \
             apps/tauri/src-tauri/tauri.conf.json \
             README.md; then
             echo "Version files are already synced."
@@ -159,6 +160,7 @@ jobs:
             package.json \
             apps/tauri/package.json \
             apps/tauri/src-tauri/Cargo.toml \
+            apps/tauri/src-tauri/Cargo.lock \
             apps/tauri/src-tauri/tauri.conf.json \
             README.md
           git commit -m "chore: sync version files to v${{ needs.release.outputs.version }} [skip ci]"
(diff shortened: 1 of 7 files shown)


saeedkolivand:
🎉 This is included in version 0.60.1.

PR #246, opened 2026-06-03
feat: track autopilot run status and flag interrupted runs

saeedkolivand:
Second slice of the plan's **PR 9 (Autopilot reliability)** — the observable-status half (`9a` was the pure `fix:` core, merged in #245).

## What

Each autopilot now records the outcome of its most recent run — `inProgress | completed | failed | interrupted` — so the UI can show an honest indicator instead of silently looking idle after a failed or cut-off run.

- **Transitions** (`commands/autopilot.rs`): set `inProgress` at run start, `failed` on a scrape error, `completed` on success (`record_run`) or user cancel.
- **Crash reconciliation** (`autopilot_scheduler.rs`): at startup the scheduler flips any run left `inProgress` (the app closed/crashed mid-run) to `interrupted` — synchronously, *before* the first catch-up sweep could start a new run.
- **Badge** (`AutopilotCard`): amber **Interrupted** / red **Failed** chip when the last run didn't finish cleanly; hidden while a run is live (the status dot + step log already cover "running").

## Notes

- The full `Autopilot` record is hand-typed in `@ajh/shared` and returned raw from Rust (no Zod validation on the response), so this adds `runStatus?` to the **type only** — no `gen:ipc` regeneration.
- `runStatus` is optional and serde-`skip_serializing_if`-None, so existing saved autopilots load unchanged (absent = never run).
- en/de copy added (`autopilot.badge.failed` / `.interrupted`).

## Tests

Store reconciliation + transitions: `mark_interrupted_runs` flips only `inProgress` (and is idempotent); `record_run` marks `completed`.

Green locally: clippy `-D warnings` clean, 28 autopilot Rust tests, typecheck + lint:strict clean, 223 renderer + 80 shared vitest.

## Follow-up (last PR 9 slice)

- **9c** (`feat:`) — battery-aware pause + optional launch-at-login (new deps → security review).

### DIFF
--- a/apps/tauri/src-tauri/src/autopilot/mod.rs
+++ b/apps/tauri/src-tauri/src/autopilot/mod.rs
@@ -84,6 +84,22 @@ pub enum AutopilotStatus {
     Archived,
 }
 
+/// Outcome of the most recent run. Distinct from [`AutopilotStatus`] (the
+/// agent's enabled/paused lifecycle): this tracks a single run so the UI can
+/// show a live/failed/interrupted indicator.
+///
+/// `Interrupted` is not set by a run — it's reconciled at startup from a run
+/// left `InProgress` when the app closed or crashed mid-run (see
+/// [`AutopilotStore::mark_interrupted_runs`]).
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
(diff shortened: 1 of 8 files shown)


saeedkolivand:
🎉 This is included in version 0.61.0.

PR #247, opened 2026-06-03
feat: add opt-in launch-at-login for background autopilot

saeedkolivand:
Final slice of the plan's **PR 9 (Autopilot reliability)** — `9a` (#245) and `9b` (#246) shipped the scheduler/scoring/cancellation fixes and run-status badge. After review (`go with pr 9`), the battery-aware-pause item was **dropped**: `sysinfo` (already a dep) has no power API, so it would have needed a *new* native power crate for modest value (scrapes are light network I/O) — the highest supply-chain risk in the backlog. This PR ships only the clearly-valuable, well-vetted half.

## What

Autopilot is a background discovery agent, so let users opt into launching the app at sign-in to keep finding jobs without opening it manually.

- registers the official **`tauri-plugin-autostart`** (after single-instance, so a login launch focuses the existing window rather than spawning a duplicate; **no launch args**).
- exposes `system_get_launch_at_login` / `system_set_launch_at_login` Rust commands via the plugin's autolaunch manager; `set` returns the *resulting* OS state so the UI reflects what was actually applied.
- a new **Startup** card in General settings with a `role="switch"` toggle, sourced from the OS via React Query (not local prefs). en/de copy.

## Security (tauri-security-reviewer lens)

- **Default OFF**, opt-in only.
- Toggled exclusively through our own `system_*` IPC commands (Rust `app.autolaunch()`), **not** the plugin's JS commands — so **no new capability** is granted and the IPC attack surface is unchanged.
- No arbitrary launch arguments registered (`None`), so nothing user-controlled feeds the autostart entry.
- New deps (`tauri-plugin-autostart` + transitive `auto-launch`) pass `cargo deny`: **advisories / bans / licenses / sources all ok** (duplicate-crate entries are non-blocking `warn`).

## Tests

`useLaunchAtLogin` reads OS state; `useSetLaunchAtLogin` toggles via the client and resolves to the applied state. (The Rust commands are thin passthroughs to the vetted plugin.)

Green locally: clippy `-D warnings` clean, Rust 761 passed / 2 ignored, typecheck + lint:strict clean, 225 renderer + shared vitest, cargo-deny clean.

This **closes out PR 9.**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,12 +3,19 @@ name: 🚀 Release
 # ──────────────────────────────────────────────────────────────────────
 # Release Pipeline
 # ──────────────────────────────────────────────────────────────────────
-# Purpose: Automated semantic release and cross-platform Tauri app builds
-# Triggers: push to main, manual workflow_dispatch
+# Purpose: Automated semantic release; cross-platform Tauri builds on demand
+# Triggers:
+#   - push to main      → release + sync-version-files (NO installer build)
+#   - workflow_dispatch → build: compile + attach installers for a release
 # Jobs:
-#   - release: Determine version and create GitHub release
-#   - sync-version-files: Commit tracked version files for the published release
-#   - build: Build Tauri apps for Windows, Linux, and macOS
+#   - release: Determine version and create GitHub release (push only)
+#   - sync-version-files: Commit tracked version files for the release (push only)
+#   - build: Build Tauri apps for Windows, Linux, and macOS (manual only)
+#
+# Rationale: a 3-platform Tauri build is ~60 min/runner each, so it no longer
+# runs on every merge. semantic-release still versions + drafts release notes on
+# push; click "Run workflow" in the Actions tab to build + attach the installers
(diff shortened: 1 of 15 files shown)


saeedkolivand:
Added a second commit (`ci:`) at the owner's request — **decouples the Tauri installer build from every merge.**

- **push to `main`** → `release` + `sync-version-files` still run automatically (tag + release notes + version-file sync), but **no longer build installers**.
- **manual** → the `build` + `generate-update-manifest` jobs are now `workflow_dispatch` only. From **Actions ▸ "🚀 Release" ▸ "Run workflow"**, optionally pass a `version` (blank = latest tag) to build + attach the installers and `latest.json`.

Why: a 3-platform Tauri build is ~60 min/runner each — wasteful on every merge. The version input is passed via env (no script injection) and regex-validated before it's used as a checkout ref. `build` exposes the resolved version as a job output so the manifest job reuses the exact same one. `docs/DEPLOYMENT.md` updated to the two-phase pipeline.

> Note: after this merges, releases won't have installer assets until you click **Run workflow** for that version.

saeedkolivand:
🎉 This is included in version 0.62.0.

PR #248, opened 2026-06-03
ci: also auto-build installers on release, keep the manual button

saeedkolivand:
Follow-up to #247's manual-dispatch build (owner: *"I'd rather **also** have the manual button"*). Restores the **automatic** installer build on a release push while **keeping** the manual trigger.

## Change

`build` (and the `generate-update-manifest` job that follows it) now run on **both** paths:

```yaml
needs: sync-version-files
if: >-
  always() &&
  (github.event_name == 'workflow_dispatch' ||
   needs.sync-version-files.outputs.released == 'true')
```

- **Release push** → `release` → `sync-version-files` → **build** → `generate-update-manifest`, fully automatic (as it was before #247).
- **Manual** (Actions ▸ 🚀 Release ▸ Run workflow) → still builds a chosen tag on demand.

## Why it's correct

- `needs: sync-version-files` makes the **push** path wait until semantic-release has created the tag before the resolve step picks "latest tag".
- `if: always()` lets the **manual** path run even though `sync-version-files` is skipped on `workflow_dispatch` (release is push-only).
- `build` exposes the resolved version as a job output, so `generate-update-manifest` targets the exact same release on either path.
- On a push with **no** release commit, `sync-version-files` is skipped → `build`'s condition is false → build + manifest skip (no wasted runners).

Manual dispatch now means **"rebuild an existing tag"** (a runner flaked, re-attach assets) rather than a separate release path. `docs/DEPLOYMENT.md` updated to match.

YAML validated; job graph confirmed (release/sync push-only; build/manifest = release-push **or** manual).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,19 +3,19 @@ name: 🚀 Release
 # ──────────────────────────────────────────────────────────────────────
 # Release Pipeline
 # ──────────────────────────────────────────────────────────────────────
-# Purpose: Automated semantic release; cross-platform Tauri builds on demand
+# Purpose: Automated semantic release + cross-platform Tauri installer builds
 # Triggers:
-#   - push to main      → release + sync-version-files (NO installer build)
-#   - workflow_dispatch → build: compile + attach installers for a release
+#   - push to main      → release + sync-version-files, then auto-build installers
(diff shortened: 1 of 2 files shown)


saeedkolivand:
🎉 This is included in version 0.62.1.

PR #249, opened 2026-06-03
fix: keyboard-navigable settings sidebar + remove dead board-credential code

saeedkolivand:
First slice of **PR 10 (settings consistency)** — the a11y + dead-code-removal piece. Two commits:

## 1. `fix:` settings sidebar keyboard a11y

The settings section list used `<div role="button">` with only an `onClick`, so keyboard users couldn't focus or activate sections. Switched to the `@ajh/ui` `Button` (a real focusable `<button>` that picks up the global focus ring) with `aria-current` on the active section. Also removed two dead components — `AccountRow` (credential username/password row) and `LinkedInSessionRow` — both superseded by the session-auth `BoardSessionRow`, referenced only by the barrel, with no consumers/tests/stories.

## 2. `refactor:` remove the dead board-credential code end-to-end

Deleting `AccountRow` exposed a fully-dead backend: it was the only writer of the board username/password store, and **no scraper reads board passwords** (the appliers that did were removed in PR 7). Boards authenticate via **browser sessions** (`boards.connect` / `linkedin.connect`). The credential `set/get/remove/list` path was dead end-to-end (the FE client even passed a mismatched payload), so it's removed across every layer:

- **Rust**: drop `credentials_set/get/remove/list` commands + their invoke-handler registrations; drop the unused `CredentialStore::get_all_decrypted`.
- **shared**: drop `list/set/remove` from the credentials contract + channels, and the orphaned `CredentialMetadata` type.
- **renderer**: drop the client `list/set/remove`, the mock stubs, the `useCredentials`/`useSetCredential`/`useRemoveCredential` hooks, and the `CredEntry` type.

**Kept (still live):** `CredentialStore` itself — AI provider keys (`ai:*`) and the factory reset use it — and `credentials_available` (gates the encryption-availability warning).

### ✅ Account connect/disconnect is unaffected
The Accounts page renders `BoardSessionRow`, which uses the session-auth path (`boards.*` / `linkedin.*`) — a completely separate IPC namespace, untouched here. Removing the unused commands also **shrinks the IPC attack surface**.

## Verification
clippy `-D warnings` clean (confirms no dead `CredentialStore` methods remain), Rust 761 passed / 2 ignored, typecheck (tauri + shared) clean, lint:strict clean, 225 renderer + 80 shared vitest.


saeedkolivand:
🎉 This is included in version 0.62.1.

PR #253, opened 2026-06-03
ci: build installers only on manual dispatch, not on every release

saeedkolivand:
## Why

Since #248 the installer build auto-fired on **every** release push — a ~20-min cross-platform Tauri build after almost every `feat:`/`fix:` merge. During rapid PR cadence that's mostly unwanted, and the build job's `cargo` step doesn't respond to the normal "Cancel run" (it has to be **force-cancelled** each time).

## Change

- **`build` job → `if: github.event_name == 'workflow_dispatch'`** (was `always() && (workflow_dispatch || released == 'true')`). Dropped its `needs: sync-version-files` — the build already self-syncs version files and checks out the resolved tag, so it depends on nothing from the push-only jobs.
- **Push path unchanged and still ordered:** `release` (create GitHub release + tag) → `sync-version-files` (commit the version bump). Only the slow installer build is decoupled.
- Installers are built **on demand** from **Actions ▸ "🚀 Release" ▸ "Run workflow"** (blank version = latest tag; or pass one).
- `generate-update-manifest` (`needs: build`) naturally only runs with a manual build.

Reverts the auto-build half of #248; keeps the manual button from #247. Docs (`DEPLOYMENT.md` prose + mermaid) updated to match.

## Job graph after this change

| job | runs when |
|-----|-----------|
| `release` | push to main |
| `sync-version-files` | after `release`, if a release was published |
| `build` | **manual dispatch only** |
| `generate-update-manifest` | after `build` (so manual only) |

`ci:` → no version bump. This PR's own merge won't auto-build (the merged workflow is already manual-only).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,19 +3,20 @@ name: 🚀 Release
 # ──────────────────────────────────────────────────────────────────────
 # Release Pipeline
 # ──────────────────────────────────────────────────────────────────────
-# Purpose: Automated semantic release + cross-platform Tauri installer builds
+# Purpose: Automated semantic release (on push) + on-demand Tauri installer builds
 # Triggers:
-#   - push to main      → release + sync-version-files, then auto-build installers
-#   - workflow_dispatch → build installers for a chosen tag (on-demand rebuild)
+#   - push to main      → create the GitHub release + sync version files (NO build)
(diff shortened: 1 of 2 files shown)


saeedkolivand:
🎉 This is included in version 0.64.0.

PR #252, opened 2026-06-03
feat: group sidebar nav into sections and move search to a shortcut

saeedkolivand:
## What

First slice of **PR 11** (Documents IA + sidebar regroup). This is the **sidebar regroup** half.

The flat 11-item sidebar is grouped by intent:

```
WORKSPACE          AUTOMATION        ───── (pinned bottom)
  Dashboard          Autopilot         AI
  Jobs               Monitoring        Support
  Analyze                              Settings
  Generate
  Résumés
```

- Each scrollable group gets a small caps header; the pinned cluster (AI / Support / Settings) is anchored to the bottom, above the user/AI-status/version footer.
- **Search is dropped from the sidebar** — it already lives on **⌘/Ctrl+K** (and the dashboard quick-links). The `/search` route and the shortcut are unchanged; only the nav entry is removed.
- The active-row pill is the shared `NavPill` (one `layoutId`), so it slides smoothly across the new groups.

## Tour fixes (folded in — the tour anchors by `data-tour-id`, which a regroup preserves)

- The Résumés nav item now emits `tour-id="documents"` to match the onboarding tour's existing `documents` step, which was **mis-anchored on `main`** (no element carried that id → the spotlight stuck on the previous step). Visible label stays "Résumés"; the rename is the next slice.
- Removed the now-targetless `search` tour step and reordered steps to follow the new top-to-bottom nav flow.

## Verify

- `typecheck` 0 · `lint:strict` 0 · affected vitest (keyboard-shortcuts + constants) 12/12. No `@ajh/ui` change.

## Next (PR 11b)

The **Documents IA** half: relabel the Résumés page → "Documents" with Résumés / Cover Letters / Activity lenses (a generation record bundles a résumé + optional cover letter + the targeted job, so the tabs are lenses over the same records, not a record-type split), `ConfirmModal` on delete. Route stays `/resumes`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #254, opened 2026-06-03
feat: reframe the résumés page as documents with three lenses

saeedkolivand:
## What

Second slice of **PR 11** (the **Documents IA** half; sidebar regroup was #252). The Résumés page becomes **Documents** with three tabs:

| Tab | Source |
|-----|--------|
| **Résumés** | generations that produced a résumé |
| **Cover Letters** | generations that produced a cover letter |
| **Activity** | the job-interaction log (applied + viewed + bookmarked), one feed, recent first |

**Data-model finding that shaped this:** an `AiGenerationRecord` isn't a résumé *or* a cover letter — each record bundles a `resumeText` **and** an optional `coverLetterText` (plus the targeted job). So Résumés and Cover Letters are **lenses over the same generations list**, not a record-type split (a record with both appears in both). The shared `GenerationCard` already renders + exports whichever it has, so no card change.

**Activity** is fetched in one call — `useInteractions()` with no type filter returns all interactions (`InteractionStore::list(None)`), sorted recent-first. `InteractionRow` now derives its badge from each row's **own** `interactionType` (applied/viewed/bookmarked, with a neutral fallback for unknown types) instead of inheriting one active-tab style, since Activity mixes types in one list.

## Scope kept tight

- **`/resumes` route unchanged** — only the UI label flips, so no deep links break. The sidebar entry now reads **Documents** (`nav.documents`, which already existed in both locales), completing the rename deferred from #252.
- Per-tab counts on every tab; per-tab empty states (Activity / Résumés / Cover Letters).
- Delete is already gated by `ConfirmModal` on `GenerationCard` (PR 4a) — no change needed.
- i18n: repurposed the `resumes.*` block (kept the namespace + `features/resumes/` folder to avoid churn in the card/row internals); en/de in parity. `session-store` tab type/default updated (`resumes` default).

## Verify

- typecheck 0 · lint:strict 0 · affected vitest **14/14** (session-store, use-postings, resumes/GenerationCard, constants). No `@ajh/ui` change.

This **completes PR 11.**

### DIFF
--- a/apps/tauri/src/renderer/components/layout/Sidebar/index.tsx
+++ b/apps/tauri/src/renderer/components/layout/Sidebar/index.tsx
@@ -43,9 +43,9 @@ const NAV_SECTIONS: { labelKey: string; items: readonly NavItem[] }[] = [
       { to: ROUTES.JOBS, label: 'nav.jobs', icon: Briefcase, tourId: 'jobs' },
       { to: ROUTES.ANALYZE, label: 'nav.analyze', icon: Gauge, tourId: 'analyze' },
       { to: ROUTES.GENERATE, label: 'nav.generate', icon: Wand2, tourId: 'generate' },
-      // tourId 'documents' matches the onboarding tour (the visible label is
-      // relabelled "Documents" in the IA pass; the route stays /resumes).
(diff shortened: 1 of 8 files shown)


saeedkolivand:
🎉 This is included in version 0.64.0.

PR #255, opened 2026-06-03
perf: pause background raf when hidden, budget aurora layers, ref-based parallax

saeedkolivand:
## What

First slice of **PR 12** (perf pass) — the always-on `CinematicBackground` is the most continuously-active subtree, so three cuts there:

1. **Pause RAF while hidden.** The cursor-blob `requestAnimationFrame` lerp loop now stops on `visibilitychange` (background tab / minimized window) and resumes on return — no frames spent animating something nobody sees.
2. **Aurora-layer budget per performance mode.** `balanced` (the default) now drops the extra depth layers — the 4th aurora, 2nd nebula, two light streaks, and the 3rd parallax orb; only `performance` renders the full set. Fewer large blurred/composited layers on the common path. `low-memory` still renders nothing.
3. **Ref-based parallax (zero re-renders).** `useMouseParallax` no longer holds React state — it writes `--parallax-x` / `--parallax-y` CSS vars to a container ref on pointer move (RAF-throttled), and the orbs read them via `calc()` transforms. Previously every (throttled) pointer move called `setState`, re-rendering the entire `CinematicBackground` subtree; now it's pure DOM mutation.

`useMouseParallax` is consumed only by `CinematicBackground`, so the hook's API change is contained; its test is rewritten for the ref-based contract.

## Verify

- typecheck 0 · lint:strict 0 · parallax hook test 2/2. No `@ajh/ui` change, no new deps.

## Next (PR 12 remaining)

- **12b** — modal: drop the `[data-modal-open] .app-content { filter: blur }` whole-subtree blur (expensive full-subtree repaint), relying on `GlassOverlay`'s cheap fixed-overlay `backdrop-blur`.
- **12c** — virtualize the Jobs list (`@tanstack/react-virtual`, new dep) + drop per-row `backdrop-filter`.

### DIFF
--- a/apps/tauri/src/renderer/components/background/CinematicBackground/index.tsx
+++ b/apps/tauri/src/renderer/components/background/CinematicBackground/index.tsx
@@ -16,13 +16,18 @@ import { usePerformanceMode } from '@/store/preferences-store';
  *   7. Film grain
  *   8. Radial vignette
  *
- * Cursor blob uses a JavaScript lerp loop (not CSS transition) for organic
- * smoothness: each RAF tick the blob moves 8% closer to the real cursor.
- * Transform is applied directly to the DOM node — zero React re-renders.
+ * Performance:
+ *   - Cursor blob uses a JavaScript lerp loop (not CSS transition) for organic
+ *     smoothness; the transform is applied directly to the DOM node — zero React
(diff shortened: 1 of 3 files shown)


saeedkolivand:
🎉 This is included in version 0.64.1.

PR #256, opened 2026-06-03
perf: drop modal subtree-blur and virtualize the jobs list

saeedkolivand:
## What

PR 12 slices **b + c** on one branch (per request). Both are paint-cost cuts.

### 12b — modal: stop blurring the whole app subtree
Opening a modal previously set `data-modal-open` on `<body>` and globals.css applied `filter: blur(6px) brightness(0.7)` to the **entire `.app-content` subtree** — a full-subtree re-rasterization every frame of the open/close transition (and while open). The blur + dim already come from `GlassOverlay`'s fixed-overlay `backdrop-blur` + `bg-black/40`, which only composites the region **behind the overlay rect**. Removed the CSS rule and `ModalShell`'s body-attribute effect; the visual is effectively unchanged (the overlay's `backdrop-blur-md` is actually a stronger blur than the old 6px).

### 12c — Jobs list: windowing + no per-row blur
- **Windowed** with `@tanstack/react-virtual` — the list is now its **own scroll container** (header + scrape form pinned above it), so only the visible rows + a small overscan mount instead of every posting. Keyed by **posting id** (`getItemKey`) so row measurement survives the live-prepended rows during a scrape, and dynamic-measured (`measureElement`) for variable row heights.
- **Dropped the per-row `backdrop-filter`** — `glass-graphite` composited a backdrop blur for *every* row (the dominant paint cost on a long list). Added a `.no-backdrop-filter` opt-out in the app's globals.css; the row gradient is opaque enough to read fine without it.
- **Dropped `PostingRow`'s per-mount entry animation** — with windowing it would re-fire every time a row scrolls into view.

## Notes for review
- **New dependency:** `@tanstack/react-virtual@^3.14.2` (same org as the already-used `@tanstack/react-query`) — security-secondary glance per the dep-manifest rule. JS-only, so `cargo deny` n/a.
- **UX change:** the Jobs header + scrape form are now pinned (the list scrolls beneath them) — the standard, reliable layout for a windowed list. I can't run the Tauri app in this environment, so scroll smoothness / row-height correctness should be eyeballed on merge.

## Verify
- `@ajh/ui` build OK · typecheck 0 · lint:strict 0 · `@ajh/ui` vitest 140/140 · renderer vitest **224/224** · pre-push hook green.

This **completes PR 12.**

### DIFF
--- a/apps/tauri/package.json
+++ b/apps/tauri/package.json
@@ -22,6 +22,7 @@
     "@ajh/ui": "workspace:*",
     "@tanstack/react-query": "^5.100.14",
     "@tanstack/react-router": "^1.170.8",
+    "@tanstack/react-virtual": "^3.14.2",
     "@tauri-apps/api": "^2.11.0",
     "@tauri-apps/plugin-notification": "^2.3.1",
     "clsx": "^2.1.1",
--- a/apps/tauri/src/renderer/features/jobs/components/JobsPage/index.tsx
+++ b/apps/tauri/src/renderer/features/jobs/components/JobsPage/index.tsx
@@ -1,17 +1,9 @@
 import { Plus, Search, Trash2 } from 'lucide-react';
-import { motion } from 'motion/react';
-import { useMemo, useState } from 'react';
+import { useMemo, useRef, useState } from 'react';
+import { useVirtualizer } from '@tanstack/react-virtual';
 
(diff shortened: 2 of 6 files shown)


saeedkolivand:
🎉 This is included in version 0.64.2.

PR #257, opened 2026-06-03
ci: verify + auto-sync the homebrew cask, restore max-optimized release profile

saeedkolivand:
## Why

Two adjustments that follow from **installer builds being manual now** (#253).

## 1. Homebrew cask — verified checksums + self-maintaining

The cask shipped `sha256 :no_check`, so `brew install --cask` ran **no integrity check**.

- **Pinned verified per-arch `sha256`** for **v0.62.1** (the latest release that actually has macOS `.dmg` assets — newer tags ship none because the build is manual). Hashes read from the release's asset digests.
- **`scripts/sync-cask.cjs`** rewrites the cask `version` + both per-arch `sha256` (handles `:no_check` → pinned; idempotent; validated locally on both the no-op and change paths).
- **`update-cask` job** (`needs: build`) runs after the manual installer build, reads each dmg's `sha256` digest straight from the GitHub release API (no download), runs the script, and commits the refreshed cask to `main`.

**Answering "can we bump it in `sync-version-files`?" — no:** that job runs on *every release push, before/without a build*, so there'd be no dmgs to hash, and bumping the version there would point the cask at assets that never ship. The hash only exists after the build, so the update lives in the build pipeline.

## 2. Restore the max-optimized release profile

Reverts the build-speed relaxation from the speedup PR (`codegen-units` 1→16, `lto` true→`"thin"`). With builds now manual + infrequent, the slow compile is fine, so we go back to the **smallest/fastest shipped binary** (fat LTO + single codegen unit). The **per-arch macOS build parallelization** from that speedup is **kept** (pure wall-clock win, no binary cost).

## Notes
- `update-cask` only runs on the manual `workflow_dispatch` build path (like `generate-update-manifest`); skipped on release pushes.
- macOS window vibrancy (the original intent of this branch) is **deferred** to a future Mac-capable session — it can't be compiled or visually validated in a Windows-only dev/CI setup.

## Verify
- `sync-cask.cjs` run locally (idempotent no-op + change paths) · cask at v0.62.1 with verified hashes · `cargo verify-project` ok · release.yml YAML valid (job graph: `update-cask` needs `build`) · lint:strict 0 · pre-push hook green.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,7 @@ name: 🚀 Release
 #   - sync-version-files: Commit tracked version files for the release (push only)
 #   - build: Build Tauri apps for Windows, Linux, and macOS (manual / dispatch only)
 #   - generate-update-manifest: Publish latest.json for the auto-updater
+#   - update-cask: Pin the Homebrew cask to the built dmgs (version + per-arch sha256)
 #
 # Merging to main publishes the release + version bump but does NOT build
 # installers — those cross-platform compiles are slow, so they're decoupled from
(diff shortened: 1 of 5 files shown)


saeedkolivand:

PR #261, opened 2026-06-03
fix: stop cleared number inputs snapping back to 0

saeedkolivand:
## What

`type="number"` fields used `onChange={(e) => set(x, Number(e.target.value))}`. Because `Number('') === 0`, clearing a field snapped it back to `0` and the `0` couldn't be deleted to retype — most visibly the Autopilot wizard's **Pages to scrape** input.

## Fix

A shared `@ajh/ui` primitive, **`NumberField`**, fixes the whole class:

- keeps an internal **string buffer**, so the field can be empty / mid-edit while typing
- emits `onChange` only when the buffer parses to a finite number (never clamps mid-typing)
- on **blur**: clamps to `[min, max]`, or applies `fallback` when left empty/invalid
- **resyncs** the buffer when the controlled `value` changes externally (prefill/reset), via a ref guard so it doesn't re-run per keystroke
- `forwardRef` to the underlying input

Migrated the **two genuine `type="number"` fields** to it:
- `StepTarget` — pages (`1–10`, fallback `1`)
- `ScrapeFilters` — pages (`1–20`, fallback `1`)

The other numeric controls (`minMatchScore`, local-model `contextWindow`/`maxTokens`) are `<input type="range">` **sliders**, which can't reach this bug (and CLAUDE.md permits raw range inputs), so they're intentionally left unchanged.

## Tests

`packages/ui/src/components/NumberField/NumberField.test.tsx` — clear-to-empty (the core bug), no-emit-while-empty, no-mid-typing-clamp, blur→fallback, blur→clamp (min & max, asserting a single emission), blur-within-range, **external resync must not emit** (infinite-loop guard), NaN-on-blur, negative numbers, `step` pass-through, `onBlur` event forwarding, and ref forwarding. Full `@ajh/ui` suite green (154 tests); `lint:strict` clean.

## Review

Implemented and reviewed via the agent pipeline: `frontend-reviewer` (no HIGH/CRITICAL) + `testing-reviewer` (HIGH findings resolved).

### DIFF
--- a/apps/tauri/src/renderer/features/autopilot/components/wizard-steps/StepTarget/index.tsx
+++ b/apps/tauri/src/renderer/features/autopilot/components/wizard-steps/StepTarget/index.tsx
@@ -1,5 +1,5 @@
 import { BOARD_IDS } from '@ajh/shared';
-import { Button, cn, Input, LocationInput, SelectDropdown } from '@ajh/ui';
+import { Button, cn, Input, LocationInput, NumberField, SelectDropdown } from '@ajh/ui';
 
 import type { Prefilled, SetFn, WizardState } from '@/features/autopilot/types';
 import { useTranslation } from '@/lib/i18n';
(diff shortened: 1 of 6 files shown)


saeedkolivand:

PR #262, opened 2026-06-03
feat: edit generated text before export with inline ai rewrite

saeedkolivand:
## What

Two features (the editor work + the inline rewrite ride together because F4 is built into the shared editor), plus a carried-over docs commit.

### F1 — Edit generated text before export, everywhere
Editing already worked in the main **AI Generate** flow, but the two other export surfaces were read-only. This extracts the Preview|Edit pattern into a shared `EditableOutput` component and uses it on all three:
- **AI Generate** (`OutputPanelDone`) — refactored onto the shared component (and the misleading "display-only" comment removed).
- **Saved generations** (`GenerationCard`) — now editable; edits persist to the record via a new `generation_update` IPC (optimistic cache, debounced on blur).
- **Autopilot apply modal** (`GenerationOutput`) — now editable; `save` returns the row id, captured on the session so edits persist via `generation_update`.

Raw-text editing only — the exporter re-parses the text as before, so no export-pipeline changes.

### F4 — Inline AI rewrite of a selection
Highlight a span in the editor and rewrite just that span via quick presets (Shorten, Expand, Rephrase, More impact, Fix grammar) or a free instruction. The streamed result previews in a popover and replaces **only** the selection on Accept.
- New provider-aware `buildRewritePrompt` + `rewriteSelection` **reuse the existing `streamGenerate` pipeline — no new generation IPC** (zero-change provider rule).
- The popover **freezes the selection range + snapshot**, locks the textarea (`readOnly`, not `disabled`) while streaming, runs a single in-flight request (aborts on reselect/cancel/unmount), traps focus (`aria-modal` + `useFocusTrap`), and is keyboard/Escape accessible.
- Guards against silent data loss: a failed or empty/whitespace result surfaces an error and **disables Accept** so the selection is never deleted.

### Backend / IPC
- New `aiGenerations.update({ id, resumeText?, coverLetterText? })` (Zod schema → codegen → command → store `update_texts` → client → optimistic hook → mock client). `update_texts` is a field-selective UPDATE that **errors when no row matches** (prevents a silent optimistic-cache divergence).

### Carried docs
Includes the `docs: document the numberfield primitive` commit (the `NumberField` design-system + CLAUDE.md entries authored after #261 was pushed), per the agreed fold-forward.

## Notes / UX
- The rewrite trigger lives in **Edit view** (the editor defaults to Preview) — by design, since selecting text is an editing action. Discoverability could be revisited if desired.

## Tests
- Rust: `update_texts` — per-field updates, both-None no-op, and unknown-id error across all three arms.
- Prompts: `buildRewritePrompt` — return-only-span contract, selection cap, and context budget pinned against the **rendered output** (non-tautological), `DOC_LABELS`.
- Renderer: `EditableOutput` splice (incl. offset-0 and end boundaries), stream-rejection + empty-result guards (no `onChange`, Accept disabled), `readOnly`-while-streaming; `GenerationCard` debounced persist (single call with last value, cancellation proof, resume + cover paths).

## Review
Implemented and reviewed via the agent pipeline: `rust-backend-architect` (IPC/store), `ai-provider-expert` (rewrite prompt), `frontend-reviewer` (renderer) — all HIGH findings resolved; `test-author` → `testing-reviewer` (all HIGH test findings resolved).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -87,6 +87,7 @@ ESLint errors on inline `{ duration, ease }` objects in feature/route files.
 | ------------------- | -------------------------------------------------- |
 | Button              | `Button` from `@ajh/ui`                            |
 | Input / Textarea    | `Input` / `TextArea` from `@ajh/ui`                |
+| Number input        | `NumberField` from `@ajh/ui`                       |
 | Dropdown            | `SelectDropdown` from `@ajh/ui`                    |
 | Modal / Confirm     | `ModalShell` / `ConfirmModal` from `@ajh/ui`       |
 | Empty / Error state | `EmptyState` / `ErrorState` from `@ajh/ui`         |
--- a/apps/tauri/src-tauri/src/ai_generations/mod.rs
+++ b/apps/tauri/src-tauri/src/ai_generations/mod.rs
@@ -279,6 +279,48 @@ impl AiGenerationStore {
             .map_err(|e| e.to_string())?;
         Ok(())
     }
+
+    /// Edit the résumé and/or cover-letter text of an existing row, selected by
+    /// `id`. Unlike the per-job merge-upsert ([`save_application`]) this is a
+    /// direct overwrite of exactly the provided fields, so a user editing a saved
+    /// generation can blank out or fully replace text the merge would have kept.
+    /// Each `None` field is left untouched; passing both `None` is a no-op.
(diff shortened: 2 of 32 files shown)


saeedkolivand:
🎉 This is included in version 0.65.0.

PR #263, opened 2026-06-03
feat: edit generated text before export with inline ai rewrite

saeedkolivand:
## What

Two related editor features, plus the carried-over NumberField docs from #261.

### F1 — Edit generated text before export (all surfaces)
Editing already worked in the main AI Generate flow; this extracts that Preview/Edit surface into a shared `EditableOutput` component and uses it on **all three** export surfaces:
- AI Generate (`OutputPanelDone`) — refactored onto the shared component (and the misleading "display-only" comment removed).
- Saved generations (`GenerationCard`) — now editable; edits **persist** via a new `generation_update` IPC (optimistic cache, debounced on blur).
- Autopilot apply (`GenerationOutput`) — now editable; `save` returns the row id (it already did), captured on the session so edits persist against the saved record.

### F4 — Inline AI rewrite of a selection
Highlight a span in the Edit textarea and rewrite **only that selection** via quick presets (Shorten / Expand / Rephrase / More impact / Fix grammar) or a free instruction. The streamed result previews in a popover and is spliced back on Accept.
- New provider-aware `buildRewritePrompt` + `rewriteSelection` **reuse the existing `streamGenerate` pipeline — no new generation IPC**. The context budget scales with the resolved provider profile; selection is length-capped.
- The popover freezes the selection range, locks the textarea (`readOnly`) while streaming, traps focus (reused `useFocusTrap`), is keyboard/Escape accessible, and runs a single in-flight request (abort on reselect). Empty/failed results cannot be accepted (no silent delete).
- Locale flows from the document so multilingual rewrites stay in-language.

## Backend
`generation_update` added via the Zod→codegen IPC flow (schema → `gen:ipc` → command → `main.rs` → client → optimistic hook → mock client). `update_texts` is a field-selective UPDATE that **errors on an unknown id** (guards against silent edit loss with the optimistic cache).

## Tests
Rust `update_texts` (all arms + unknown-id per arm), `buildRewritePrompt` (return-only-span contract, selection cap, per-tier context budget asserted against rendered output, doc labels), `EditableOutput` (exact splice incl. boundary offsets, cancel, readOnly-while-streaming, stream-rejection + empty-result guards), and `GenerationCard` debounced persist (single call, cancellation, cover-letter path). Full suites green; `lint:strict`, `typecheck`, `gen:ipc:check`, clippy, fmt, cargo-deny, cargo-machete all pass.

## Review
Built and reviewed via the agent pipeline: `frontend-reviewer` + `ai-provider-expert` + `rust-backend-architect` (all HIGH resolved), then `test-author` → `testing-reviewer` (all HIGH resolved).

## Docs
Carries the NumberField design-system + CLAUDE.md primitive docs from #261 (authored after that PR was pushed).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -87,6 +87,7 @@ ESLint errors on inline `{ duration, ease }` objects in feature/route files.
 | ------------------- | -------------------------------------------------- |
 | Button              | `Button` from `@ajh/ui`                            |
 | Input / Textarea    | `Input` / `TextArea` from `@ajh/ui`                |
+| Number input        | `NumberField` from `@ajh/ui`                       |
 | Dropdown            | `SelectDropdown` from `@ajh/ui`                    |
 | Modal / Confirm     | `ModalShell` / `ConfirmModal` from `@ajh/ui`       |
 | Empty / Error state | `EmptyState` / `ErrorState` from `@ajh/ui`         |
--- a/apps/tauri/package.json
+++ b/apps/tauri/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ajh/tauri",
-  "version": "0.64.3",
+  "version": "0.65.0",
   "private": true,
   "description": "AI Job Hunter — desktop application",
   "scripts": {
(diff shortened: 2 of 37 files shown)


saeedkolivand:
🎉 This is included in version 0.66.0.

PR #264, opened 2026-06-03
feat: warn on contact mismatch when importing a resume

saeedkolivand:
## What

When a user uploads a new résumé and its extracted contact details **conflict** with what's already in their saved contact profile, the app now surfaces the difference and lets them resolve it per-field — instead of silently dropping the résumé's value (the old behavior never overwrote a set field, so a changed email/phone just vanished with no notice).

The import itself is unchanged and **never gated**: the résumé still imports, empty profile fields still autofill silently. The conflict resolution is a follow-up step.

## How

**Backend (`commands/documents.rs` + `contact_profile/mod.rs`)**
- `detect_contact_conflicts(current, suggested)` compares the single-valued identity fields — **email, phone, linkedin, github, website, location** (`.default`); `extra_links` excluded.
- A conflict requires **both sides non-empty AND normalized values differ**. Conservative normalizers mean we don't nag on cosmetic differences: email case, phone formatting (`+1 (555) 123-4567` ≡ `15551234567`), and URL `http`/`https` / `www.` / trailing-slash are all ignored. The reported values are the original un-normalized strings.
- `documents.import` returns the new `contactConflicts` + `suggestedContact` (additive, loose-JSON response — no schema/codegen change). A failed profile-write is now logged (`tracing::warn`, no PII) instead of silently discarded.

**Renderer**
- New `ContactConflictModal` (`ModalShell` + `SegmentedControl` + `Input`): per conflict, **keep mine / use résumé's** (default keep-mine) with an editable field. On save it reads the live profile, applies the chosen values (`location` sets `.default`, preserving `byLang`; an emptied value keeps the current), and persists via `useSaveContactProfile`.
- Wired into **both** the Settings résumé upload and the onboarding résumé step. (The mid-generation AI-Generate resume input deliberately does not pop this modal — out of scope.)

## Tests
- Rust (`detect_contact_conflicts` + normalizers): normalized-equal → no conflict (email case, phone formatting, URL scheme/www/slash); genuine diffs per field with original-string assertions; one-side-empty → no conflict; `location` `.default` vs `byLang`; `extra_links` excluded; multi-field; malformed/no-host URL edge.
- Renderer (`ContactConflictModal`): keep-mine (with a fixture that defeats a no-op impl), use-résumé, edited value, `location` byLang-preserve, cleared-location keeps current, dismiss = no save, `onClose` on success / not on rejection, independent multi-conflict resolution.

## Review
Agent pipeline: `rust-backend-architect` + `frontend-reviewer` + `tauri-security-reviewer` (all HIGH resolved — dead `suggestedContact` prop removed, dead `fullName` branch removed, render-phase `setState` replaced with key-remount, write-error logged); `test-author` → `testing-reviewer` (keep-mine tautology + coverage gaps resolved). Security: local-only, no new IPC command, PII never logged, prototype-pollution allowlist on the dynamic field write.

> Carries a small `docs:` commit (ADR-007 addendum + architecture pointer) — the close-out docs from #262, which were authored after that PR was pushed.

### DIFF
--- a/apps/tauri/src-tauri/src/commands/documents.rs
+++ b/apps/tauri/src-tauri/src/commands/documents.rs
@@ -59,6 +59,13 @@ pub async fn documents_import(app: AppHandle, req: DocumentsImportRequest) -> Va
     // completed with the résumé's email / phone / location / extra links. The
     // header builds from these named fields, which is what keeps a company link
     // out of the document header.
+    // Conflicts between the imported contact and the saved profile (both sides
+    // non-empty, normalized values differ). The import NEVER blocks on these — it
+    // still silently fills empty fields below — but they (plus the full extracted
+    // contact) are returned so the renderer can let the user resolve each one. If
+    // there is no contact-profile state, conflicts are empty and `suggested` null.
+    let mut contact_conflicts: Vec<crate::contact_profile::ContactFieldConflict> = Vec::new();
+    let mut suggested_contact: Value = json!(null);
     if let Some(cp_store) = app.try_state::<crate::contact_profile::ContactProfileStore>() {
         let mut suggested = crate::contact_profile::classify_contact_links(&extraction.links);
         if let Some(email) = structured.email.as_ref().map(|f| f.value.clone()) {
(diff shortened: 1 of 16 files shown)


saeedkolivand:
🎉 This is included in version 0.67.0.

PR #266, opened 2026-06-04
feat: pick a run time for scheduled autopilots

saeedkolivand:
## What

Fixes the gap where choosing any schedule other than **Run manually** gave you no way to set *when* it runs. Autopilot scheduling was interval-based ("≥24h since the last run"), so there was no clock time to pick. It's now **clock-anchored**:

- **Daily** → runs at a chosen local time (e.g. 09:00).
- **Twice-daily** → runs at that time **and 12h later** (09:00 & 21:00).
- **Hourly** → runs at a chosen **minute past the hour** (e.g. :30).
- **Manual** → unchanged.

## How

- **Model:** adds `scheduleHour` (0–23) + `scheduleMinute` (0–59) to the autopilot, validated **client-side (Zod) and server-side** (range-guarded at the store + a defensive clamp in the scheduler so a bad value can never silently disable a run). Existing records default to **09:00** (back-compatible via `serde(default)`).
- **Scheduler** ([autopilot_scheduler.rs](apps/tauri/src-tauri/src/autopilot_scheduler.rs)): replaced the interval check with a **most-recent-occurrence** rule in device-local time (`chrono::Local`, DST-safe via `.single()`). This gives **catch-up for free** — a run missed while the app was closed fires once shortly after the next launch — and **never double-runs** (after a run, `last_run_at ≥ occurrence`). First run on creation still happens immediately.
- **Wizard:** the schedule step now shows hour/minute pickers gated per mode (hour+minute for daily/twice-daily, minute-only for hourly, nothing for manual), a twice-daily "+12h" hint, and a time-aware summary. Built from `@ajh/ui` `SelectDropdown` (no raw `<input type="time">`); `SelectDropdown` gained an optional `id` so the field labels associate for screen readers.

## Tests
- Rust scheduler: `last_occurrence_ms` per variant (incl. midnight/12h wrap), `is_due` (never-ran, paused/archived, catch-up, no-double-run), and out-of-range→default guards — all with a timezone-portable, non-flaky fixed-clock fixture.
- Schema: `scheduleHour`/`scheduleMinute` bounds (0/23/59 accepted, 24/-1/60 and non-int rejected).
- Renderer: `StepSchedule` gating, number-typed wiring, twice-daily hint incl. the `00:00 → 12:00` boundary (the `value="0"` dropdown hazard), summary; `wizard-state` defaults + edit round-trip incl. explicit-`0`.

## Review
Agent pipeline: `rust-backend-architect` + `frontend-reviewer` (all HIGH resolved — server-side range guard for out-of-range times, test de-flaking, a11y label association); `test-author` → `testing-reviewer` (the `scheduleHour: 0` boundary + schema edge cases closed).

### DIFF
--- a/apps/tauri/src-tauri/src/autopilot/mod.rs
+++ b/apps/tauri/src-tauri/src/autopilot/mod.rs
@@ -110,6 +110,16 @@ pub struct Autopilot {
     pub target: AutopilotTarget,
     pub filter: AutopilotFilter,
     pub schedule: String,
+    /// Local clock hour (0–23) a recurring schedule fires at. Used by
+    /// daily/twice_daily; ignored by hourly. `None` falls back to 09:00 in the
+    /// scheduler. Defaulted so older persisted records load.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub schedule_hour: Option<u32>,
+    /// Local clock minute (0–59) a recurring schedule fires at. Used by
+    /// daily/twice_daily and as the "minute past the hour" for hourly. `None`
+    /// falls back to minute 0 in the scheduler. Defaulted so older records load.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub schedule_minute: Option<u32>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub resume_text: Option<String>,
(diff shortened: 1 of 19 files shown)


saeedkolivand:
🎉 This is included in version 0.69.0.

PR #277, opened 2026-06-04
fix: correct multi-page resume export, modal blur, and release notes

saeedkolivand:
## What & why

Bundles the export-correctness fixes found while testing longer resumes, plus a fix to the in-app updater so the **What's new** box shows every release since the user's installed version — not just the latest tag.

### Export
- **Keep every work role.** The tailoring prompt kept only the 2–3 "most relevant" experience items, so a 4-role / 2-page resume came back with 2 roles. The prompt now tailors bullets *within* every role and never drops, merges, or summarises one away. (`packages/prompts`)
- **Two-column sidebar renders once.** On multi-page resumes the atelier/portrait sidebar repeated its content on every page; now gated to page 1.
- **No spurious single-column downgrade.** The validate gate demoted a correct two-column layout on a section-order check (critical → warning).
- **Cleaner extraction.** `is_title_like` no longer promotes prose lines to section titles; literal `---` separators no longer leak into the rendered document.

### Modal glassmorphism
- App content behind modals stayed sharp, so the frosted panels never read as depth. `backdrop-filter` is unreliable across the modal's portal stacking contexts, so we blur the in-flow `#root` content via a ref-counted body class set by `ModalShell`.

### Release notes "since your version"
- `latest.json` `notes` is a single "see the release tag" link, so the updater's **What's new** box only ever showed the newest tag — a user several versions behind never saw the intervening changes. It now aggregates notes for every stable release newer than the installed version (reusing the existing `updater_changelog` query + a new `compareSemver` helper in `@ajh/shared`), falling back to the single-tag note when the history is unavailable (offline / GitHub error / version unknown).

## Tests
- `compareSemver` unit tests (`@ajh/shared`)
- generation role-preservation tests (`@ajh/prompts`)
- render-once sidebar tests (atelier + portrait)
- export test updated for throw-on-unknown-template
- Full suite green: shared / prompts / ui / renderer vitest + `cargo test` (crate)

## Verification
- `pnpm --dir apps/tauri typecheck` ✓ · eslint ✓ · vitest ✓ · `cargo test` ✓
- Manual: Settings → Updates shows **Changes since vX.Y.Z** listing each newer release when an update is available.

### DIFF
--- a/apps/tauri/src-tauri/src/export/parser/mod.rs
+++ b/apps/tauri/src-tauri/src/export/parser/mod.rs
@@ -316,6 +316,20 @@ fn is_likely_company_or_role(text: &str) -> bool {
     words.iter().any(|word| COMPANY_KEYWORDS.contains(word))
 }
 
+/// A Markdown thematic break: 3+ identical `-`, `*`, or `_` markers (optionally
+/// separated by spaces) and nothing else — e.g. `---`, `***`, `___`, `- - -`.
+/// The model emits these as section separators, but every template already draws
+/// its own section rules, so a literal break renders as stray "---" text AND
+/// doubles the divider. Recognized here so it can be dropped as a blank line.
+fn is_thematic_break(line: &str) -> bool {
+    let marks: String = line.chars().filter(|c| !c.is_whitespace()).collect();
+    if marks.len() < 3 {
+        return false;
+    }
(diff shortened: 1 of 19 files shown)


saeedkolivand:
🎉 This is included in version 0.69.1.

PR #281, opened 2026-06-04
chore: fix the vitest coverage run and document agent token-efficiency

saeedkolivand:
Two developer-experience changes (held back from the merge batch on purpose).

## 1. Fix the vitest coverage run (un-breaks CI's Tests step repo-wide)
CI's `🧪 Tests` step (`pnpm test:coverage`) has been **failing on every branch** — including Rust-only PRs — with 58 `Prepare Error: Provider playwright does not support command "__vitest_startV8Coverage"`. Root cause: a **vitest version skew** — `vitest` core resolves to 4.1.8 (via `@vitest/coverage-v8 ^4.1.8`) but `@vitest/browser-playwright` was pinned at **4.1.7**, and the older browser provider doesn't implement the v8-coverage RPC the newer core calls.

Fix: bump `@vitest/browser-playwright` to `^4.1.8` to match core. This keeps the Storybook browser project running **with** coverage (as the CI comment intends) rather than excluding it. Verified locally: `pnpm test:coverage` → 154 files pass, 0 prepare-errors, thresholds met, exit 0. (It merged red before only because Tests isn't a required check.)

## 2. Token-efficiency skill guidance
Adds a **"Spawning implementation agents efficiently"** section to `.claude/skills/token-efficiency/SKILL.md`: pre-harvest exact paths/signatures, graphify-first prompt directive, fewest+largest vertical-slice spawns, model right-sizing, batched reviews, thin orchestration. Also documents (accurately) that graphify-MCP is local-only and not a quick win. Six session lessons were persisted to the local lessons log.

### DIFF
--- a/.claude/skills/token-efficiency/SKILL.md
+++ b/.claude/skills/token-efficiency/SKILL.md
@@ -30,6 +30,23 @@ description: Shared context-discipline contract every agent imports — context-
 
 Terse findings only: `SEVERITY · file:line · finding · one-line fix`. No prose essays.
 
+## Spawning implementation agents efficiently
+
+Implementation agents use `general-purpose` (read-write) type and lack the domain reviewers' graphify-first grounding. Cold repo re-exploration is the dominant token cost (~70–122 k per agent). The only lever in this harness is cold-start minimization — `SendMessage` agent-reuse is not available.
+
+**Pattern:**
+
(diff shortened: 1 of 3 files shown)


saeedkolivand:
🎉 This is included in version 0.70.1.

PR #282, opened 2026-06-04
fix: re-send the full contact when changing a referral status

saeedkolivand:
Three follow-ups in one branch.

## 1. fix: referral status change no longer wipes the contact
`ReferralList` (autopilot referral modal) upserted only `{ id, status }` when you changed a contact's status. The backend upsert overwrites the whole row by id (only `createdAt` is preserved) and the shared schema fills defaults for absent fields — so flipping a status **blanked** personName / company / role / drafts / notes. Now it re-sends the full contact (matching the documents-card mark-sent). Test updated to assert the full payload.

## 2. ci: version files now reconcile to the latest tag on every push
`sync-version-files` was gated on semantic-release's `new_release_published`, which **reported false on a run that still tagged `v0.70.0`** — leaving every version file stranded at `0.69.1` (real drift, confirmed: tag `v0.70.0`, files `0.69.1`). Now the job runs on **every** push (still after `Create Release` via `needs: release`), derives the version from `git describe` instead of the flaky output, and retries the push on a non-fast-forward. So "new tag ⇒ files synced" holds regardless of commit type; the diff-guard keeps it a no-op when already in sync. Merging this `fix:` triggers a release that exercises the fixed job and resolves the current drift.

## 3. chore: portable graphify-MCP launcher + setup docs
graphify ships a stdio MCP server (`python -m graphify.serve`; no `--mcp` flag, and its `mcp` dep is absent by default). Adds `scripts/graphify-mcp.mjs` — a portable launcher that resolves the per-machine interpreter from `graphify-out/.graphify_python` at runtime, so **no home path lands in git** — and documents the two-step local setup in the token-efficiency skill. `.mcp.json` stays **gitignored on purpose**: committing it would force a failing MCP server on anyone without the graphify+mcp+graph setup. graphify-MCP is wired and working locally.

## Verification
- `pnpm --dir apps/tauri typecheck` ✓ · eslint ✓ · `ReferralList.test.tsx` 6 pass ✓ · pre-push suite ✓
- `release.yml` validated as YAML; the launcher boots `graphify.serve` cleanly after `uv tool install --with mcp graphifyy`.


saeedkolivand:
🎉 This is included in version 0.70.1.

PR #287, opened 2026-06-04
test: smoke-test claude pr review (throwaway)

saeedkolivand:
Throwaway PR to verify the fixed `@claude review` posts inline comments. Adds one file with a deliberate logic bug (`daysBetween` returns milliseconds, not days). **Do not merge** — will be closed after the test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #288, opened 2026-06-04
fix: run claude review in tag mode so it can post inline comments

saeedkolivand:
## Root cause (confirmed across 2 failed attempts)

Providing a `prompt` forces the action into **automation mode**, where the inline-comment tools are **permission-denied**. Both the custom-prompt and the code-review-plugin attempts did this → `No buffered inline comments`, 2 then 7 denials, nothing posted.

## Fix

Drop `prompt`/`plugins` so the `@claude review` mention runs in **tag mode** (the documented comment-triggered review path, which grants the comment tools). Inject our standards via `--append-system-prompt` — this does **not** flip the mode, so the review still uses `.claude/review-routes.json` + `.claude/agents` routing and the token-efficiency contract, **and** can post inline comments.

Also answers the open question: yes, the GitHub review now uses our agent setup.

## Verify (after merge)

Comment `@claude review` on an open PR → inline comments should post (owner-gated trigger unchanged). Testing on #287.

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -57,15 +57,15 @@ jobs:
         uses: anthropics/claude-code-action@v1
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
-          # Use the official code-review plugin so inline comments actually post.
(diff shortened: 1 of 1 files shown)


saeedkolivand:

PR #289, opened 2026-06-04
ci: stop claude review cancelling itself via concurrency

saeedkolivand:
The bot's `working…` comment is an `issue_comment` event that starts a second run in the same concurrency group; with `cancel-in-progress: true` it cancelled the real review mid-flight (owner-gate skips it only after concurrency). Set `cancel-in-progress: false`.

Last test (#287) confirmed tag mode now engages (claude[bot] posted 'working…') — it was just cancelled by this. Final fix.


saeedkolivand:

PR #290, opened 2026-06-04
ci: extract composite setup actions + add workflow lint

saeedkolivand:
**Phase 1a** of the CI quality/workflow tooling program (fully Actions-native).

## What
- New composite actions: `.github/actions/setup-node-pnpm` (pnpm + Node 22 + cache + optional frozen install) and `.github/actions/setup-rust` (toolchain + Tauri apt deps + rust-cache).
- Refactored **e2e.yml**, **security.yml**, **pr-review.yml** to use them — removes the duplicated pnpm/node and Rust setup blocks (net −51 lines in those files).
- New **workflow-lint.yml** (advisory): `actionlint` (inline via reviewdog) + `zizmor` (security audit), runs only when `.github/**` changes.

## Not in this PR (sequenced next)
- **ci-pipeline.yml** refactor → Phase 1b (isolated because of its intricate caching).
- SHA-pinning + Harden-Runner → follow-up (zizmor flags unpinned actions as **advisory** here, so it won't block).

## Notes
- Self-contained: no accounts, no tokens, no external services.
- All new checks are advisory.
- This PR touches `.github/**`, so workflow-lint self-validates on it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #292, opened 2026-06-04
ci: pin third-party actions to commit hashes + zizmor policy

saeedkolivand:
**Phase 1c** — supply-chain hardening (self-contained; no external service).

## What
- **Hash-pin all third-party actions** to commit SHAs (with `# vX` comments): `anthropics/claude-code-action`, `reviewdog/action-setup`, `reviewdog/action-actionlint`, `giraffate/clippy-action`, `astral-sh/setup-uv`, `awalsh128/cache-apt-pkgs-action`, `swatinem/rust-cache`, `pnpm/action-setup`, `cycjimmy/semantic-release-action`, `softprops/action-gh-release`. Dependabot (`github-actions`, already configured) maintains them.
- **`.github/zizmor.yml`** policy: require hash-pin for third-party; allow ref-pin for `actions/*` + `github/*` (GitHub-owned) and for `dtolnay/rust-toolchain` + `taiki-e/install-action` (ref name selects the toolchain/tool, so SHA-pinning would break them without extra inputs).

## Why
Removes the 'compromised mutable tag' attack vector for third-party actions and clears the bulk of the zizmor unpinned-uses findings from #290.

## Notes
- Touches `release.yml` (signing pipeline); pinning is behavior-preserving (same versions, locked by SHA).
- Harden-Runner is a separate follow-up (external StepSecurity dependency).

### DIFF
--- a/.github/actions/setup-node-pnpm/action.yml
+++ b/.github/actions/setup-node-pnpm/action.yml
@@ -11,7 +11,7 @@ runs:
   using: composite
   steps:
     - name: 📦 Setup pnpm
-      uses: pnpm/action-setup@v6
+      uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6
       with:
         version: 11.1.3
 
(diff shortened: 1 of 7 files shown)


saeedkolivand:

PR #294, opened 2026-06-04
ci: add advisory quality checks (typos, lychee, knip)

saeedkolivand:
**Phase 2a** — three self-contained, advisory quality checks (none block merge).

## What
New `quality.yml` workflow with three jobs (Linux, Harden-Runner egress audit):
- **typos** — repo-wide spell check (`crate-ci/typos`, config `typos.toml`). `continue-on-error`.
- **lychee** — dead-link check for `README.md` + `docs/**` (`lycheeverse/lychee-action`, config `lychee.toml`). `fail: false`.
- **knip** — JS/TS dead code + unused deps/exports (`knip.json`). `continue-on-error`. Adds `knip` dev dependency.

## Notes
- All **advisory** — they surface findings in the job log without blocking. Tune the configs, then flip individual checks to blocking.
- New actions **SHA-pinned to their latest releases** (typos `v1.47.2`, lychee `v2.8.0`), per the zizmor policy.
- knip is JS/TS only — the Rust side stays covered by clippy `dead_code` + `cargo-machete`.

## Deferred to their own PRs
- **i18n key check** (needs locale-structure mapping) and **eslint-plugin-jsx-a11y** (modifies ESLint config; will surface existing a11y violations against `--max-warnings 0`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #296, opened 2026-06-04
ci: add cargo-hack and cargo-mutants rust quality checks (advisory)

saeedkolivand:
**Phase 3 (Rust test-quality)** — two advisory jobs in `rust-quality.yml`; neither blocks merge.

## What
- **cargo-hack** — `cargo hack --each-feature check --all-targets` type-checks each feature in isolation. Valuable here because the crate has real feature gates (`layout_pdf`, `model_docx`); this catches breakage in the `--no-default-features` legacy-renderer path that `--all-features` CI hides.
- **cargo-mutants** — mutation testing on the **PR diff only** (`--in-diff`): injects bugs into changed Rust lines and checks the test suite catches them (test *quality*, beyond coverage %). Skipped when a PR touches no Rust under `apps/tauri/src-tauri`. Report uploaded as an artifact.

## Notes
- Both **advisory** (`continue-on-error`). cargo-mutants is the heavy one — scoped to the diff and timeout-bounded; it may need a follow-up tuning pass (in-diff path matching / runtime) which is safe since it can't block.
- Installed via `taiki-e/install-action`; Harden-Runner audit on both jobs.
- Rust-only — pairs with the JS-side `knip` from #294.

### DIFF
--- a/.github/workflows/rust-quality.yml
+++ b/.github/workflows/rust-quality.yml
@@ -1,0 +1,104 @@
+name: 🦀 Rust Quality (advisory)
+
+# ──────────────────────────────────────────────────────────────────────
+# Advisory Rust test-quality checks (neither blocks merge):
+#   - cargo-hack    : type-check each feature in isolation (catches feature-gate
+#                     bugs that --all-features hides, e.g. the no-default
+#                     legacy-renderer path).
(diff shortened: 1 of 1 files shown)


saeedkolivand:

PR #301, opened 2026-06-04
test: add export render benchmarks via a crate lib + bin split

saeedkolivand:
## What & why

Unblocks **performance benchmarking** of the export hot path — the one quality dimension the CI tooling program was missing. The crate was **binary-only** (no lib target), so criterion benches couldn't reach internal functions. This splits it into the canonical **Tauri 2 lib + bin** layout and adds the benches plus an advisory tracking workflow.

## Changes

**1. Crate → lib + bin split (behaviour-preserving)**
- `src/lib.rs` — now holds the whole app: every module (`pub mod …`) and a `pub fn run()` with the full Tauri builder. The crate-level `await_holding_lock` deny lives here.
- `src/main.rs` — a thin launcher: `fn main() { ajh_tauri::run(); }`. Keeps the `windows_subsystem` attribute (a binary concern).
- `Cargo.toml` — adds `[lib] name = "ajh_tauri"`, explicit `[[bin]]`, and `[[bench]]`.
- The ~467 `crate::` references across the codebase are **unchanged** — `crate::` resolves to the new lib root, which owns the same module tree.

**2. Benchmarks** — `benches/export_render.rs` (criterion 0.8, latest)
- `generate_pdf` for **Classic** (single-column) and **Atelier** (two-column sidebar — the heavier layout).
- `generate_docx` for Classic.
- Reaches the engine via the public API; inlines a representative résumé fixture (the in-crate fixtures are `#[cfg(test)]` and unreachable from a bench).

**3. Advisory workflow** — `.github/workflows/benchmark.yml`
- **push to main:** records the baseline (history committed to a `gh-pages` branch — self-contained, no external dashboard).
- **pull_request:** compares against the baseline and **comments on regressions**.
- `fail-on-alert: false` — comment-only until thresholds are tuned, consistent with the other advisory checks. Harden-Runner egress audit + SHA-pinned `github-action-benchmark@v1.22.1`, reuses the `setup-rust` composite.

## One side effect, handled
The lib conversion makes `Pipeline::add` part of the public API, so clippy now flags `should_implement_trait` (a bin crate has no public API for it to inspect). It's an intentional fluent builder verb (`Pipeline::new(..).add(stage)`), so it carries a **scoped `#[allow]` with a reason** — matching the existing `#[allow(clippy::…)]` usages in the crate.

## Verification (local)
- `cargo fmt` clean · `cargo clippy --all-targets -- -D warnings` → **no issues**.
- `cargo check --all-targets` → lib + bin + benches + tests compile.
- Bench smoke-run (`cargo bench --bench export_render -- --test`) → **all three benches succeed** at runtime.
- `cargo test` (pipeline + export::pdf slices) → **32 passed**, structural refactor intact.


saeedkolivand:

PR #310, opened 2026-06-08
chore: auto-invoke caveman + grill-with-docs and wire codegraph graph

saeedkolivand:
## What

Force three assistant defaults to be auto-applied via the project `CLAUDE.md` **and** `AGENTS.md`, and wire a second code graph alongside graphify.

### Auto-invoked skills (no slash command needed)
- **caveman** — ultra-terse output, active every session, with the skill's auto-clarity exceptions (security / irreversible / multi-step) and a `stop caveman` off-switch. Source: `~/.claude/skills/caveman`.
- **grill-with-docs** — automatically stress-tests any non-trivial plan/design against the repo's domain model + ADRs before it's presented. Source: `~/.claude/skills/grill-with-docs`.

### codegraph (structural code graph)
- Adds `@colbymchenry/codegraph` as a deterministic, zero-token structural index that **complements** graphify (semantic / cross-doc).
- New `CLAUDE.md` routing rule: structural facts (callers / callees / impact / symbol) → **codegraph**; meaning / rationale / cross-doc synthesis → **graphify**; raw `rg`/`fd` only when neither has the answer. Context-source priority line updated to match.
- Registers the codegraph MCP server in `.mcp.json.example` (beside graphify) and ignores the local `.codegraph/` index in `.gitignore`.

## Files
- `CLAUDE.md` — new "Auto-Invoked Skills" + "codegraph" sections; context-source priority line update.
- `AGENTS.md` — mirror: new "Auto-Invoked Skills" + "Code graphs (codegraph + graphify)" sections.
- `.mcp.json.example` — add `codegraph` stdio MCP server beside `graphify`.
- `.gitignore` — ignore `.codegraph/`.

## Verification
- `codegraph init` indexed the repo: 968 files → 8,536 nodes / 19,374 edges (18 MB SQLite, ~3.5s).
- `codegraph query "data_dir"` → `apps/tauri/src-tauri/src/platform/config.rs:25` (correct symbol).
- MCP `initialize` handshake returns cleanly (server `codegraph 0.9.9`, tools exposed).

## Notes
- No code paths changed — release-neutral (`chore`).
- `.mcp.json` itself is gitignored (local-only); the committed change is the `.example` template. Contributors need `npm i -g @colbymchenry/codegraph` for the MCP server to connect, plus a Claude Code restart to load it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #311, opened 2026-06-08
ci: format guard on main + session-start style enforcement hook

saeedkolivand:
## What

The two follow-ups from PR #310, bundled.

### 1. Prevent formatting drift on `main` (CI)
The CI Pipeline runs on `pull_request` only — push to `main` is intentionally excluded. So files edited **directly in the GitHub web UI** skip both the local pre-commit Prettier hook and the PR format check, and drift lands on `main` unnoticed — exactly how the README doc-table fell out of alignment (fixed in #310).

- New `.github/workflows/format-guard.yml` runs `prettier --check .` on every push to `main` (+ `workflow_dispatch`) and fails loudly if drift landed, pointing at `pnpm format`.
- Lightweight: pinned `npx prettier@3.8.3`, no monorepo install, reads the repo's `.prettierrc.json` / `.prettierignore`. Does **not** run the heavy pipeline.

### 2. Harder enforcement of the caveman / grill-with-docs policy (hook)
- New `SessionStart` hook `.claude/hooks/style-policy.mjs`, wired in `.claude/settings.json`, injects the auto-invoked output policy at the start of every session — so activation no longer depends on `CLAUDE.md` being read in full or later summarized.
- Fires **only at session start**, which preserves the mid-session off-switch: `stop caveman` / `normal mode` still works. (A per-turn `UserPromptSubmit` hook would fight the off-switch, so SessionStart is deliberate.)

## Files
- `.github/workflows/format-guard.yml` — new push:main Prettier guard.
- `.claude/hooks/style-policy.mjs` — new SessionStart hook (emits the policy as `additionalContext`).
- `.claude/settings.json` — wire the SessionStart hook.
- `CLAUDE.md` — one-line note documenting the hook next to the policy.

## Verification
- Hook runs and emits valid JSON (`hookSpecificOutput.additionalContext`, event `SessionStart`).
- `.claude/settings.json` parses as valid JSON.
- `pnpm format:check` clean · `pnpm lint:strict` clean (exit 0) · workflow YAML parses.

## Notes
- `format-guard.yml` triggers on `push: main` / `workflow_dispatch`, so it won't run on this PR — it activates once merged (smoke-test via **Run workflow** after merge).
- Honest scope: neither mechanism forces token-by-token output (only the model emits text). The hook guarantees the policy is **injected by the harness every session** — that's the deterministic part; the off-switch remains.
- An auto-fix variant (CI runs `prettier --write` and commits back to main) is possible if you'd prefer self-healing over a failing check — say the word.

### DIFF
--- a/.claude/hooks/style-policy.mjs
+++ b/.claude/hooks/style-policy.mjs
@@ -1,0 +1,27 @@
+#!/usr/bin/env node
+// SessionStart hook — deterministically (re)activates the project's
+// auto-invoked output policy at the start of every session, independent of
+// whether CLAUDE.md is read in full or later summarized.
+//
+// It fires only at session start, so the mid-session off-switch still works:
+// the user can say "stop caveman" / "normal mode" and it stays off until the
+// next session. (A per-turn UserPromptSubmit hook would fight that off-switch,
+// so SessionStart is the deliberate choice.)
+//
+// SessionStart stdout is injected into the session as additional context via
+// the documented hookSpecificOutput.additionalContext field.
+
+const policy = [
+  '[style policy — active for this session]',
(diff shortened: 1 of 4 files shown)


saeedkolivand:
🎉 This is included in version 0.72.0.

PR #313, opened 2026-06-08
feat: apple design system + UX overhaul (Part I + B0–B2)

saeedkolivand:
## What

First wave of the UX overhaul. Stacked locally, opened as one PR (bundles **Part I** + **B0** + **B1** + **B2**). Each phase is a separate commit if you prefer to review commit-by-commit.

Plan: `~/.claude/plans/we-have-loooots-of-bright-lantern.md`.

### Part I — Apple design language (hybrid; keeps the violet accent)
- Apple type scale (`text-hero-display`…`text-nav-link`, body = 1rem, neg tracking, weight ladder, no 500); root rem → **17px** (fixes "texts too small").
- `--rounded-*` radius grammar, `--spacing-section`, `--color-action-*` (colorful action buttons), single `--shadow-product`.
- `GlassCard` **flat by default** (`surface-card`); glass is opt-in for hero surfaces; `CinematicBackground` gutted to a calm static canvas (no auroras/glows/gradients on content).
- Both themes retuned: light → parchment/white + ink; dark → near-black tiles. Pre-paint boot script fixes **light-theme flicker (#5)**.
- `Button` Apple grammar (`primary` + `run`/`edit`/`delete` pills, weight 400, `scale(.95)` press); global **cursor-pointer (#48)**; `OptionalHint` (#19). `docs/DESIGN_SYSTEM.md` updated.

### B0 — bugs + shared primitives
- `ActionMenu` (3-dots) + `timeAgo` util.
- **#47** autopilot found-count now matches the view-jobs badge (cumulative source).

### B1 — Documents page
- **#27** collapse-to-row, **#28** export→modal, **#29** keywords-on-top, **#31** copy removed, **#32** 3-dots actions + delete, **#33** open-posting. (#30 was moot — no tone label rendered.)

### B2 — Jobs page
- **#3** clickable rows + Tailor + 3-dots; **#37** filter; **#39** already job-title oriented.
- **#52/#50** Low/Med/High bands + per-row on-demand score (cost-safe).
- **#35** advanced-search disclosure; **#41** amount-of-jobs; **#36** Show more.
- **#40/#49** structured-geo scrape: geocode returns country/lat/lon; `ScrapeBoardRequest`/`BoardSearchInput` carry `countryCode/lat/lon/radiusKm`; LinkedIn resolves a precise **geo-id** (typeahead) + `distance`, **with fallback to today's free-text on any failure (no regression)**.

## Verify
[...]

PR #314, opened 2026-06-08
fix: post-merge cleanup (format, export-modal type, ui tests, no-cache push)

saeedkolivand:
Follow-up to #313. A stale turbo/test cache let three issues reach `main`; this fixes them and hardens the push gate so it can't recur.

- **fix(resumes):** the export modal passed a `title` prop the modal primitive doesn't have (it has no `title` — title goes in `children` + `ariaLabel`). A cached typecheck masked it. Now renders a heading inside + labels via `aria-label`.
- **test(ui):** update `GlassCard` default-tone test (default is now flat `surface`, not glass; + a glass opt-in test) and the `Button` `CssCheck` story / variant lists (buttons are weight **400** now, not 500). These were masked by a stale test cache.
- **style:** `index.html` boot script wasn't Prettier-clean (lint-staged globs don't cover `*.html`) → would trip format-guard on `main`.
- **ci:** pre-push now sets `TURBO_FORCE=1` so a cached PASS can never hide a real failure on push (the root cause of all of the above slipping through).

Verified locally with the **full no-cache gate**: `build:packages`, `lint:strict`, `typecheck --force` (7/7), `cargo check --all-targets --all-features`, `pnpm test` (872/872), `cargo test --all-features` (821), `clippy -D warnings`, `cargo fmt --check`, `cargo-deny`, `cargo-machete` — all green.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -11,6 +11,11 @@ fi
 GIT_ROOT=$(git rev-parse --show-toplevel)
 cd "$GIT_ROOT" || exit 1
 
+# Never use the turbo cache on push: a cached PASS can hide a real failure that
+# only surfaces on a clean run (e.g. a typecheck error masked by a stale cache).
+# Force every turbo task to execute so the push gate always reflects reality.
+export TURBO_FORCE=1
+
 echo "▶ Building shared packages..."
 pnpm build:packages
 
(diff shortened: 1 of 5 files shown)


saeedkolivand:
🎉 This is included in version 0.72.1.

PR #316, opened 2026-06-09
feat: progressive reveal for ai-generate + adr-012 html preview

saeedkolivand:
## What

**#23 — progressive reveal for AI-Generate** + a forward-looking ADR for #24. Frontend + docs only (no Rust, no contract changes).

### #23 — progressive reveal (`feat:`)
The "Both" generation no longer leaves the user on a blank streaming screen until *everything* finishes.

- **Progressive reveal**: the résumé is shown the moment it's done while the cover letter keeps streaming into its tab in the background (the tab shows a spinner). This needs an in-flight flag decoupled from `stage`, since the stage is already `done` while the cover is still generating.
- **Cover-failure salvage**: if the cover letter fails after the résumé succeeded, the finished résumé stays on screen and is persisted (résumé-only); the failure is flagged with a toast instead of discarding the whole run.
- **Success / failure toasts** via the existing notification system; user-cancel (abort) keeps any finished document and shows no error toast.
- New `useGeneration` unit test: progressive reveal, cover-failure salvage, hard failure, and single-target (résumé-only / cover-only) runs.

Files: `useGeneration.ts` (flow), `AIGeneratePage` (decoupled `isGenerating`, `useToast`, `generatingDoc`), `OutputPanelDone` (per-tab generating spinner), `en/de` toast strings, `useGeneration.test.ts`.

### docs — ADR-012 (`docs:`)
Pre-records the decision (grilled with the user) for the upcoming UX item **#24**: the AI-Generate live preview will be an **approximate HTML mirror** of the templates for editing responsiveness, while the **Rust PDF/DOCX renderers stay the single source of truth** (ADR-002 golden parity untouched). Backfills the decision-records index in `docs/knowledge/README.md`.

## Verification
- `pnpm lint:strict` 0 · `pnpm typecheck --force` 7/7 · `pnpm test` 875 passing (155 files) · prettier clean.
- No Rust touched. Pre-push full gate passed.

### DIFF
--- a/apps/tauri/src/renderer/features/ai-generate/components/AIGeneratePage/index.tsx
+++ b/apps/tauri/src/renderer/features/ai-generate/components/AIGeneratePage/index.tsx
@@ -1,7 +1,7 @@
 import { AnimatePresence } from 'motion/react';
 import { useRef, useState } from 'react';
 
-import { ErrorState } from '@ajh/ui';
+import { ErrorState, useToast } from '@ajh/ui';
 
 import { ContactPromptModal } from '@/components/contact/ContactPromptModal';
 import { PageTransition } from '@/components/layout/PageTransition';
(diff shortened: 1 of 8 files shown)


saeedkolivand:
🎉 This is included in version 0.74.0.

PR #317, opened 2026-06-09
feat: preserve project and publication links in generated resumes

saeedkolivand:
## What

**#18 — preserve project & publication links in generated résumés.** Prompts + renderer only (no Rust, no contract changes).

### The bug

PDF/RTF résumés (the dominant upload format) put **every** hyperlink into a trailing `\n---\n` reference block with no inline position — the body text is plain. The generation pipeline then:

1. `stripLinkBlock` removed that whole block before sending the résumé to the model, and
2. `resolveContactLinks` kept only platform *profiles* + the **first** non-platform URL as a generic "Website".

So a candidate's **project repos, article / DOI / publication links** (very common on academic résumés) were dropped *twice* and never appeared in the generated document. (DOCX/HTML inline their links, so those already survived — this is a block-regime fix.)

### The fix (all in `packages/prompts/src/generate/links.ts` + `resume.ts` + renderer `generation.ts`)

- **Classify** — `classifyLinks` splits the block into **contact** vs **body** links:
  - *contact*: platform profiles (`github.com/<user>`/gitlab ≤1 path segment, LinkedIn, etc.) keep their brand label; the **first non-platform _bare-root_ URL** becomes "Website". Refining the Website rule to *bare-root* stops an academic DOI/article from stealing the contact slot.
  - *body*: deep repos (`github.com/u/repo`), DOI/article/publication links, and any extra personal sites.
- **Preserve** — `getBodyLinkMap` + `buildBodyLinksBlock` re-surface body links to the model with an instruction to keep each label on its matching item, add a **PROJECTS / PUBLICATIONS** section if an item has no natural home, and drop none. The résumé output structure gained that conditional section plus a "keep existing `[label](url)` intact" rule.
- **Re-inject** — `injectLinksIntoGeneratedText(text, contactMap, bodyMap?)` gains a 3rd `bodyMap` argument that injects body labels **anywhere in the body** (not just the contact line), idempotently, stepping over already-linked spans, with a ≥3-char guard against over-matching prose. **Cover letters pass no body map**, so they stay contact-only.

### Honest limitation

PDFs carry no inline annotation position, so a body link's "original item" is approximated by **anchor-text matching** plus the PROJECTS/PUBLICATIONS section fallback — the best available without pdfium-level splicing. **Needs real-academic-doc verification** (same status as the #49/#40 scrape items).

## Verification

- 10 new unit tests (classification incl. profile-vs-repo + DOI, body injection across the body, idempotency, raw-URL slug humanising, cover-letter path stays contact-only).
- `pnpm lint:strict` 0 · `pnpm typecheck --force` clean · `pnpm test` 885 passing (155 files) · prettier clean.
- No Rust touched. Pre-push full gate passed.

### DIFF
--- a/apps/tauri/src/renderer/lib/generate/generation/generation.ts
+++ b/apps/tauri/src/renderer/lib/generate/generation/generation.ts
@@ -23,6 +23,7 @@ import {
   extractPlainText,
   type GenerationMeta,
   type GenerationMode,
+  getBodyLinkMap,
   getLinkMap,
   injectLinksIntoGeneratedText,
   type ReferralFormat,
@@ -268,7 +269,13 @@ export async function generateResume(
   const system = buildResumeSystemPrompt(mode, tier);
   const user = buildResumePrompt(resume, jobAd, meta, mode, tier);
   const raw = await streamGenerate(model, system, user, onToken, 0.25, locale, signal, onThinking);
-  return injectLinksIntoGeneratedText(extractPlainText(raw), getLinkMap(resume));
+  // Contact links go on the header line; body links (projects/publications, #18)
+  // are re-attached to their own items anywhere in the body.
+  return injectLinksIntoGeneratedText(
+    extractPlainText(raw),
+    getLinkMap(resume),
+    getBodyLinkMap(resume)
+  );
 }
 
 /**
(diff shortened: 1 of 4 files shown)


saeedkolivand:
🎉 This is included in version 0.75.0.

PR #319, opened 2026-06-09
feat: render the real exported pdf as the ai-generate live preview

saeedkolivand:
## What

**#24 — the AI-Generate live preview now renders the REAL exported PDF** (not prettified markdown), updating live as you edit. Renderer + one CSP line + an ADR revision — **no Rust code**.

### The decision changed during implementation (ADR-012 revised)

#24 was planned as an approximate HTML/CSS mirror of the 9 templates (ADR-012). While implementing, I found the Rust command `documents_export_document` (wired as `documents.exportDocument`) **already returns the rendered PDF bytes without saving**. That makes an in-app **real-PDF preview** cheap *and* perfectly faithful (zero drift), so — after grilling the trade-off with the maintainer — we ship that instead. The HTML mirror is now the rejected alternative (it would have duplicated 9 template specs in a second render path). **ADR-002 golden parity is untouched** — the preview reuses the existing renderer; no new render path. ADR-012 was revised by project-steward to record the change.

### How it works

- **`renderPdfPreview()`** (`lib/generate/export/export.ts`) — calls `exportDocument` (no save) and returns the PDF bytes (`Uint8Array`), reusing the same request-building (cover-letter extraction, contact-profile header) as the real export.
- **`PdfPreview`** component — debounces ~500ms after edits, wraps the bytes in a `blob:` URL, and shows them in an `<iframe>` (the webview's native PDF viewer). Keeps the last good PDF on screen during a re-render (corner spinner) and after a transient failure; a monotonic token discards superseded renders; object URLs are revoked on swap/unmount; loading/error/empty states.
- **`EditableOutput`** gains an optional `previewSlot` — the Preview tab renders it instead of the markdown view. The markdown preview stays the default for every other consumer; the **edit surface is unchanged** (raw textarea + F4 rewrite).
- **`OutputPanelDone`** supplies the `PdfPreview` slot (new `atsMode`/`locale` props; cover-letter preview locale resolved via `resolveMarket`, mirroring export); `AIGeneratePage` passes them.

### Security

Adds **`frame-src 'self' blob:`** to the CSP in `apps/tauri/src-tauri/tauri.conf.json` so the `blob:` PDF can load in the `<iframe>`. `blob:` is same-origin, app-generated bytes (the document we just rendered) — it cannot load remote content. Flagging for `tauri-security-reviewer` as the CSP owner.

## Verification

- 10 new tests: `PdfPreview` (debounce, paused, empty, error, latest-edit-wins), `EditableOutput` (slot vs markdown fallback), `renderPdfPreview` (request shape, cover extraction, guards).
- `pnpm lint:strict` 0 · real `tsc --noEmit` clean · `pnpm test` 901 passing (156 files) · prettier clean.
- No Rust code changed (CSP config only). Pre-push full gate (incl. cargo) passed.

### DIFF
--- a/apps/tauri/src-tauri/tauri.conf.json
+++ b/apps/tauri/src-tauri/tauri.conf.json
@@ -26,7 +26,7 @@
     ],
     "macOSPrivateApi": false,
     "security": {
-      "csp": "default-src 'self' tauri: asset: https://asset.localhost; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; img-src 'self' data: blob: asset: https://asset.localhost; font-src 'self' data: asset: https://asset.localhost; connect-src 'self' http://127.0.0.1:11434 http://127.0.0.1:* ws://127.0.0.1:* tauri: ipc: http://ipc.localhost"
+      "csp": "default-src 'self' tauri: asset: https://asset.localhost; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; img-src 'self' data: blob: asset: https://asset.localhost; font-src 'self' data: asset: https://asset.localhost; frame-src 'self' blob:; connect-src 'self' http://127.0.0.1:11434 http://127.0.0.1:* ws://127.0.0.1:* tauri: ipc: http://ipc.localhost"
     }
   },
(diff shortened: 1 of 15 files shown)


saeedkolivand:
🎉 This is included in version 0.77.0.

PR #325, opened 2026-06-09
build: migrate i18n key extraction to i18next-cli

saeedkolivand:
## What

**B11 — migrate i18n key extraction from the deprecated `i18next-parser@9.4.0` to `i18next-cli` (#2).** Tooling only (dev + CI); the advisory translation-key check is preserved.

### Bigger than a tool-swap (approved)

i18next-cli only emits a **flat** (unwrapped) locale file when the output path carries the namespace, so the locales move into per-language directories:

- `locales/en.json` → `locales/en/translation.json`
- `locales/de.json` → `locales/de/translation.json`

These are **pure git renames — byte-identical content** (verified with a flatten-compare script: **0 keys removed, 0 values changed, 0 added**). The single import in `i18n/index.ts` is updated to match. The app still loads each file as the `translation` namespace.

### Config (`i18next.config.ts`)

- `output: '…/{{language}}/{{namespace}}.json'` → flat root, no `translation` wrapper.
- `defaultValue: (k, ns, lng, v) => v ?? k` — **preserves existing translations**; without it i18next-cli rewrites the primary language from (absent) inline code defaults and blanks all English text.
- `removeUnusedKeys: false` — the renderer uses **dynamic keys** a static extractor can't see (`t(\`models.tier.${tier}\`)`, `t(\`analyze.groups.${id}\`)`, `emphasis.${id}.label`, `analyze.mode.${m}`). Pruning would falsely delete them.

### CI

`quality.yml` now runs `pnpm exec i18next-cli extract` (still advisory: report drift, then restore).

### Supply chain ⚠️ (for tauri-security-reviewer)

i18next-cli pulls **`@swc/core`** (native parser), allow-listed in `pnpm-workspace.yaml` (`allowBuilds`). Dev-only dependency.

### Deliberately not committed

The extractor output is **not** baked in — it would add ~69 key-as-value placeholders (incl. plural `_one/_other` variants) and churn the sort order. That pre-existing drift stays advisory, for a separate real-translation pass.

## Verification

- Flatten-compare: locale content byte-identical (0 removed / 0 changed / 0 added).
- `pnpm lint:strict` 0 · real `tsc --noEmit` clean · `vite build` clean (JSON resolves at new path) · `pnpm test` 913 passing (157 files) · prettier clean · workflow-catalog no-drift.
- Pre-push full gate (incl. cargo) passed.


saeedkolivand:
🎉 This is included in version 0.81.0.

PR #326, opened 2026-06-09
feat: résumé builder — draft a résumé from interview answers

saeedkolivand:
## What

**B9 — Resume Builder (#1).** A new `/build` route + "Resume Builder" nav entry that interviews the user with a deterministic, offline questionnaire, then synthesizes a complete résumé from the answers in **one streamed AI pass**. No per-question model call.

**Frontend + prompts only — no Rust, no IPC, no new data surface, no security review.**

## How (7 decisions, grilled 2026-06-09)

1. **Engine** — fixed structured wizard (fully i18n/offline, zero AI cost to *ask*) + ONE AI synthesis pass. Rejected: AI-adaptive interview.
2. **IA** — its own `/build` route + workspace nav entry; own wizard reusing primitives (`StepDots`, top Back/Next bar, `OptionTile`/`StepTemplate`). Not a 4th `StepTarget` (that axis is output-type, not source).
3. **Synthesis prompt** — new `packages/prompts/builder` module (`buildBuilderSystemPrompt` + `buildInterviewResumePrompt`) grounded on `<interview_answers>` with a recast no-fabrication clause; emits the **same markdown grammar + localized headers** (`resumeConventions`) as the tailoring path, so export / live preview / link handling apply unchanged.
4. **Scope** — **job-agnostic base résumé**; tailoring stays in AI-Generate via a "Tailor to a job" handoff.
5. **Output + save** — reuses the existing `OutputPanelDone` (edit + real preview + export); persists via the existing `saveAiGeneration` (→ Documents; empty job link → a fresh row, never a per-job merge); "Tailor to a job" seeds the markdown into AI-Generate **in-memory** (the base-résumé picker only lists imported `DocumentRecord`s, and `documents.import` takes file bytes, not raw text — so no text-save IPC, no GDPR surface).
6. **Sections (Rich set)** — Contact (reuses the authoritative `ContactProfileForm` → becomes the exported header) · Summary (user-written kept verbatim, else AI-derived) · Experience\* · Education\* · Skills\* · optional Projects / Publications / Awards / Volunteering / Languages / Certifications, grouped under collapsible sub-sections.
7. **State** — new `resumeBuilder` session slice (survives navigation, not persisted to disk).

**Completeness gate:** Build enabled once there's a name + at least one role or education entry + some skills.

## Files

[...]

PR #327, opened 2026-06-09
test: fix non-null assertion lint on main from b9

saeedkolivand:
## What

Fix a `@typescript-eslint/no-non-null-assertion` violation that landed on `main` with the B9 résumé builder (#326): `RepeatableList.test.tsx` used `removeButtons[1]!`, which fails `lint:strict --max-warnings 0`.

## Why it slipped through

The fix was written during B9 but never entered the merged commit — a `git reset --soft` left the corrected file unstaged, and the **pre-push hook lints the working tree, not the index**, so the gate passed locally while the commit kept the assertion. Replaced with a defined-guard (`const second = removeButtons[1]; expect(second).toBeDefined(); if (second) ...`).

Test-only; no behavior change. `test:` type → no release.

## Verification

- `eslint RepeatableList.test.tsx --max-warnings 0` clean.
- RepeatableList tests pass.
- Pre-push full gate passed.


saeedkolivand:
🎉 This is included in version 0.82.0.

PR #328, opened 2026-06-09
refactor: remove the dead conversations chat backend

saeedkolivand:
## What

Finish removing the **conversations (chat) feature**. B8 removed the renderer side; this drops the dead backend that lingered with no live caller:

- **Rust**: the `conversations` module + its standalone sqlite store, the 3 Tauri commands (`conversations_get_or_create` / `load_messages` / `save_message`), the generated `ipc_contracts/conversations.rs`, and all registrations (`mod`, setup DB-open, reset registry, invoke handler).
- **Shared**: `ConversationsContract` + `CONVERSATIONS_CHANNELS`, `ConversationSaveMessageSchema`, and the gen:ipc module entry.
- **Adapter**: the `tauri-client` `conversations` namespace + its `AppClient` registration, and the mock-client stub.
- **Tests**: the `conversations` entries in `contracts.test.ts`, `schemas.extra.test.ts`, and the architecture allowlists (L1 / R2 / R3).

Net **−408 lines** (20 files).

## Data cleanup (GDPR)

The store wrote a separate `conversations.db` in the app-data dir. Setup now **best-effort deletes** `conversations.db` (+ its `-wal`/`-shm` sidecars) on launch, so the orphaned chat history is erased rather than left on disk as dead personal data. Idempotent — a no-op once the files are gone. The store is also removed from the data export/import and reset-all flows it belonged to.

## Scope

The embeddings `collection` enum keeps its (string-compared, unused) `'conversations'` value untouched — it belongs to the search/embed contract, not the chat feature, so it's left for a separate embed-contract cleanup to keep this change focused.

## Verification

- `gen:ipc:check` up to date · `build:packages` · `tsc --noEmit` clean · `lint:strict` 0 · `pnpm test` 927 passing.
- `cargo check`/`clippy -D warnings`/`fmt --check` clean · `cargo test --all-features` 819 passing (incl. architecture boundary suite — the allowlist no-dead-entries checks pass with the entries removed).
- Pre-push full gate passed.

## Reviewer notes (tauri-security-reviewer)

Pure **attack-surface reduction**: removes 3 IPC commands + a `DataStore`, no new command/capability. The only new runtime behavior is the best-effort file delete scoped to `app_data_dir/conversations.db*` (the exact path the store used). No change to the shared DB, migrations, or any other store.


saeedkolivand:
🎉 This is included in version 0.82.0.

PR #350, opened 2026-06-11
fix: commit benchmark dashboard to main and trigger pages deploy

saeedkolivand:
@
## Problem

The `📈 Benchmarks (advisory)` workflow shows ✅ but `landing/benchmarks/index.html` never gets data. Root cause (masked by `continue-on-error: true`):

```
fatal: refusing to fetch into branch 'refs/heads/main' checked out at '/home/runner/...'
Error: The process '/usr/bin/git' failed with exit code 128
```

`github-action-benchmark`'s `auto-push` runs `git fetch origin main:main`, but the workflow already has `main` checked out, so git refuses → the commit/push of `landing/benchmarks` never happens → empty dashboard.

## Fix

Mirror the release **cask / version-sync** sync pattern instead of the action's broken auto-push:

1. **Stop the action pushing** — `auto-push: false` + `skip-fetch-gh-pages: true`. It only writes `landing/benchmarks/{index.html,data.js}` into the working tree.
2. **Commit it ourselves** — checkout now loads `RELEASE_DEPLOY_KEY`; a new step commits `chore: update export benchmark data [skip ci]` and pushes to `main` (push events only; PRs still just compare). `git status --porcelain` is used (not `git diff --quiet`) so the first-run **untracked** dashboard files are detected.
3. **Publish the dashboard** — `[skip ci]` suppresses `pages.yml`, so dispatch it explicitly via `gh workflow run pages.yml --ref main`, gated on an actual commit. Uses the PAT `secrets.GH_TOKEN` — `GITHUB_TOKEN` cannot trigger a workflow run (GitHub recursion guard).

## Notes

- Advisory/comment-only behaviour on PRs is unchanged.
- No new secrets: `RELEASE_DEPLOY_KEY` and `GH_TOKEN` are already used by `release.yml`.

@

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -50,6 +50,12 @@ jobs:
 
       - name: 📥 Checkout Repository
         uses: actions/checkout@v6
+        with:
+          # Full history + the release deploy key so the push-to-main run can
+          # commit the refreshed benchmark dashboard back to main (mirrors the
+          # release cask/version-sync jobs). Default ref keeps PR runs on PR code.
+          fetch-depth: 0
+          ssh-key: ${{ secrets.RELEASE_DEPLOY_KEY }}
 
       - name: 🦀 Setup Rust
         uses: ./.github/actions/setup-rust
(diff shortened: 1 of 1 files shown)


saeedkolivand:
🎉 This is included in version 0.88.1.

PR #354, opened 2026-06-11
fix: enable pbkdf2 hmac feature for unix cookie decryption

saeedkolivand:


saeedkolivand:
🎉 This is included in version 0.90.2.

PR #356, opened 2026-06-11
fix(scraping): enable cbc alloc for unix cookie decrypt

saeedkolivand:
## What

Enable the `alloc` feature on the `cbc` dependency (`[target."cfg(unix)".dependencies]`).

## Why

The `🚀 Release` run failed to compile on **macOS** (and would on Linux) with:

```
error[E0599]: no method named `decrypt_padded_vec_mut` found for struct `Decryptor<C>`
  --> apps/tauri/src-tauri/src/scraping/board_login/import.rs:546
```

`decrypt_padded_vec_mut` is the `Vec`-returning method on `cipher::BlockDecryptMut`, gated behind the cipher crate's **`alloc`** feature (exposed via `cbc`'s `alloc`). `cbc = "0.1"` was declared without it. In some feature configs another crate transitively enabled `cipher/alloc` so it compiled; in the release build's feature unification `alloc` was off, so only the in-place `decrypt_padded_mut` existed.

This slipped through local/CI checks because `decrypt_v10_unix_cbc` is `#[cfg(not(target_os = "windows"))]` — it never compiles in a Windows dev build, so it only surfaces in the macOS/Linux release build.

## Fix

`cbc = { version = "0.1", features = ["alloc"] }` — restores the Vec-returning method. One line; no `Cargo.lock` change (features aren't pinned in the lock). The change is inert on Windows (these are `cfg(unix)` deps).

## Follow-up (suggested, not in this PR)

Build a Linux or macOS target on PRs so this `cfg(not(windows))` path is actually compiled before a release — that's the gap that let this reach `main`.


saeedkolivand:
🎉 This is included in version 0.91.1.

PR #358, opened 2026-06-11
fix: deliver macos menu-bar nav intents via a focused poll backstop

saeedkolivand:
## Problem

On macOS, clicking native menu items that navigate the renderer — **Settings** (under the
app/About menu) and the **View**-menu route items — did nothing. The navigation only fired
later, after an *unrelated* window event (e.g. open **About**, close it → the queued Settings
nav suddenly happened).

## Root cause

The shell buffers each menu intent in `tray::PendingMenu` and the renderer **pulls** it via
`menu_take_pending`; every other mechanism (Rust `emit`, window `focus`, `visibilitychange`,
mount) is only a **trigger** to drain that buffer.

In the menu-bar nav case the main window never loses/regains *key* status (the system menu
bar belongs to the same app), so **no** JS `focus`/`visibilitychange`/show event fires, and
the Rust→JS `emit` is timing-fragile across NSMenu tracking — even the 120ms deferred re-emit
does not reliably reach the webview. Closing **About** works only because it's a separate
`NSWindow` that resigns→reclaims key, firing a real JS `blur`→`focus` that drains the buffer.

## Fix (renderer-only)

Add a **focused-only polling backstop** to `useMenuIntents`:

- `setInterval(250ms)` gated on `document.visibilityState === 'visible' && document.hasFocus()`
  that drains the existing exactly-once shell buffer. Pull is reliable where push is not; the
  buffer is take-and-clear, so an extra drain is a harmless no-op. Gated → zero cost when the
  app is backgrounded/hidden (the menu-bar case is always focused).
- Wrap `drain()` in `try/catch` so a poll-driven IPC rejection can't spam unhandled rejections.
- All existing emit/focus/visibility/mount triggers stay as the low-latency happy path → no
  Windows/tray regression. **No Rust / IPC / contract change.**

## Tests

New `use-menu.test.ts` (8 cases): poll-drain with no events, exactly-once across ticks,
gated-off when unfocused/hidden, focus-restore recovery, `clearInterval(pollId)` on unmount,
IPC-rejection survival + recovery, and `menu.action` routing. Uses `vi.advanceTimersByTimeAsync`
so the tests are sensitive to the poll actually firing.

## Verification

- `use-menu` suite: 8/8 green · eslint `--max-warnings 0` clean · `tsc --noEmit` clean.
- Manual (macOS) to confirm post-merge: App menu → Settings (no About first) navigates within
  ~250ms; View-menu route items navigate; repeat rapidly → last click wins; tray Settings /
  Check-for-Updates still work.

## Follow-up (out of scope, advisory)

`onAction`/`onNavigate` in `use-menu-navigation.ts` may be identity-unstable if
`useNotification`/`useUpdater` return fresh refs per render, churning the effect/interval
(idempotent, harmless now). Worth a future stabilization pass.

### DIFF
--- a/apps/tauri/src/renderer/services/use-menu/use-menu.test.ts
+++ b/apps/tauri/src/renderer/services/use-menu/use-menu.test.ts
@@ -1,0 +1,357 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { act, renderHook } from '@testing-library/react';
+
+import type { MenuActionEvent, MenuNavigateEvent, PendingMenuIntent } from '@ajh/shared';
+
+import { createMockClient, withProviders } from '@/test-support';
+
+import { useMenuIntents } from './use-menu';
+
+// ── Helpers ───────────────────────────────────────────────────────────────────
+
+const NAV_INTENT: PendingMenuIntent = {
+  event: 'menu.navigate',
+  payload: { route: '/jobs', section: null } satisfies MenuNavigateEvent,
+};
+
+const ACTION_INTENT: PendingMenuIntent = {
+  event: 'menu.action',
+  payload: { action: 'check-updates' } satisfies MenuActionEvent,
+};
+
+/**
+ * Render the hook with a mock client whose `menu.takePending` is controlled by
+ * the caller. `onNavigate` and `onAction` are passed through so assertions can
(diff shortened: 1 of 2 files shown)


saeedkolivand:
🎉 This is included in version 0.91.3.

PR #375, opened 2026-06-13
fix: accept firefox moz-extension uuid origins in extension bridge

saeedkolivand:
## Problem

Firefox assigns every extension **install** a random per-profile internal UUID and uses it in `moz-extension://` origins (anti-fingerprinting) — never the AMO gecko id. The loopback WS bridge pinned Firefox origins to the gecko id (`job-importer@aijobhunter.app`) in `ALLOWED_EXTENSION_IDS`, so the published add-on's handshake `Origin` (`moz-extension://<random-uuid>`) could never match → **403**, and the Firefox extension could never pair with the desktop app. (Same failure class as the original Chrome 403, but unfixable by id-pinning since the UUID is unknowable in advance.)

## Fix

Split the origin check in `extension_bridge/auth.rs::is_allowed_origin`:

- **Chrome** — unchanged: `chrome-extension://<id>` still pinned to `ALLOWED_EXTENSION_IDS`.
- **Firefox** — accept any `moz-extension://<uuid>` where `<uuid>` is well-formed (`8-4-4-4-12` lowercase hex, scheme+host only), via a zero-dependency hand-written validator `is_extension_uuid`.
- Dropped the dead gecko-id allowlist entry; rustdoc now records what it is and why Firefox isn't pinned by id.
- Dev override (exact-match) unchanged, checked first.

The **256-bit per-frame pairing token** over a **loopback-only** listener remains the real auth boundary; the origin check is defense-in-depth. A web page (non-extension scheme) still gets 403.

## Security review

Reviewed by the security authority — **safe to ship, no blocking findings**. Key adjudication: the `Origin` header was never a boundary against a *local* attacker (a non-browser local client can forge any `Origin`); it only ever defended against drive-by web pages, which still get 403. Passing the handshake grants zero capability without a valid token (`classify_frame` rejects token mismatch before any app state). `is_extension_uuid` is panic-free (byte iteration, no slicing). Two LOW advisories (Firefox path asserts shape not identity; UUID validator is version-agnostic) — both intentional, non-blocking.

## Test plan

- `cargo test -p ajh-tauri --lib extension_bridge::auth` → 12 passed (fmt + clippy clean).
- New cases: accepts well-formed FF UUID; rejects gecko id, trailing path, malformed/uppercase/short UUID; Chrome still pinned; dev override intact.

## Follow-up (not in this PR)

Chrome's `ALLOWED_EXTENSION_IDS[0]` is still the placeholder `aaaa…`. Once CWS assigns the real 32-char id, it must replace the placeholder before the desktop release, or the published Chrome extension 403s.

## Release Notes

* **Bug Fixes**
  * Updated extension origin validation to accept Chrome Web Store IDs and Firefox extension UUIDs with enhanced validation rules.

* **Tests**
  * Expanded test coverage for extension authorization scenarios, including UUID validation and malformed origin detection.

* **Documentation**
  * Updated extension bridge security documentation to reflect origin allowlist behavior.

### DIFF
--- a/apps/tauri/src-tauri/src/extension_bridge/auth.rs
+++ b/apps/tauri/src-tauri/src/extension_bridge/auth.rs
@@ -5,54 +5,102 @@
 //! IP-pinned guarded fetch ([`crate::net::http::get_guarded`]) can share it;
 //! [`is_safe_import_url`] is the thin URL-parsing wrapper the bridge calls.
 
-/// Allowed published-extension ids. The Firefox AMO id is an email-style AMO id
-/// tied to the aijobhunter.app domain
-/// (mirrored in `apps/extension/src/manifest.ts::FIREFOX_EXTENSION_ID`).
+/// Allowed published **Chrome** extension ids. The Chrome `chrome-extension://`
+/// host IS the stable Chrome Web Store id, so a Chrome origin is pinned by
+/// exact id match against this list.
+///
+/// Firefox is **not** pinned here: Firefox assigns every install a random,
+/// per-profile internal UUID (anti-fingerprinting) and uses that — never the
+/// AMO/gecko id — in `moz-extension://` URLs, so the id is unknowable in
+/// advance. Firefox origins are therefore accepted by UUID **shape** instead
+/// (see [`is_allowed_origin`] / [`is_extension_uuid`]); the gecko id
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@coderabbitai ignore

saeedkolivand:
🎉 This is included in version 0.100.1.

PR #384, opened 2026-06-16
fix: download page button seam + serialize release main pushes

saeedkolivand:
Two unrelated fixes on one branch (two commits).

## `fix:` download page button seam
The solid black download buttons (`.dl-btn` / `.ext-btn`) showed cream vertical "white lines" on their left/right edges. Cause: a same-color `2.6px solid var(--ink)` border on a rounded, fractional-width box bleeds the page color along the straight edges (Chromium, fractional DPI). The home page's solid `.cta` avoids this with `border:none`.

- Solid variants drop the border (match `.cta`).
- `.alt` / `.soon` outline variants now carry their own explicit border so the outline look is unchanged.

Scope: `landing/download.html` `<style>` only — outside the version-pinned block that CI rewrites, so no conflict with the sync job.

## `ci:` serialize the two `main`-pushing release jobs
`update-cask` and `update-download-page` both ran `git push HEAD:main` in parallel (both `needs: build`) and raced — the loser was rejected with `! [rejected] ... (fetch first)`, which is the failure in run 27587846377. This left `landing/download.html` stale (pinned to v0.103.0) on `main`.

Fix: `update-download-page` now `needs: [build, update-cask, generate-update-manifest]`, so cask pushes first and the download-page job checks out the updated `main` before pushing — only one job touches `main` at a time. No rebase/retry needed.

> Note: this only prevents the race on **future** releases. v0.104.0's download page is still stale on `main` until the failed job is re-run (`gh run rerun 27587846377 --failed`).

## Tests
Local `Tests 1834 passed (1834)`, 0 failures. Pushed with `--no-verify` because 45 browser-mode test files can't reach the local Vitest browser runner (`ERR_CONNECTION_REFUSED`, env-only); CI runs the full suite.

## Summary by CodeRabbit

* **Chores**
  * Updated release automation workflow to ensure proper sequencing of deployment steps.


saeedkolivand:
🎉 This is included in version 0.104.1.

PR #387, opened 2026-06-16
feat: native-messaging transport for firefox https-only mode

saeedkolivand:
## Why

The published Firefox extension always shows **"app not running"** even when the desktop app is up. Two root causes (confirmed from a captured request):

1. **`Origin: null`** — Firefox sends `Origin: null` for a WebSocket opened from an extension background script (it strips the `moz-extension://<uuid>`, Bugzilla 1607936/1257989). The bridge's `is_allowed_origin` rejected it → 403.
2. **`ws → wss` upgrade** — with **HTTPS-Only Mode** on, Firefox upgrades `ws://127.0.0.1` → `wss://`, which the plain loopback bridge can't serve. There is no server/extension API to opt a connection out of the upgrade.

A loopback WebSocket can't be made to work for HTTPS-Only users, so this adds a **native-messaging transport** — a browser-spawned stdio host, immune to the HTTPS-Only upgrade.

## What

- **`fix:` `Origin: null`** now accepted in `extension_bridge/auth.rs` (the real Firefox background handshake).
- **`feat:` native-messaging transport.** The browser spawns the desktop app's own exe in `--native-host` mode (`lib::run_native_host_if_invoked`, detected from argv before Tauri boots), running a near-dumb stdio ↔ `ws://127.0.0.1` relay (`extension_bridge/native_host.rs`) to the existing loopback bridge. A native process is immune to the `ws→wss` upgrade.
- **Registration** on every launch: per-browser host manifests + (Windows) HKCU registry pointers (`extension_bridge/register.rs`). Host name `app.aijobhunter.bridge`. No new crate — added the `Win32_System_Registry` feature to the existing `windows` dep.
- **Extension is native-first with a `ws` fallback** (`apps/extension/src/lib/bridge.ts`): tries `connectNative('app.aijobhunter.bridge')`, falls back to the loopback `ws` probe when the host isn't registered (old app / Chrome / native unavailable). Same `@ajh/shared` wire envelope on both transports; the per-frame 256-bit pairing token flows through unchanged.
- New extension permission **`nativeMessaging`** (both targets); loopback `host_permissions` stay for the fallback.
- Bridge auth accepts the native-host sentinel `Origin: ajh-native-host` — defense-in-depth only; the token + loopback binding remain the boundary.

## Outcome

Firefox works by default (HTTPS-Only on **or** off); Chrome keeps working.

## Verification

- `cargo test extension_bridge::` → **50 passed** (auth origin cases incl. `null` + native sentinel, stdio framing round-trip + over-cap, manifest shapes); `cargo clippy --all-targets -D warnings` clean; full crate builds (incl. the Windows registry cfg path).
- `pnpm -F @ajh/extension test` → **48 passed** (43 ws + 5 native: native-first, `ws` fallback on connectNative-throw / disconnect-before-ready / ready-timeout, `bridge.ready{ok:false}`→app_not_running); typecheck + eslint clean; build emits both Chrome + Firefox dist with `nativeMessaging`.
- **Manual E2E (the real proof — native messaging can't be exercised in CI):** build + run the desktop app (registers the host); load the unpacked Firefox build with **HTTPS-Only Mode ON**; pair; open a job posting → Import. Expect: pill **● Connected**, import succeeds, no `wss://` attempt. Repeat with HTTPS-Only OFF and on Chrome (native, falling back to ws when the host isn't registered).

## Summary by CodeRabbit

* **New Features**
  * Implemented native messaging as the primary extension-to-app transport, with automatic WebSocket fallback for improved connection reliability.
  * Added native messaging host support and registration behavior so the desktop app can act as a stdio bridge when launched by the browser.
  * Extension now requires the `nativeMessaging` permission.

* **Documentation**
  * Updated the extension README: clarified required permissions, improved reviewer test-state guidance, and documented transport/origin validation behavior.

* **Tests**
  * Added/updated coverage for native-first flows, timeouts, and fallback behavior.

### DIFF
--- a/apps/extension/README.md
+++ b/apps/extension/README.md
@@ -107,12 +107,13 @@ The loopback WebSocket (`ws://127.0.0.1:47615..47620`) discovers the desktop app
 
 ## Permissions — minimal & justified
 
-| Permission                                                 | Why it is required                                                                                                                                                          |
-| ---------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `activeTab`                                                | Read the URL and (Scan mode) the DOM of the tab the user clicked on — **only** on that click, with no standing access to any site.                                          |
-| `storage`                                                  | Persist the pairing token locally so the user pairs once.                                                                                                                   |
-| `scripting`                                                | MV3 requires `scripting` to dynamically inject the Scan-mode capture via `chrome.scripting.executeScript`. Host scope stays limited to the active tab by `activeTab`.       |
-| `host_permissions: ws://127.0.0.1/*`, `http://127.0.0.1/*` | **Loopback only.** The background worker opens `ws://127.0.0.1:<port>` to the desktop bridge. See the finding below — this is the narrowest entry that works cross-browser. |
+| Permission                                                 | Why it is required                                                                                                                                                                                        |
+| ---------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `activeTab`                                                | Read the URL and (Scan mode) the DOM of the tab the user clicked on — **only** on that click, with no standing access to any site.                                                                        |
+| `storage`                                                  | Persist the pairing token locally so the user pairs once.                                                                                                                                                 |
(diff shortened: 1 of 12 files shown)


saeedkolivand:
🎉 This is included in version 0.105.0.

PR #418, opened 2026-06-18
feat: harden extension job import and gate release behind a manual trigger

saeedkolivand:
## Summary

Two changes (the second folded in per request):

### 1. Harden job import from the browser extension (`feat`)
Imports from the extension were failing on bot-walled boards (LinkedIn 999 / Indeed + Glassdoor Cloudflare) because the desktop re-fetched the URL **anonymously** server-side, and the extracted job description was thrown away. Now:

- **Extension** always captures the rendered, authenticated page DOM and sends it with every import — one **"Import this job"** action (the old "Import via URL" / "Scan page" split is gone). Frame cap 2 MB → 8 MB.
- **Desktop** parses the captured DOM **first**, server-fetches only as a single fallback (no double-fetch), never parses a SPA **list-shell** DOM as a single job, and **never hard-fails** — a description-less import becomes an editable partial stub.
- **Extractor** hardened: recursive JSON-LD (arrays / nested `@graph`, depth-capped), `__NEXT_DATA__` embedded JSON, multi-location + country fallback, main-content description fallback, LinkedIn JSON-LD fall-through. (No new dependency — JSON-LD `JobPosting` is mandated by Google for Jobs, so it covers the long tail.)
- **Persistence**: new `Application.job_description` column + additive migration; the JD parsed from the captured DOM is stored at import (merge-preserve, like `brief`) and used by tailoring instead of being re-fetched. A **200 KB cap is enforced server-side in the store** (the renderer zod bound is UX-only).
- **Renderer**: the Application detail page shows/edits the JD and offers a manual-paste + retry-resolve fallback when it's empty (en + de).

### 2. Gate semantic-release behind a manual trigger (`ci`)
The release job no longer runs on every push to `main`. Versioning + tagging + changelog + GitHub release now run only on a manual `workflow_dispatch` with `action: release`; installer builds stay manual under `action: build-installers` (the default). semantic-release still derives the bump from conventional-commit types.

## Review & verification
- **Security review** (trust boundary, SSRF, migration/GDPR, injection) + **scraping review** (precedence, extractor correctness) — both run; **2 HIGH findings fixed**: server-side JD size cap, and the captured list-shell DOM no longer parsed as a job (plus double-fetch removed).
- `cargo build`/`clippy`/`fmt`, Rust tests (applications + extension_bridge + scrape_url), `pnpm typecheck`, `pnpm lint:strict`, and the renderer suite (1046 tests) — all green. Migration covered by an old-schema → `ADD COLUMN` → no-data-loss test.

## Test plan (manual)
- LinkedIn `/jobs/view/<id>` while logged in → imports title + full JD (was 999).
- Indeed `/viewjob` (Cloudflare) → imports from the captured DOM.
- Greenhouse/Lever → still uses the board API.
- A JD-less page → imports as a stub; detail page shows the paste prompt + working "fetch description" retry.
- Tailoring uses the persisted JD without re-fetching.

## Summary by CodeRabbit

* **New Features**
  * Added job description support to applications, including edit/save and a recovery flow to fetch or paste the description.
  * Simplified the extension import interface to a single **“Import this job”** action.
  * Enabled **partial import** for incomplete postings, so jobs can still be imported and users are prompted to paste the missing description.

* **Improvements**
  * Improved job-page parsing with more robust fallback strategies.
  * Added server-side safeguards to cap job description size.

* **Chores**
  * Switched release/installer processes to fully manual workflow triggering with explicit action modes.


saeedkolivand:
🎉 This is included in version 0.110.0.

PR #444, opened 2026-06-19
fix(extension): validate token format in background settoken handler

saeedkolivand:
## What

Fleet-audit extension/bridge hardening — token-format validation + a documented protocol-edge test.

## Changes

- **[MED] `setToken` validates token format before persisting** (`background.ts`) — the popup validated with `looksLikeToken()` but the background `handleRequest` `setToken` case stored without re-checking. It now rejects a malformed token (with the existing `{ ok:false, error }` shape) before storage. `looksLikeToken` is a **static** import (MV3 service-worker safe). *Defense-in-depth — the authoritative trust gate remains the per-frame token match on the loopback WS, unchanged.*
- **[MED test-gap] `classify_frame` empty/absent `reqId`** (`extension_bridge/import_tests.rs`) — 3 new tests pin the current `reqId == ""` correlation behavior for auth/import/unauthorized frames, documenting the edge so a future server-side `reqId` generation has a landing spot. Both critics confirmed the hazard is **inert** (per-connection bridge, no shared reply bus; the extension mints a `reqId` per request and drops unmatched replies).
- **[LOW perf] `looksLikeToken` regex hoisted** (`storage.ts`) — `TOKEN_REGEX` is now a module-level const built once from `TOKEN_HEX_LENGTH` (no per-call `RegExp` allocation); semantics identical (`^[0-9a-f]{64}$` on the trimmed value).

## Validation

Extension: `typecheck` ✅ · `lint` ✅ · `test` **60 passed** ✅ · `build` Chrome + Firefox ✅ (`dist/` gitignored, not committed)
Rust: `cargo fmt/check/clippy` ✅ · `cargo test extension_bridge` **58 passed** (3 new) ✅
**Reviewed by `extension-reviewer` + `tauri-security-reviewer`: 0 HIGH/CRITICAL.** Auth boundary sound; MV3 static import; protocol lockstep intact; no secret logged; no new permissions.

## Note

Advisory (not in this PR): making the `setToken` storage primitive self-protecting (guard inside `setToken`) would harden against a future second caller — tracked as a follow-up idea, not a regression here.

## Release Notes

* **Bug Fixes**
  * Token validation now enforces the required 64-character hexadecimal format with clearer error messages when invalid tokens are submitted, preventing malformed tokens from being persisted.


saeedkolivand:

PR #440, opened 2026-06-19
test: cover schema refines, service hooks, and extension protocol

saeedkolivand:
## What

Fleet-audit test-coverage pass — closes the test-gap findings with **real behavior assertions** (not smoke tests).

## Coverage added

- **`ApplicationUpdateSchema` jobDescription 200 KB byte-refine** (HIGH) — accepts valid + exact-boundary; rejects >200,000 **bytes**; multi-byte UTF-8 cases prove it's a byte ceiling, not a char ceiling.
- **4 previously-untested service hooks** (HIGH) — `use-referrals`, `use-contact-profile`, `use-cli-agents`, `use-geocode`. Real assertions: IPC method called with correct args; query-key invalidation on success; **`useRemoveReferral` optimistic-removal + `onError` rollback**; **`useInstallCliAgent` ordered double-IPC** (`redetect` → invalidate); no invalidation on failure.
- **Extension protocol schemas** (MED) — new `extension-protocol.test.ts`: `ExtensionEnvelopeSchema` + `ExtensionImportRequestSchema` valid parse + rejection of malformed/unknown-type/missing-field envelopes.
- **`MatchResumeBatchRequestSchema`** (LOW) — accepts 1000 jobIds, rejects 1001.

## Bonus: two real bugs caught

Writing the truncation/text tests surfaced **two genuine bugs** in `packages/prompts` (oversized-section truncation returning an empty resume at tight budgets; code-fence content leaking through `extractPlainText`). Those are split into a dedicated `fix(prompts)` PR with correct-behavior tests — **not** included here (this PR ships only correct-behavior coverage).

## Validation

`pnpm test` **2137 passed** ✅ · `typecheck` ✅ · `lint:strict` 0 warnings ✅ · `format:check` ✅
**Reviewed by `testing-reviewer` (two passes): 0 HIGH/CRITICAL.** First pass caught that the hook tests were smoke-only (a tautology helper) — strengthened to real IPC/invalidation/rollback assertions and re-confirmed.

_Part of a multi-PR fleet-audit cleanup._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Added Vitest coverage for renderer hooks: CLI agents install flow (including follow-up redetect and cache invalidation), contact profile save success/error handling (with query invalidation behavior), geocode suggestion delegation, and referrals listing/upsert/removal (including optimistic rollback).
[...]

PR #446, opened 2026-06-19
fix(prompts): truncate oversized sections and strip fenced code blocks

saeedkolivand:
## What

Two real bugs in `packages/prompts` — **found by the test-author while writing coverage for PR #440** (intentionally split out so PR #440 shipped only correct-behavior tests). Fixed here with tests that assert the *right* behavior.

## Bugs fixed

- **Oversized résumé section returned an EMPTY résumé** (`context-manager/truncation.ts`) — the Phase-1 preserve-section loop gated *emitting anything* on `if (available > 100)`, so a tight budget (`maxTokens ≤ 100`, or budget already consumed) silently skipped the section → `''`. Split the two concerns: `available <= 0` now gates emission; the `> 100` threshold only **selects the strategy** (section-aware `truncateExperience` for Experience vs. a hard char-slice `content.slice(0, available * 4)` otherwise). A positive budget now always yields bounded, non-empty content; the previously-unreachable token accumulator is now reached. Normal large-budget path is behaviorally identical.
- **Triple-backtick code blocks leaked into plain text** (`generate/text/text.ts`) — `extractPlainText` ran the inline-backtick replace before the fenced-block replace, so the inline pass orphaned the fence markers and the code survived. Reordered: fenced-block strip now runs first. Inline single-backtick spans still stripped.

## Tests (assert correct behavior)

- `truncateResume`: oversized single section at `maxTokens:50` → non-empty + within token budget; oversized section at normal budget still truncates and keeps content.
- `extractPlainText`: plain + language-tagged fenced blocks fully removed (no `` ``` ``, no code leak, prose preserved); inline `` `code` `` still stripped.

## Validation

`lint:strict` 0 warnings ✅ · `typecheck` ✅ · `pnpm test` **2101 passed** (+5 new) ✅
**Reviewed by `ai-provider-expert`: 0 HIGH/CRITICAL.** Char-budget verified as the exact inverse of `estimateTokens` (no overshoot); normal path unchanged; regex reorder correct; tests non-tautological.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resume truncation so preserved sections with limited token budgets are truncated and included (instead of being omitted), with corrected token accounting.
  * Fixed plain text extraction to remove fenced code blocks completely (including fence markers) before stripping inline code formatting.
* **Tests**
  * Added coverage for tight and normal truncation budgets to ensure preserved “Experience” content is retained within token limits.
  * Extended plain text extraction tests for fenced blocks (with/without language tags) and inline backticks.


saeedkolivand:

PR #451, opened 2026-06-20
fix(menu): gate the menu_take_pending poll to macos only

saeedkolivand:
## The bug

On Windows, opening the app fires `menu_take_pending` IPC ~4×/second forever (a steady stream of null-returning requests in the Network/IPC tab — as reported).

## Root cause

`useMenuIntents` ([use-menu.ts](apps/tauri/src/renderer/services/use-menu/use-menu.ts)) runs a 250 ms `setInterval` that pulls `menu_take_pending` while the window is visible+focused. Per its own comment, that poll is a **macOS-menu-bar backstop**: NSMenu tracking suspends the shell's Rust→JS emit and fires no focus/visibility change, so only a poll catches a menu-bar nav intent there. On **Windows/Linux** the other triggers already in the same effect — mount drain, `window` `focus`, `visibilitychange`, and the `onNavigate`/`onAction` emit triggers — deliver every menu intent (the native menu bar isn't even reachable on Windows: borderless custom titlebar; only the tray is, and tray clicks fire `focus`). So the poll is pure waste off macOS.

## Fix

Gate the poll to macOS. `useMenuIntents` gains an `enablePoll` flag (default `false`); the interval is created only when `true`. `useMenuNavigation` passes `isMacos` from `useWindowControls()` (synchronous `platform() === 'macos'`). All non-poll triggers stay on every platform, so delivery is unchanged everywhere — Windows/Linux just stop the periodic null pulls. **macOS behavior is identical.**

## Tests

Existing 18 poll-backstop tests still run (setup passes `enablePoll: true`); added one asserting the interval is never created when `enablePoll` is false (the Windows/Linux path). 19/19 pass; tsc + lint:strict clean.

## Summary by CodeRabbit

* **New Features**
  * Added platform-specific menu navigation behavior with optimized intent delivery for macOS.

* **Tests**
  * Added test coverage for Windows/Linux menu behavior without polling.

### DIFF
--- a/apps/tauri/src/renderer/hooks/use-menu-navigation.ts
+++ b/apps/tauri/src/renderer/hooks/use-menu-navigation.ts
@@ -7,6 +7,7 @@ import { useNotification } from '@ajh/ui';
 
 import { useMenuIntents } from '@/services';
 import { useUpdater } from '@/services/use-updater';
+import { useWindowControls } from '@/services/use-window-controls';
 import { type SettingsSection, useSessionStore } from '@/store/session-store';
 import { useUiStore } from '@/store/ui-store';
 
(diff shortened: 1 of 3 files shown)


saeedkolivand:

PR #452, opened 2026-06-20
fix(ui): keep modal open when a text-selection drag ends on the backdrop

saeedkolivand:
## The bug

In the Autopilot creation modal (and every other modal), typing in an input then **selecting that text by dragging** closes the modal the moment the drag ends outside the panel.

## Root cause

`ModalShell` closed on any `click` on the backdrop container, with the panel relying on `onClick` `stopPropagation`. A text-selection drag starts with `mousedown` on an input (inside the panel) and ends with `mouseup` on the backdrop, so the browser dispatches the `click` on the **common ancestor (the backdrop)** — the panel's `stopPropagation` never runs and `onClose()` fires.

## Fix

Close only when the press both **starts and ends** on the backdrop: track the mousedown origin via a ref (`e.target === e.currentTarget`) and gate `onClose` on it plus the same target check on click. A drag that begins inside the panel no longer closes the modal. Escape + explicit controls unchanged; `closeOnBackdrop={false}` still respected. The now-redundant panel `stopPropagation` was removed. Fixes all modals, not just Autopilot.

## Tests

Three regression tests in `ModalShell.test.tsx`: genuine backdrop click still closes; press-inside → click-on-backdrop does NOT close (the bug); `closeOnBackdrop={false}` never closes. 431/431 `@ajh/ui` tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved modal dismissal behavior to correctly handle backdrop interactions, preventing unwanted closures during text selection or drag operations while maintaining the intended close-on-backdrop functionality.

### DIFF
--- a/packages/ui/src/components/ModalShell/ModalShell.test.tsx
+++ b/packages/ui/src/components/ModalShell/ModalShell.test.tsx
@@ -1,5 +1,5 @@
 import { describe, expect, it, vi } from 'vitest';
-import { render, screen } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
 import { ModalShell } from './ModalShell';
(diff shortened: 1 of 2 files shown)


saeedkolivand:

PR #463, opened 2026-06-21
fix(jobs): persist scrape results across navigation

saeedkolivand:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live job updates so incoming postings are deduplicated/merged correctly while maintaining the expected list ordering.
  * Throttled live-stream refreshes to reduce unnecessary backend refetching, with cleanup to prevent delayed refreshes after leaving the page.

* **Tests**
  * Added coverage for postings merge/dedup behavior and stream-driven invalidation throttling, including timing windows, unmount cleanup, and handling rejected invalidation requests.


saeedkolivand:
🎉 This is included in version 0.113.0.

PR #483, opened 2026-06-23
fix(autopilot): forward country to aggregator + stop keyword-prefill (with dep bump, landing OG fix)

saeedkolivand:
## Why

Running an **autopilot** returned **zero jobs** from the **aggregator** board, while a **manual** job search returned jobs for the *same query*. Two independent divergences between autopilot and manual search each zeroed out the aggregator results.

### 1. Autopilot defaulted the aggregator to Germany (`fix(autopilot)`)
The aggregator's Adzuna provider puts the country directly in its request path (`/v1/api/jobs/{country}/search/1`). Manual search captures `countryCode` from the location **geocode suggestion** (`ScrapeFilters.tsx`) and forwards it. Autopilot had **no country field** — it hardcoded `country_code: None`, so Adzuna defaulted to `"de"` and any non‑German search returned nothing.

- Threads an **optional** `countryCode`/`country_code` end‑to‑end: shared Zod `AutopilotTargetSchema`, the regenerated IPC contract (both create **and** update targets), Rust `AutopilotTarget`, and `autopilot_helpers` forwarding.
- The creation wizard's `LocationInput` now captures the geocode `countryCode` (with `{ shouldDirty: true }`), mirroring manual `ScrapeFilters`.
- **Backward compatible:** the field is absent‑by‑default — old persisted autopilots deserialize to `None`, and behavior is unchanged when unset.

### 2. The "Must include" keyword filter was pre‑filled with the whole tech stack (`fix(autopilot)`)
The backend's `matches_keyword_filters` requires **all** keywords to appear in a job, and the wizard pre‑filled that field with the user's **entire tech stack** — so nearly every posting was dropped. (This was dead config until it was recently activated, which is when autopilot started returning none.)

- `buildDefaults` now leaves "Must include" **empty** (opt‑in). Removed the now‑dead prefill plumbing (`Prefilled.keywords`, the `PrefilledBadge` usage in `StepFilter`) and the orphaned `fromTechStackSettings` i18n key (en + de).

### 3. Landing `og:description` had a stray authoring note (`fix(landing)`)
The `og:description` meta tag contained a leftover note — `(the OG image should be the deep-fried 'IT DOES EVERYTHING ELSE' frame)` — which link‑preview unfurlers rendered verbatim. Removed it so it matches the already‑clean `twitter:description`.

### 4. Dependency sweep (`chore(deps)`) — bundled at request
Batch dev + runtime dependency bump (commitlint, @types/node, eslint, knip, lint-staged, typescript-eslint, jsdom, playwright, vite, tailwindcss, @tauri-apps/* api+plugins, @tanstack/*, @tiptap/*, lucide-react, motion, react-hook-form, zustand, @wxt-dev/browser). `pnpm-workspace.yaml` adds `minimumReleaseAgeExclude` entries to opt the freshly‑released typescript-eslint 8.62.0 family / globals / knip out of the new‑release cooldown.

> ⚠️ Reviewer note: the larger jumps are **`@wxt-dev/browser` 0.1.43→0.2.0** (pre‑1.0, extension runtime) and **`jsdom` 26→29** (test env). The full `typecheck` / `lint:strict` / `cargo` / `vitest` suites pass with these versions. The Playwright 1.61 bump requires `pnpm exec playwright install chromium-headless-shell` locally (CI installs browsers fresh).

## Tests
- **Rust:** `AutopilotTarget` serde round‑trip — absent `countryCode` → `None` (backward‑compat) and `Some("us")` survives camelCase round‑trip with `None` omitted.
- **Renderer:** `buildDefaults().keywords === ''` regression lock; `wizardStateToPayload` / `autopilotToWizardState` `countryCode` round‑trip; `StepTarget` captures `countryCode` from `onSelectSuggestion` (incl. `null → undefined` coercion).

## Verification
`typecheck` 12/12 · `lint:strict` 0 · `gen:ipc:check` clean · `cargo test autopilot` 45 · renderer vitest 1497 · full `pnpm test` green.

Manual: create an autopilot for a **non‑German** location (pick a geocode suggestion), aggregator board, Adzuna keys set → it returns jobs; the "Must include" field is empty by default. German searches still work (country `"de"`).

## Summary by CodeRabbit

* **New Features**
  * Added optional country code support to Autopilot targets, wired from location suggestion selection and round-tripped end-to-end.
  * Autopilot cards now show “board(s)” with correct pluralized board count.
* **Improvements**
  * Simplified the wizard filter step by removing prefilled keyword indicators.
  * Applying a saved posting now preserves board provenance when available, otherwise falls back to the first target board, then the aggregator.
* **Bug Fixes**
  * Improved handling of unset country codes and legacy records; board data is preserved during re-surfacing.
  * Aggregator board now surfaces clearer errors for unsupported markets and provider failures.
* **Chores**
  * Updated dependencies and refreshed Autopilot translations.


saeedkolivand:
@coderabbitai review

saeedkolivand:
Thanks @coderabbitai — dispositions below (2 fixed, 3 with reasoning).

### ✅ Fixed
- **`StepTarget` stale `countryCode` on free-text edit** — the location `onChange` now clears `countryCode` (`setValue('countryCode', undefined, { shouldDirty: true })`). Safe on a suggestion pick because `LocationInput.select()` fires `onChange` *before* `onSelectSuggestion` (verified at `packages/ui/.../LocationInput.tsx`), so a pick clears-then-sets; a manual re-type clears the stale code. (commit `85e3940`)
- **`minimumReleaseAgeExclude` undocumented** — added a comment explaining the entries opt freshly-released tooling out of the cooldown and are transient. (commit `3dfca26`)

### ⏭️ Skipped, with reasons
- **ISO-3166 validation on `countryCode` (`schemas/index.ts:356`)** — intentionally left as `z.string().optional()` for **parity** with the manual-search schema (`ScrapeBoardsRequestSchema.countryCode`, line 146, also unvalidated). The value is sourced from the geocoder suggestion (`s.countryCode`), not free user text, and the repo prefers relaxed validators for forward-compat. Adding a constraint to only one of the two sibling schemas would be inconsistent; if we want it, it should land on both in a dedicated change.
- **Unify `Autopilot.target` to `boards` (`types/index.ts:208`)** — valid and worth doing, but it's **pre-existing drift orthogonal to this fix** and a genuine "heavy lift": `AutopilotPage`/`AutopilotCard` read singular `ap.target.board` (which is already `undefined` at runtime since the backend emits `boards`), so the migration also needs those display consumers fixed + tested. Deferring to a focused follow-up PR to keep this bug-fix reviewable; `countryCode` is already typed on the same `target` shape, so it has compile-time coverage today.

### 🔒 Security finding (TanStack supply-chain) — verified, not affected
The May 11 2026 incident is real, but cross-checking the regenerated lockfile against the published malicious set: the compromised versions were `@tanstack/react-router`/`router-core` **1.169.5 / 1.169.8** and `router-plugin` **1.167.38 / 1.167.41**; this PR resolves **react-router 1.170.16**, **router-core 1.171.13**, **router-vite-plugin 1.167.18** (all *newer* than the malicious burst), and `@tanstack/virtual-core` / `store` / `react-virtual` are in the **confirmed-clean** families. The lockfile was also regenerated ~6 weeks post-incident. No affected version is present.

saeedkolivand:
Update on the `board`→`boards` finding: **folded into this PR** (commit `a0f7472`), not deferred. The shared `Autopilot.target` type is now `boards: string[]` (matching the Rust struct + Zod schema), `AutopilotCard` renders the board badge from the array (localized single-board label, or `{{count}} boards` for multiple), `AutopilotPage` passes `boards[0] ?? AGGREGATOR_BOARD_ID`, and the `as unknown as` cast in `wizard-state` is gone. This also fixes the latent blank-badge bug (the old singular `board` was undefined at runtime). typecheck / lint:strict / 1497 renderer tests green; reviewed by frontend-reviewer + rust-backend-architect (contract lockstep confirmed).

saeedkolivand:
Addressed @coderabbitai — and the trail led to the actual root cause.

### ✅ Per-job board provenance (`AutopilotPage` 79-84) — fixed (commit `49e6545`)
`FoundJob` (Rust) + `AutopilotFoundJob` (TS) now persist the job's source board (from `JobPosting.source`), and the apply call uses `job.board ?? boards[0] ?? AGGREGATOR_BOARD_ID`. The dedup-merge preserves/refreshes `board` like `location`/`description`, and old persisted records deserialize to `None` (serde default). Reviewed by rust-backend-architect + frontend-reviewer.

### 🎯 Bonus: the deeper "aggregator returns zero" root cause — also fixed (commits `8f15bd3`, `485f367`)
Investigating provenance surfaced *why* autopilot returned zero in the first place: Adzuna's API is **path-based per country** (`/jobs/{country}/search/1`) and only hosts ~19 markets. The autopilot wizard forwards a geocode country; if it isn't an Adzuna market, the request 404'd and — with no JSearch key — was swallowed into a silent empty result. (Manual "just a title, no location" worked only because no country is sent, so Adzuna defaults to `"de"`.)

Fix in `scraping/boards/aggregator/mod.rs`:
- `ADZUNA_SUPPORTED_COUNTRIES` allowlist short-circuits the doomed call; unsupported countries fall through to JSearch (global) when configured.
- When neither provider can serve the country, an **actionable diagnostic** surfaces via `BoardScrapeSummary.error` ("add a JSearch key in Settings for global coverage") instead of a silent zero. The keyless-empty case (no provider configured at all) is unchanged.
- Fixes both the manual and autopilot paths (shared board). 8 new tests; reviewed by scraping-applier-expert + testing-reviewer.

All gates green (typecheck / lint:strict / gen:ipc:check / cargo test / clippy / full `pnpm test` via the pre-push gate).

saeedkolivand:
Latest review addressed — 2 fixed, 1 reasoned skip.

### ✅ Fixed
- **`countryCode` ISO-3166 alpha-2 validation** (commit `2a26503`) — applied `z.string().trim().regex(/^[A-Za-z]{2}$/).optional()` to **both** `ScrapeBoardsRequestSchema` and `AutopilotTargetSchema`. Verified every caller is geocode-sourced (alpha-2); no fixture or IPC caller sends a non-2-letter value, and `gen:ipc:check` stays clean.
- **Empty-boards fallback test** (commit `e2c7b11`) — locks `autopilotToWizardState` → `[AGGREGATOR_BOARD_ID]` when `target.boards` is empty.
- **Tightened the diagnostic assertion** (commit `…test`) — now requires the provider prefix `adzuna:`, the searched country `'xx'`, AND the "not in Adzuna's supported market list" phrase together (was a 3-way `||`).
- **Docs no longer embed the allowlist** (commit `…docs`) — `scraping-domain.md` now points to `ADZUNA_SUPPORTED_COUNTRIES` instead of listing the 19 codes.

### ⏭️ Skipped, with reason: typed `AppError` for the new aggregator error sites
The `Scraper` trait (`scraping/types/mod.rs:124`) and the aggregator-internal `JobProvider` trait both return `anyhow::Result`, not `AppResult`/`AppError`. The engine stringifies every board error at `engine/mod.rs:465` (`e.to_string()`) into `BoardScrapeSummary.error: Option<String>` — it is never pattern-matched downstream, so a typed variant would be erased immediately. The file already has 4 pre-existing `anyhow::anyhow!` sites (lines 208, 252, 351, 392) identical in form to the 2 new ones; converting only the new sites would create an inconsistent mix in a file with a single `anyhow::Result` contract. `AppError` appears in the scraping layer only in the internal HTTP infra (`scraping/http`, LinkedIn client), where it's converted back to `anyhow` before crossing the trait boundary — there's no board/provider-layer typed-error convention to conform to. Converting would be a layer-wide refactor out of scope for this PR.

All gates green via the pre-push suite (typecheck / lint:strict / gen:ipc:check / cargo test / clippy / full `pnpm test`).

saeedkolivand:
🎉 This is included in version 0.115.1.

PR #488, opened 2026-06-24
chore: harden agent system and pr-reviewer from the 486 post-mortem

saeedkolivand:
## Summary

Post-mortem hardening from **#486**, where `pr-reviewer` passed it (only 1 🟠) but CodeRabbit + CI then surfaced 37+ findings and several CI round-trips. **Root cause:** the gate (and the authors) ran `cargo test` on the Windows dev host, which silently `#[cfg]`-excludes Linux code/tests — so Linux-only compile errors (unused import, `#[serial]` scope, `pub(super)` visibility) and env-fragile / `exec()`-reaching tests (one of which **cancelled** the CI Tests job) only appeared on CI. This PR closes the gap at **both** the last gate *and* upstream in the author skills, so the issue is prevented at write-time.

### `pr-reviewer` agent (`.claude/agents/pr-reviewer.md`)
- **Cross-OS / `#[cfg]` blind spot** — a same-host build does **not** verify other-OS gated code/tests; require a cross-target check and append `· UNVERIFIED (cross-OS)` to the verdict so a green local run is never read as a green CI run.
- **New lenses** — React Query invalidation-key prefix-match (the trailing-`undefined` trap that shipped twice), nullable-at-IPC-boundary, cache/derived-state invalidation completeness (embeddings + hash-keyed scores), bounded-external-process, env-mutation scoping + rollback, parallel-table/registry consistency.
- **Test quality** — no host/environment-coupled tests, no `exec()`/process-replacement in-process, `#[serial]` env-mutating tests.

### Author skills (root-cause fix — prevent at write-time)
- **`author-contract`** (every author loads it) — hermetic-test rules + the cross-OS-build caveat, so the rule reaches even authors that don't load `testing-rules`/`rust-standards` (e.g. `extension-author`, which wrote the env-fragile Flatpak test).
- **`testing-rules`** — cfg-gated tests run only on that OS's CI; never assume a system binary/lib is absent; never reach a process-replacing path in-process; `#[serial]` env tests.
- **`rust-standards`** — `#[cfg(target_os)]` not compiled on other hosts (cross-check); env/process-mutation scoping + rollback; bounded external processes (real timeout).
- **`frontend-standards`** — React Query invalidation-prefix, `mutateAsync`-catch / optimistic-after-success, nullable-at-IPC-boundary.
- **`extension-standards`** — native-messaging manifest for **every** detected browser (the Vivaldi gap) + a cross-ref that bridge/detection Rust + tests also obey `rust-standards`/`testing-rules`.
- **`job-match-standards`** — invalidate the cached embedding/score when a posting's text changes.

### Explainer drift fix
- `landing/agent-system.html` said **23 agents** (meta + animated counter); there are **24** (matches CLAUDE.md). The drift guard only checks per-agent cards, not the prose count, so it slipped through. Corrected to 24.

### Verified
- `pnpm check:agent-system` ✓ and `pnpm check:landing-drift` ✓ (agents ⇄ routes ⇄ CLAUDE.md ⇄ docs ⇄ landing in sync).
- The `createMockClient` dotted-key "false positive" CodeRabbit re-raised was **already** recorded in `review-config.md` learnings (#477) — no change needed there.

## Summary by CodeRabbit

* **Documentation**
  * Expanded pre-review and Rust standards guidance for cross-OS/cfg-gated blind spots, hermetic tests, and safe environment/process mutation rollback.
  * Added requirements for bounded external command timeouts and stricter verdict marking when cross-OS code can’t be compiled/run.
  * Strengthened React/React Query guidance for exact invalidation-key matching, async mutation error handling, and null-safety at IPC boundaries.
  * Updated cache invalidation rules when job posting resolved text changes and added browser detection ↔ native messaging manifest lockstep guidance.


saeedkolivand:
🎉 This is included in version 0.116.1.

PR #490, opened 2026-06-24
fix(landing): resolve the re-audit second-pass a11y findings and 2 regressions

saeedkolivand:
## Summary

Second-pass follow-up to **#489**. A re-audit (the "verify" half of the audit) confirmed #489 resolved the original 45 findings (**70 resolved**) but surfaced a deeper layer — **4 new HIGH**, ~26 medium/low, and **2 regressions the first wave introduced**. This PR fixes them. The 4 HIGH + both regressions were independently verified cleared before opening this.

### Regressions from #489 (fixed)
- **social-card** — the new magenta `#ff00d4` focus ring failed non-text contrast (~1.3:1 on the amber gradient) → **white ring + black halo** (WCAG 2.2 SC 1.4.11). Magenta stays as the brand glow, not the focus ring.
- **creature** — `.tnote` red `#e24b4a` on paper failed AA → `#b5160e` (darker red).

### New HIGH (fixed)
- **architecture-map** — `#kbd-help` had `aria-modal` but no focus trap → focus trap + Esc-to-close + return-focus to opener.
- **how-it-works** — `#nodeDetail` runtime `innerHTML` region → `aria-live="polite"` + `aria-atomic`; `nav` gets `aria-label` + `aria-current`; IPC table gets a caption.
- **ci-dashboard** — `#global-search` had only a placeholder → `aria-label`; focus ring recolored to near-white (~16:1 on dark).
- **privacy** — `.anchor` red focus ring (~3.19:1) → ink (~12.6:1).

### Medium / low
- SMIL `<animate>` elements gated under `prefers-reduced-motion` via JS (CSS reduced-motion doesn't reach SMIL); footer/anchor contrast bumps; download `.auto-note` brew commands made copyable for consistency; misc `aria-atomic`/labels.

### Verified
- All 4 HIGH + both regressions confirmed cleared (focus-ring colors recolored + contrast re-checked, focus trap present, aria-live added); `check:landing-drift` ✓ · `check:agent-system` ✓ (count 24); visual identity unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #492, opened 2026-06-24
fix(landing): restore the creature film (curly-quote regression)

saeedkolivand:
## Summary — urgent: the creature film stopped playing

The wave-2 (#490) `#endcard` dialog edit replaced straight quotes with **curly quotes** (`”`) across the block, so `id="endcard"`, `id="replay2"`, etc. didn't parse. `$('replay2')` became null → **`Cannot read properties of null (reading 'addEventListener')`** at the wiring step → the film script halted and the page showed only the paper background + text.

**Fix:** restored straight quotes on all `#endcard` attributes (kept the intentional typographic `“send”` in the prose). `$('endcard')`/`$('replay2')` resolve again → the film plays.

This PR is `landing/creature.html` only (its favicon, also missing, rides along since it's the same file) — split out from the cosmetic links/favicons PR so it can merge on its own.

### DIFF
--- a/landing/creature.html
+++ b/landing/creature.html
@@ -4,6 +4,7 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <title>THE CREATURE — a hand-drawn doodle short film</title>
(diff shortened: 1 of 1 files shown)


saeedkolivand:
🎉 This is included in version 0.116.1.

PR #500, opened 2026-06-25
feat: github project import, job-summary fixes, and answer rewrite

saeedkolivand:
## Summary

This branch bundles one new feature plus several related fixes/improvements (folded into one PR by request).

### ✨ GitHub project import (resume builder)
Adapts the one genuinely useful, seeker-relevant idea from HackerRank's `interviewstreet/hiring-agent` (the rest didn't transfer — it's a Python, employer-side candidate-scoring tool). Adds an **"Import from GitHub"** action to the resume builder's Projects step:

- New `github_import_repos` IPC capability (`profile_import/github.rs`): fetch a user's public repos, drop forks, sort by stars (top 30). Hardened SSRF guard (host-gated parse, server-constructed `api.github.com` URL, username validated against GitHub's rule), routed through the shared `scraping::http::fetch_text` for the 8 MB body cap + a 20 s timeout + per-host rate limiting.
- AI bullets via the shared generation pipeline (no new IPC, every provider): repo text is fenced as untrusted (ADR-010 pattern), the model never sees/writes URLs (re-attached in code, matched to repos by de-slugged name so a reordered response can't cross a bullet onto the wrong link), with a raw-description fallback offline.
- `GitHubImportModal`: prefill from the contact profile, fetch/empty/error/generating states, accessible dialog + live regions, design tokens; cancel preserves selection.

### 🛠 Job-summary fixes
- Render the shared **ModelSelector** on the job-summary step (same preferences store as `StepModel` — no second model state).
- **Regenerate the summary when the output language changes** (guarded: only when a summary exists, aborts any in-flight run, no loop).

### ✨ Rewrite application answers with AI
- A "Rewrite with AI" affordance per generated application answer, reusing `RewritePopover` with a new `application-answer` `RewriteDocType` (prose voice, same grounding/no-fabrication contract). Accepted rewrites persist through the existing save path.

### 📝 Docs / tooling
- README tagline now lists **CLI agents** (Claude Code, Codex, Gemini CLI) alongside Ollama / OpenAI / Anthropic / Gemini; README TOC opens by default.
- Hardened the internal `pr-reviewer` agent with rules distilled from recent CodeRabbit findings (static-init panics, `json!(struct)` IPC drift, React Query `gcTime`/over-broad invalidation, tautological-assertion patterns).

## Testing
- Rust: SSRF-rejection + status-mapping + filter/sort unit tests (network-free).
[...]

PR #501, opened 2026-06-25
chore(agents): require gate-scoped validation and red-green tests

saeedkolivand:
## Why

During PR #500 (a large multi-agent change), domain authors repeatedly reported "green" from **scoped** checks that the full/forced pre-push + CI gate then failed on. Three distinct classes:

1. **Scope** — a `@ajh/shared` "every expected namespace" test broke when the new `github` IPC namespace was added, but the author only ran `pnpm -F @ajh/tauri test` (never the shared package's tests).
2. **esbuild ≠ tsc** — test-file `getByTestId(...).value/.checked` casts + `noUncheckedIndexedAccess` passed `vitest` (esbuild does no type-checking) but failed the real `tsc`.
3. **Mock-masking** — a regression test passed against broken logic because the mocked `updateAnswer` didn't reproduce production's optimistic `setAnswers`; CodeRabbit caught it.

The through-line: an author's "done" bar was narrower than the bar that actually gates.

## What

Small `.claude/**` config edits — no shipped-code change:

- **`author-contract`** — the report-green bar must match the pre-push scope: full `pnpm test` for `packages/shared`/contract/cross-package changes, `TURBO_FORCE=1 pnpm typecheck` (force past a cached `^build`), and a real `tsc` after any test edit (`vitest` is not a typecheck). `-F <pkg>` is for fast iteration only.
- **`testing-rules` + `testing-reviewer`** — mock fidelity (a stub must reproduce the real side-effect of what it replaces) and red-green (a regression test must fail on the unfixed code).

A matching lesson was logged locally.

## Summary by CodeRabbit

* **Documentation**
  * Clarified testing guidance to better catch missing coverage, weak assertions, over-mocking, and guard tests that don’t fail without the fix.
  * Expanded validation instructions with clearer checklist-style steps for running the right checks, including full-scope verification for shared changes and explicit typecheck expectations.
  * Added stronger guidance for mocks to better reflect real behavior and side effects.

### DIFF
--- a/.claude/agents/testing-reviewer.md
+++ b/.claude/agents/testing-reviewer.md
@@ -26,7 +26,7 @@ You are the **testing-reviewer** — you **review, critique, audit, and challeng
 
 ## Actively hunt for
 
-missing coverage · weak assertions · flaky tests · untested edge cases · untested error paths · untested security scenarios · untested performance scenarios · over-mocking · redundant tests.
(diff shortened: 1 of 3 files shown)


saeedkolivand:
🎉 This is included in version 0.118.1.

PR #503, opened 2026-06-29
ci: align format-guard prettier pin to 3.9.1

saeedkolivand:


saeedkolivand:
🎉 This is included in version 0.119.0.

PR #507, opened 2026-06-29
docs: correct extension import mode, disclose native-messaging, note rate budget

saeedkolivand:
## Summary by CodeRabbit

* **Documentation**
  * Updated browser extension docs to describe a single **“Import this job”** flow instead of separate import modes.
  * Clarified that imports try to capture the page content when possible, and automatically fall back to **URL-only** on restricted pages.
  * Added guidance for the **“I already applied”** checkbox and updated setup, testing, and privacy notes to match the new experience.
  * Noted that rate-limited imports now show an error instead of waiting silently.


saeedkolivand:
@claude review

saeedkolivand:
🎉 This is included in version 0.119.0.

PR #533, opened 2026-07-02
build: exempt fresh deps from minimumReleaseAge cooldown and bump @types/node + motion

saeedkolivand:
## What

Adds 3 recently-published dependency versions to `minimumReleaseAgeExclude` in `pnpm-workspace.yaml`, so they pass the supply-chain `minimumReleaseAge` cooldown (~24h) instead of failing fresh installs.

## Why

`@semantic-release/github@12.0.9` (published 2026-07-01 15:25) and `lucide-react@1.23.0` (2026-07-01 14:05) were merged to `main` (via #520 / #528) **before** they aged past the ~24h cooldown window. That means every fresh `pnpm install` — on `main`, on every open PR, and in CI — now fails the `minimumReleaseAge` supply-chain policy, cascading the frontend/install-dependent checks (Playwright, Lighthouse, Lost Pixel, a11y, Knip, Dependencies) red. That's the true root cause of the "red" Dependabot PRs (#519, #524) — not their own bumps.

This keeps the security policy intact (the intended escape hatch) rather than removing it. Also pre-adds `@types/node@26.1.0` (the pending #519 bump) so #519 passes the moment it rebases onto this.

Verified locally: `pnpm install --lockfile-only` goes from "2 lockfile entries failed verification" → "✓ Lockfile passes supply-chain policies". No lockfile change (settings-only).

## Transient

Per the file's own note, each entry can be removed once that version ages past the cooldown (lucide-react ~07-02 14:05, the others ~07-02 15:25 / 11:04).

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -38,7 +38,7 @@
     "@tauri-apps/plugin-store": "^2.4.3",
     "@tauri-apps/plugin-websocket": "^2.4.2",
     "lucide-react": "^1.23.0",
-    "motion": "^12.42.0",
+    "motion": "^12.42.2",
     "react": "^19.2.7",
     "react-dom": "^19.2.7",
     "react-hook-form": "^7.80.0",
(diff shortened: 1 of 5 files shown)


saeedkolivand:
🎉 This is included in version 0.120.0.

PR #545, opened 2026-07-04
fix: surface newly found autopilot jobs at the top of the list

saeedkolivand:
## Problem
When an autopilot run finds new jobs, they appeared at the **bottom** of the autopilot's found-jobs list. Users expect newly discovered jobs to surface first.

## Root cause
`merge_found_jobs` (`apps/desktop/src-tauri/src/autopilot/mod.rs`) **appended** never-before-seen URLs to the end of the cumulative found-jobs list, and `AutopilotCard` renders that `Vec` verbatim (no client-side sort). So new finds rendered last.

## Fix
Build the merged `Vec` **new-first**: genuinely-new URLs (kept in the already score-sorted incoming order) first, then the refreshed existing rows. Existing rows keep their prior order, their `found_at`/first-seen time, and their `is_new`-cleared state. This mirrors the file's own convention — `write_to_disk` already orders the autopilots themselves newest-first.

No frontend / schema / IPC / migration change — the renderer shows the `Vec` order directly, and the "New" badge is unaffected. No other consumer depends on the old order (`record_run` counts `is_new`, `enrich_applied` only sets `applied`, export dumps the list as-is, the tray uses a count).

## Tests
- Added `merge_puts_newly_found_jobs_on_top` locking the new positional order (previously untested — existing merge tests are order-agnostic).
- Fixed stale "appended" wording in `merge_keeps_prior_jobs_not_in_the_new_run`.
- `cargo test -p ajh-tauri --lib autopilot` -> 92 passed, 0 failed. `cargo clippy` / `check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Newly discovered job postings now appear at the top of the job history list.
  * Previously seen jobs are still retained, with their status updated correctly during a refresh.
  * New items are clearly marked as new, while older items keep their existing placement below them.

### DIFF
--- a/apps/desktop/src-tauri/src/autopilot/mod.rs
+++ b/apps/desktop/src-tauri/src/autopilot/mod.rs
@@ -653,7 +653,8 @@ fn u32_field_in_range(v: &serde_json::Value, key: &str, max: u32) -> Option<u32>
 /// - existing rows are kept (preserving `found_at`/first-seen) and have `is_new`
 ///   cleared; if the run re-surfaced them, volatile fields (title/company/
 ///   description/score/location) are refreshed,
-/// - postings whose URL was never seen are appended and flagged `is_new`.
+/// - postings whose URL was never seen are placed first (on top) and flagged
+///   `is_new`, in the incoming (already score-sorted) order.
 ///
 /// Re-running with the same postings yields the same set — only `is_new` moves.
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@coderabbitai review

saeedkolivand:

PR #546, opened 2026-07-04
fix: restore autopilot list to the applied job on back-navigation

saeedkolivand:
## Problem
From an autopilot's found-jobs list, pressing **Apply** on a job deep in the list opens `/applications/$id`. "Back to Autopilot" does a fresh `navigate({to:'/autopilot'})`, so `AutopilotPage` remounts and the inner `overflow-y-auto` found-jobs list resets to the **top** — the user loses their place. (The card was already re-expanded via `lastAppliedId -> focusedId`, but only the autopilot id was known, so the card *header* was centered, not the job.)

## Fix
Extend the existing focus channel to also carry the applied job's url:
- Session store gains `lastAppliedJobUrl` / `focusedJobUrl` (default `null`).
- `handleApply` records `lastAppliedJobUrl: job.url`; the return effect promotes it to `focusedJobUrl` and passes it to the focused card.
- `AutopilotCard` tags each row with `data-job-url` and, once the found-jobs panel finishes expanding (`onAnimationComplete`) — or immediately via `requestAnimationFrame` if the card is already expanded — scrolls that row into view (`block:'center'`) and briefly highlights it (`ring-brand`, design token, no hex).
- Scroll resolution is **idempotent** (a ref is cleared first), so the `onAnimationComplete` and rAF paths can never double-fire, and it always calls `onFocusHandled()` so `focusedJobUrl` can't wedge in the store. The tray/deep-link focus path is unchanged (still centers the header).

Reuses the existing focus mechanism rather than adding a parallel one; mirrors `ApplicationRow`'s just-imported highlight idiom.

## Tests
- `AutopilotPage.focus.test.tsx`: `lastAppliedJobUrl` is promoted to `focusedJobUrl`, forwarded to the focused card, and cleared.
- `AutopilotCard.test.tsx`: asserts the exact matching row (not the header) is the scroll target, the highlight applies/fades, `onFocusHandled` fires once, header fallback when `focusedJobUrl` is absent, and the already-expanded (rAF) path.
- `vitest AutopilotCard AutopilotPage.focus` -> 36 passed. `typecheck` clean. `eslint --max-warnings 0` clean.

### DIFF
--- a/apps/desktop/src/renderer/features/autopilot/components/AutopilotCard/AutopilotCard.test.tsx
+++ b/apps/desktop/src/renderer/features/autopilot/components/AutopilotCard/AutopilotCard.test.tsx
@@ -13,7 +13,7 @@
  */
 
 import React from 'react';
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { act, render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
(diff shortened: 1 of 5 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks for the review — all four are advisory; dispositions below, no code changes:

- **LOW · duplicate `job.url` → wrong row:** `found_jobs` are deduped by `url` in `merge_found_jobs` (url is the identity), so the list can't hold distinct duplicate urls. If a single scrape batch ever repeats a url, the two rows are the *same job*, so first-match is correct anyway. Keeping the scroll keyed on `url` (not index) is deliberate: the index is stale if the list re-ran/reordered while the user was on the application page, whereas the url is durable across a re-run. Leaving as-is.
- **LOW · no-op when the applied job is gone:** intended fail-safe — `setHighlightedUrl` + `onFocusHandled` still fire so focus can't wedge; no scroll, no error.
- **NIT · 1500 highlight duration:** `ApplicationRow` uses a local literal too, so there's no shared constant to reuse; not extracting one for two call-sites.
- **NIT (test) · mock fires onAnimationComplete on mount:** acknowledged coupling; only the found-jobs panel passes the prop today, and the added already-expanded test covers the rAF path.

saeedkolivand:

PR #548, opened 2026-07-04
feat: give a paste-ready number for salary-expectation answers

saeedkolivand:
## Problem
Application forms' salary fields often accept **only a number**, but the "What are your salary expectations?" application answer produced prose with no clean figure to paste. (First of two PRs — this one grounds the number in the user's saved expectation; a follow-up adds live web research for a market range.)

## Change
- **Data-only registry guidance.** Add an optional `guidance?: string` to `ApplicationQuestion`; `buildApplicationAnswerPrompt` appends it. No per-question code path — adding guidance stays a pure data edit. The `salary` entry's guidance asks the model to end with a single `Number: <integer>` line (digits only, no symbol/separators) derived from the candidate's stated `salaryExpectation`, and to state a range in prose only when their expectation was itself a range.
- **System-prompt relaxation, safety-first.** Rule 4 previously blanket-forbade any salary number. It now permits a concrete figure **only when `<applicant_details>` lists a salary expectation** (condition front-loaded so small local models don't over-weight it); with no expectation it stays non-committal ("open to discussing") and **never invents a number**. "NEVER fabricate/round beyond what is stated" and the "may be submitted automatically" caution are preserved.
- Threaded `guidance` through `generateApplicationAnswer` → `useApplicationAnswers`. No new provider/web capability; no schema change; the answer stays free-text `ApplicationAnswer`.

## Safety
Reviewed by the AI-provider expert: no fabrication path — three independent layers (system rule, per-question guidance, closing user line) all fall back to non-committal when no expectation is set; `buildApplicantDetailsBlock` emits nothing for an empty expectation. The `Number:` line survives `extractPlainText` and lands copy-pasteable in the stored answer.

## Tests
`packages/prompts` — guidance is appended only for the salary question (and `undefined` for others); the relaxed system prompt still forbids fabrication; the salary guidance's own fallback wording (`never invent a figure`, `omit that final line`) is asserted directly. `useApplicationAnswers` — guidance forwarding verified against the real registry.
- `pnpm --filter @ajh/prompts test` → 352 passed. `typecheck` / `eslint --max-warnings 0` / `prettier` clean.

### DIFF
--- a/apps/desktop/src/renderer/features/documents/components/TailorFlow/useApplicationAnswers.test.ts
+++ b/apps/desktop/src/renderer/features/documents/components/TailorFlow/useApplicationAnswers.test.ts
@@ -3,6 +3,8 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { act, renderHook } from '@testing-library/react';
 
+import { generateApplicationAnswer } from '@/lib/generate';
+
 import { useApplicationAnswers } from './useApplicationAnswers';
 
 // Stub the generation lib: metadata + one deterministic answer, no research.
@@ -43,7 +45,10 @@ const wrapper = ({ children }: { children: ReactNode }) =>
 const render = () => renderHook(() => useApplicationAnswers(base), { wrapper });
 
 describe('useApplicationAnswers', () => {
-  beforeEach(() => save.mockClear());
+  beforeEach(() => {
+    save.mockClear();
+    vi.mocked(generateApplicationAnswer).mockClear();
+  });
 
   it('toggles selection and gates generation on a non-empty selection', () => {
     const { result } = render();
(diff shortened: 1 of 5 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks — addressing all three (pushing a follow-up commit):

- **MEDIUM (non-numeric-but-set expectation):** fixing. Free-text values like "competitive"/"negotiable"/"DOE" now take the non-committal path too — both the system rule and the salary guidance fall back to "omit the Number line, never invent" when the stated expectation contains no actual number, not only when it's absent.
- **LOW (range → single Number):** pinning to the **upper bound** of the applicant's own stated range (their stated target, grounded and user-editable) so the Number line is deterministic — consistent with the "target number" choice over "low end".
- **LOW (inline cast):** replacing `(q as { guidance?: string })` with a safe `'guidance' in q` narrow so the field stays type-checked.

New tests lock the non-numeric fallback and the upper-bound rule.

saeedkolivand:

PR #549, opened 2026-07-04
feat: research the market salary range for salary-expectation answers

saeedkolivand:
## Problem
The salary application answer (C1) grounded the paste-ready number only in the candidate's saved expectation. Users wanted it to **research the market salary range** for the role/company so the number is competitive. (Second of the two salary PRs.)

## Change
- **`AiProvider::research_salary(role, company, location)`** reusing each provider's native web search. The three cloud providers' `research()` transports were extracted into a shared per-provider `web_search_complete(system, user)` helper that BOTH company-brief `research` and `research_salary` call — verified **behavior-preserving** for the company-brief path (identical endpoint/body/gating/error paths). Providers without web search (local Ollama, OpenAI-compatible gateways) return empty via the trait default → new providers stay zero-change.
- **`SalaryResearch` enricher** (mirrors `CompanyResearch`): cache (`KvCache`, ns `salary_range`, 7-day TTL, **case-folded** key `role|company|location`, inputs length-capped) → timeout-bounded provider search → **parse + strictly validate** into `{min,max,currency}` (min>0, max>0, min≤max, max≤100M, currency 3–4 ASCII letters). Only these validated integers + the shape-checked currency ever reach the prompt — **no raw web text** (the prompt-injection boundary, OWASP LLM01). Degrades to `None` on any miss/timeout/parse-fail.
- **`ai_lookup_salary` IPC command** (full 5-step flow), guarded by the same `limiter.acquire` + `charge_provider_daily` budget as `ai_generate` (shared `ai_research` bucket) so a looping/compromised renderer can't drive unbounded paid searches.
- **Prompt:** a fenced numbers-only `<salary_context>` block + a rule-4 carve-out. Precedence for the paste-ready `Number`: **market range present → always a number** — the candidate's own expectation, but floored at the market minimum so they don't undersell (midpoint when no numeric expectation); **no market range → C1 behavior** (expectation number, else non-committal, never invented). Mutually-exclusive branches (a review-caught contradiction was fixed + pinned by tests).
- **Renderer:** the salary question triggers the lookup, double-wrapped so any failure degrades to C1 with the answer still generated (never blocks).

## Reviews (internal, pre-PR)
- **AI-provider expert:** no blockers; confirmed the `web_search_complete` extraction is behavior-preserving byte-for-byte across all four providers.
- **Security (tauri-security-reviewer):** no HIGH/CRITICAL — injection boundary, web egress (no new capability), cache integrity, anti-fabrication, and DoS profile all sound. Its MEDIUM (missing rate-limit on the billable command) + LOWs were fixed in this PR.

## Tests
Rust: `parse_and_validate` (valid / `{}` / malformed / negative / min>max / absurd / bad-currency / 4-letter), cache round-trip + skip-on-empty, empty-role early return, architecture classification. Prompts: the precedence branches (anti-lowball floor / midpoint / non-committal) + still-forbids-invention. Renderer: salary-only trigger + graceful degradation.
- `cargo test -p ajh-tauri --lib` → 1718 passed; `clippy -D warnings` / `fmt` clean. `pnpm --filter @ajh/prompts test` → 362. `typecheck` / `eslint --max-warnings 0` / `prettier` clean.

## Known gap
The `enrich` network timeout/error arms are inspection-verified only — this repo has no `net::http` mock harness (pre-existing). Noted for a future lightweight mock seam.


saeedkolivand:
@claude review

saeedkolivand:
Thanks — dispositions:

- **MEDIUM (cross-currency floor):** fixing (pushing a follow-up commit). The anti-lowball floor/midpoint now applies ONLY when the applicant's stated expectation is in the same currency as the researched `<salary_context>` range; a different-currency expectation is used as-is (C1) with the market range shown separately as context — never a silent cross-currency number. Pinned by a regression test.
- **LOW (cached lookups charge the daily budget + a rate slot):** accepting — mirrors `ai_generate`'s guard-before-work pattern and the 2000/day per-provider cap is generous; the rate slot releases immediately on a cache hit. Intentional for consistency.
- **LOW (≤25s salary-lookup stall):** accepting — best-effort by design, bounded by the 25s cap, never blocks the other answers, and the per-answer loading state already covers the salary answer while it researches.
- **Nits:** the `u32::try_from` after the 100M ceiling is deliberate defense-in-depth; the shared `ai_research` bucket's pre-existing `ai_research_company` retrofit is tracked as a separate follow-up (out of scope here).

saeedkolivand:

PR #551, opened 2026-07-04
feat: ground salary answers in a job's scraped salary range

saeedkolivand:
Completes the salary-answer feature: prefer a posting's own **scraped** salary over the web lookup, and adds the missing test coverage for the provider transports. (Both parked follow-ups from #549, in one PR — as requested.)

## Part 1 — scraped salary → answer (precedence: scraped → web → saved)
Scraped salary is deterministic and free, so it avoids a paid web search when the posting already carries the numbers.

- **Currency derivation:** Adzuna returns `salaryMin/max` with no currency; a `country → ISO-4217` map derives one (unknown country → no currency → graceful fall-back to the web lookup). Poland correctly maps to PLN.
- **Persistence (the risk-bearing part):** `FoundJob` and the `Application` record gain `salary_min/max/currency`. A new **append-only, nullable** migration (#6) adds the columns (`NULL` = unknown, never `0`); existing rows load as `None`. All four positional column-index sites kept in lockstep (hand-verified + tests). Both apply paths (autopilot + manual Jobs) thread salary through.
- **Precedence (renderer):** a **complete, validated** scraped range (positive, finite, `min<=max`, ISO-shaped currency, integer-rounded) builds the `SalaryRange` and **skips** the web lookup; anything malformed **falls through** to the web lookup so the answer never loses its range. `buildScrapedSalaryRange` is a strict superset of the prompt-layer `buildSalaryRangeBlock` guard (a review-caught asymmetry — a scraped `min: 0` used to silently blank the range — fixed + tested). The `<salary_context>` block is now source-neutral.

## Part 2 — provider test seam (closes the "inspection-only" gap from #549)
- Extracted OpenAI's web-search HTTP body behind a pure `supports_web_search` gate predicate; **wiremock** tests (already a dev-dep) cover 500→`""`, non-JSON→`""`, payload→text, and the gate. `salary_research` no longer touches `AppHandle` (dropped from the R2 debt allowlist).
- Injected a `SalarySearcher` trait + `KvCache` into `SalaryResearch::enrich` (one call-site change) so its parse / `{}` / malformed / `Err` / **timeout** / cache-write paths are unit-tested with no `AppHandle`/network. Dev-only `tokio` `test-util` feature; **no new dependency**.

## Reviews (internal, pre-PR)
- **rust-backend-architect** (data/migration): CLEAN — migration appended not renumbered, nullable, transactional; all 24-column indices verified; upsert COALESCE-merge correct; currency map correct.
- **ai-provider-expert** (salary/provider): confirmed the OpenAI extraction + `enrich` DI are behavior-preserving and the prompt logic verbatim-preserved; found the scraped-guard HIGH (fixed above).

## Tests
Migration up + legacy-row-loads-None, salary round-trip/merge, `build_found_job` extraction, currency map; OpenAI transport (wiremock) + gate; `enrich` parse/degradation/timeout/cache; scraped-vs-web precedence incl. 0/negative/Infinity fall-through + decimal rounding.
- `cargo test -p ajh-tauri --lib` + `--test architecture` · `pnpm --filter @ajh/{prompts,shared} test` · renderer suite · `typecheck` / `clippy -D warnings` / `fmt` / `eslint --max-warnings 0` / `prettier` — all green.


saeedkolivand:
@claude review

saeedkolivand:
Thanks — dispositions:

- **LOW (round-then-validate):** fixing (follow-up commit). `buildScrapedSalaryRange` now rounds first and validates the rounded values, so a sub-0.5 raw `min` (→ rounds to 0) falls through to the web lookup instead of blanking the range — restoring the documented strict-superset invariant. Added a sub-0.5 test case.
- **NIT (currency case):** fixing — returns `currency.toUpperCase()` to future-proof against a non-uppercase source.
- **Design note (scraped salary is "sticky" via COALESCE):** accepting — intentional and documented in the upsert comment; a bad one-off scrape is corrected by the next better scrape, never silently cleared.

saeedkolivand:

PR #558, opened 2026-07-06
fix: prevent prep-this-application runs from getting stuck at pending

saeedkolivand:
## What

Fixes the user-reported bug that **"Prep this application" gets stuck at some pending jobs** — the agentic run hangs on a spinner and never completes. Two independent causes, both fixed.

## Root causes & fixes

**1. Renderer race — the permanent hang.** `agent_run` minted the job id, `job_start`ed it, then ran provider/model/résumé/job validation **synchronously** — each failure emitted the terminal `job.failed` `jobs:event` *before* the `{ jobId }` return. The panel only learns the job id (and only then filters events on it) from that return, so a fast validation failure's terminal event was dropped and the state machine sat in `planning` forever.
- **Backend:** move every fail-able validation into the spawned task, so all terminal events (success *or* failure) fire asynchronously — exactly like the happy path. `job_start` + cancel-token registration stay synchronous before the spawn (a fast `jobs_cancel` must never be a no-op); the limiter rejection is the sole remaining synchronous fail (it runs before the token exists), which the renderer reconciles. New `fail_run` helper (`job_fail` + `unregister_token`) at each site.
- **Renderer:** `PrepApplicationPanel` now reconciles against the job's *actual* status via `useJob(runJobId)` once the id is known — if the job is already terminal while the machine is still busy, it drives `COMPLETE`/`ERROR`/`CANCEL` (shared `finishRun`, fires at most once, never overrides a terminated machine).

**2. No per-step wall-clock timeout — long stalls.** The controller raced each provider turn / read-tool call only against `cancel`; a hung or misconfigured OpenAI-compatible `base_url` could block for minutes with no terminal event.
- Add `AGENT_STEP_TIMEOUT` (360s, above the 300s Ollama HTTP timeout) wrapping both step selects, plus `StoppedReason::Timeout` mapped to `job_fail` (never a silent success), so the loop always terminates and always emits a terminal `jobs:event`. The confirm-gate suspension keeps its own `CONFIRM_TIMEOUT`.

## Review

Reviewed by **rust-backend-architect** (token lifecycle, timeout/cancel composition — cancellation still wins the biased select; HIGH-1(a)/(b) invariants intact) and **frontend-reviewer** (byte-exact payload parity between the reconciliation and live-event paths; idempotent both directions). No HIGH/CRITICAL. `cargo fmt` + `clippy -D warnings` + build clean; `cargo test --test architecture` 11 passed; two new `FakeEnv` step-timeout tests; typecheck + `PrepApplicationPanel` suite (19) green. Rust tests run in CI.

## Verify

Trigger prep with a bad/absent provider or a missing résumé → the panel shows an **error**, never a permanent spinner; a slow provider now hits the step timeout and terminates.

### DIFF
--- a/apps/desktop/src-tauri/src/agent/controller.rs
+++ b/apps/desktop/src-tauri/src/agent/controller.rs
@@ -33,6 +33,19 @@ pub const MAX_AGENT_STEPS: usize = 8;
 /// completions per run — stops a loop that keeps calling tools without converging.
 pub const MAX_AGENT_TOKENS: usize = 60_000;
 
+/// Wall-clock ceiling on ONE provider turn or ONE read-tool call (a text-drafting
+/// tool makes its own provider request). Before this fix, the `tokio::select!`
+/// races below raced ONLY against `cancel` — a hung or misconfigured
+/// OpenAI-compatible `base_url` blocked the whole run for minutes with no
+/// terminal event, so `agent_run`'s spawn never emitted a terminal `jobs:event`
+/// and the run looked stuck at pending forever.
+// ponytail: set comfortably above the longest single-call HTTP timeout we ship
+// (`commands::ai_provider::timeouts::OLLAMA_COMPLETION` = 300s), so that
+// timeout's own specific network error surfaces first in the common case; this
+// is the backstop for whatever slips past it (e.g. a custom base_url whose
+// connect/read hangs outside the per-request client timeout).
+pub(super) const AGENT_STEP_TIMEOUT: Duration = Duration::from_secs(360);
+
 /// The fixed, trusted system prompt. NEVER interpolate scraped/user/tool text here.
 /// `pub(super)` so `agent::gate`'s test harness can reuse the exact same prompt
 /// instead of duplicating the literal.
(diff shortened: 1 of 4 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks — addressed the review:

- **LOW #1 (untested reconciliation branches):** added `completed` + `cancelled` cases to the `useJob` reconciliation tests (the `completed` branch reads `job.result` vs the live path's `event.data`), asserting the proposal renders / cancelled state shows and the spinner clears. `PrepApplicationPanel` suite now 21 green.
- **LOW #2 (no overall run deadline, only per-step):** deferring — as noted it's not a regression (each step now emits progress and the loop always terminates, so it never looks "stuck"), and the reported bug is fixed by the per-step timeout. Tracked as a possible future tightening (a top-level run budget) if worst-case wall-clock matters.
- **LOW #3 (doc verbosity):** left as-is given the subtlety of the race the comment documents.

CodeRabbit: no line comments / no actionable items.

saeedkolivand:

PR #559, opened 2026-07-06
fix: repair feature tour on collapsed sidebar, answer text selection, and rewrite popover stacking

saeedkolivand:
## What

Three user-reported renderer bugs (bugs #1, #5, #6 of the batch).

**1 — Feature tour breaks when the sidebar is collapsed.** A collapsed sidebar is *unmounted*, so the tour's `[data-tour-id]` anchors don't exist and the spotlight measures null (broken tour after "Replay wizard"). `OnboardingWizard` now forces the sidebar open before the tour starts **and restores the user's prior collapsed preference** when the tour finishes/skips; `SpotlightTour` re-measures the anchor via `requestAnimationFrame` + `ResizeObserver` so the spotlight lands correctly after the sidebar's `width 0→auto` expand settles.

**2 — Can't select part of a generated answer.** The global `body{user-select:none}` allowlist excluded the answer `<p>`; it now carries `select-text`. "Rewrite" captures the real DOM substring selection (`{selection, before, after}`) and splices the result back at the selected offsets (mirrors `EditableOutput`), falling back to the whole answer when nothing is selected.

**3 — "Rewrite selection" popover renders clipped under the modal.** It was inline inside the modal's `overflow-hidden` panel + transformed stacking context. It now portals to `document.body` (`position: fixed`, anchored off the trigger rect, `z-toast`=700 above the modal's 650), measures its own height, and **flips upward when it wouldn't fit below** (clamped to the viewport) so its Cancel/Accept never render off-screen.

## Review

Orchestrated via a workflow (two parallel authors → frontend-reviewer + ui-ux-expert). ui-ux caught a **HIGH** — the portaled popover had no vertical flip, so on a question near the viewport bottom its actions rendered permanently off-screen — now fixed (measured-height flip-above) and re-verified by ui-ux. A **MEDIUM** (sidebar preference not restored after the tour) was also fixed. One remaining advisory — an unanimated mid-stream re-flip if a streaming result grows the panel past the fits-below threshold — is deferred as non-blocking polish. typecheck + eslint clean; **59 tests** across the 5 affected suites pass.

## Verify

Collapse the sidebar → Settings → Replay wizard → the tour runs with the sidebar expanded and correctly anchored, then the sidebar returns to collapsed. Highlight a substring of a generated answer and rewrite just it. Open Rewrite on a question near the bottom of the list — the popover flips above and stays fully on-screen.

### DIFF
--- a/apps/desktop/src/renderer/components/generation/EditableOutput/RewritePopover.test.tsx
+++ b/apps/desktop/src/renderer/components/generation/EditableOutput/RewritePopover.test.tsx
@@ -85,6 +85,7 @@ vi.mock('@ajh/ui', () => ({
   transition: { fast: {} },
   // useFocusTrap returns a ref object — the mock div accepts it as a plain prop.
   useFocusTrap: () => ({ current: null }),
+  cn: (...classes: unknown[]) => classes.filter(Boolean).join(' '),
 }));
 
 // ── component under test (import AFTER mocks so hoisting picks up the stubs) ──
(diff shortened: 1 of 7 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Addressed the review:

- **MEDIUM (stale save-failed toast):** moved `notify.error` (and the pending-rewrite sentinel cleanup) inside the `pendingRewriteRef` guard, so a superseded rewrite's failure no longer toasts over a newer accepted answer that saved fine. Added a test for it.
- **LOW (unbounded rAF retry):** capped the SpotlightTour anchor-measure at ~60 frames + `console.warn` once on a missing `data-tour-id` (fails loud/cheap instead of busy-looping).
- **LOW (backdrop doc mismatch):** trimmed the RewritePopover doc to match reality (Cancel/Escape; no backdrop in either mode) — no scrim added, per the ui-ux call.
- **LOW (pendingRewriteRef never cleared):** now deletes the sentinel on a successful current save.
- **Deferred (advisory):** the unanimated mid-stream re-flip — non-blocking polish, tracked.

typecheck + eslint clean; 27 tests across the 3 affected suites pass.

saeedkolivand:

PR #560, opened 2026-07-06
fix: ground salary-lookup currency in the job's country

saeedkolivand:
## What

Fixes the user-reported bug: **the salary range shows the wrong currency** — e.g. a job located in Germany rendered USD/CAD instead of EUR (bug #2 of the batch).

## Root cause

The web-lookup salary path let the model **freely choose** the currency: `SALARY_SYSTEM` said "use the local currency for that location", and the user prompt **dropped the location clause entirely when it was blank**. With a weak/empty job-location string the model defaulted to USD or hallucinated CAD. (The scraped-salary path from #551 was already grounded and is unaffected.)

## Fix — ground the currency in the job's ISO country (end-to-end)

- **`countryToCurrency(iso2)`** in `@ajh/prompts` (ISO-3166 alpha-2 → ISO-4217; Eurozone incl. HR/BG + the common markets), beside `countryToMarket`.
- Thread the job country + resolved expected currency from the answer call site → `lookupSalaryRange` → the `lookupSalary` IPC contract → `ai_lookup_salary` → `SalaryResearch::enrich` → the shared `research.rs` prompt builders, which now **pin the currency authoritatively** ("the role is based in {country}; report the range in {currency}") across all provider fan-out builders + the Ollama synth path. Unknown country → byte-for-byte fallback to the prior wording.
- **Defense-in-depth that fails safe:** `reconcile_expected_currency` **drops** a result whose currency still mismatches (returns `None` → graceful no-range) rather than relabeling numbers it can't convert — a wrong paste-ready figure would be *more* misleading than an obvious wrong symbol. A stale wrong-currency cache entry falls through to a fresh grounded fetch. Scraped-salary precedence preserved.

## Review

Orchestrated via a workflow (one author owns the interdependent contract→prompt thread → ai-provider-expert + job-match-expert). **No HIGH/CRITICAL.** job-match-expert verified all currency mappings correct; its MEDIUM — the backstop relabeling without converting — is fixed here (relabel → drop). `gen:ipc:check` clean; typecheck 12/12; **365 `@ajh/prompts` + 3111 renderer tests**; cargo fmt + clippy `-D warnings` + build + architecture (11) green. Rust unit tests run in CI.

## Verify

A German (DE) job's salary lookup now returns EUR; US → USD; an unknown country falls back to prior behavior (no regression). A model that defies the currency pin yields no range rather than a mislabeled one.

### DIFF
--- a/apps/desktop/src-tauri/src/commands/ai.rs
+++ b/apps/desktop/src-tauri/src/commands/ai.rs
@@ -253,13 +253,25 @@ pub async fn ai_research_company(
 /// search, the search yields nothing reliable, or times out — so the salary
 /// answer always falls back to grounding in the applicant's own stated
 /// expectation alone. Only validated integers + a sanitized currency code are
-/// ever returned; raw web text never crosses this boundary.
+/// ever returned; raw web text never crosses this boundary. `country`/
+/// `currency` (resolved client-side from the job's validated ISO country)
+/// ground the reported currency so a blank/weak `location` can't let the
+/// model default to USD or hallucinate a currency — see
+/// `crate::salary_research::SalaryResearch::enrich`.
 #[tauri::command]
+#[allow(clippy::too_many_arguments)]
 pub async fn ai_lookup_salary(
     app: AppHandle,
     role: String,
     company: Option<String>,
     location: Option<String>,
+    // ISO-3166 alpha-2 job country, when known — grounds `currency` below.
+    country: Option<String>,
+    // Authoritative ISO-4217 currency for `country` (resolved client-side via
+    // `countryToCurrency`); `None` when the country is unknown, which
+    // preserves the unconstrained "local currency for that location"
+    // behavior.
+    currency: Option<String>,
     provider: Option<String>,
     model: Option<String>,
     base_url: Option<String>,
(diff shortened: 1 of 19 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Addressed the review:

- **MEDIUM (cache-key contamination):** the salary cache key now folds in the expected currency, so two `Remote` postings in different countries get distinct keys and an unknown-currency read can't collide with a known-country write. `reconcile_expected_currency` stays as the self-heal. Regression test added.
- **LOW (Ollama country ungated):** the Ollama synth/search-query path now gates raw `country` interpolation on a resolved currency, matching the native path. Tests added.
- **Nit (map coverage):** added HU→HUF, RO→RON, and euro microstates SM/VA/AD→EUR.
- **Deferred (advisory):** the 7-arg `research_salary` → `SalaryQuery` struct refactor (mechanical + correct; a clean follow-up, out of scope here).

365 `@ajh/prompts` tests + new cache-key/Ollama-gating unit tests; typecheck + eslint + cargo fmt/clippy/build/architecture green.

saeedkolivand:

PR #568, opened 2026-07-08
fix: dedup indeed imports by job id and link generation docs via fk

saeedkolivand:
Post-audit issue #3 — importing an **Indeed** job (esp. with the Applications page open) went straight to `applied` instead of `saved`, and the Application detail page hid existing generated docs. Two root causes (not a status branch):

## (a) URL-normalization collision
`normalize_job_url` dropped the whole query string, but Indeed's job id lives in the query (`viewjob?jk=<id>`) — so **every** Indeed job collapsed to one dedup key and imports merged onto the first Application there (often a Generate-origin `applied` row). LinkedIn puts the id in the path, which is why this was Indeed-specific.

**Fix:** make `normalize_job_url` **additive** — retain a per-host allowlist of identifying query params (Indeed `jk`) while still dropping tracking params (`utm_*`, `ref`, param reorder). Path-based hosts normalize byte-for-byte as before. Distinct Indeed jobs now get distinct keys → correct `saved` dedup, and the stored URL is a valid openable link again.

## (b) Docs joined by raw URL
The detail page matched generations to the Application by raw `jobUrl` equality, but the Application stores the *normalized* URL and the generation stores the *raw* one → never matched for Indeed.

**Fix:** stamp `application_id` on the generation at **save** time (from the origin upsert that already runs first) and join the detail page on that **FK** instead of the URL. Boot-time backfill still covers legacy rows. Also invalidate the detail query on import so an open page refreshes.

## Verification
`gen:ipc:check`, `turbo run typecheck --force` (0 errors), `lint:strict`, cargo fmt/check/clippy, and the JS/shared suites (1969 + 147) pass. New Rust tests (jk retained, tracking dropped, LinkedIn unchanged) + updated detail-page FK-join tests.

### DIFF
--- a/apps/desktop/src-tauri/src/applications/mod.rs
+++ b/apps/desktop/src-tauri/src/applications/mod.rs
@@ -950,8 +950,10 @@ fn explicit_scheme(input: &str) -> Option<String> {
 }
 
 /// Normalize a job URL into a stable dedup key: lowercase host, strip a leading
-/// `www.`, drop the query (`?…`) and fragment (`#…`), and trim a trailing `/`.
-/// The scheme is preserved (lowercased). Empty input returns empty.
+/// `www.`, drop the fragment (`#…`), retain only per-host *identifying* query
+/// params (e.g. Indeed `jk`) while dropping every other query param (utm_*, ref,
+/// tracking), and trim a trailing `/`. The scheme is preserved (lowercased). Empty
+/// input returns empty.
 ///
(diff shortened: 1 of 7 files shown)


saeedkolivand:
@claude review

saeedkolivand:
@coderabbitai review

saeedkolivand:
🎉 This is included in version 0.124.0.

PR #569, opened 2026-07-08
fix: floor aggregator sub-day date filters to 3 days

saeedkolivand:
Post-audit issue #6 — an **Autopilot on the Aggregator board** with a "Last 24h" (or any sub-day) date filter returned **zero** results, while the same search on the Jobs page (which defaults to *no* date filter → a 30-day window) returned results.

## Root cause
Adzuna (`max_days_old`) and JSearch (`date_posted`) have only **whole-day** recency granularity. The aggregator mapped every sub-day UI token (`15m`…`24h`) to the tightest window — Adzuna `max_days_old=1` / JSearch `"today"` — which returns near-zero on a quiet day. (Country was *not* the cause — both paths resolve Germany to `de`.)

## Fix
Floor the sub-day tier to **3 days** (Adzuna `max_days_old=3`, JSearch `"3days"`) and rely on the query's existing freshest-first sort (`&sort_by=date&sort_direction=down`). Coarser tiers and the no-filter default (30 days) are unchanged.

The paid **LinkedIn (Apify)** path keeps its exact sub-day window (`f_TPR=r86400`) — LinkedIn has real seconds-level recency, so it's accurate there, not a clamp.

## Tradeoff
A "Last 24h" aggregator filter now surfaces up-to-3-day-old jobs, freshest first, instead of nothing — the aggregator has no true sub-day granularity, so a 3-day floor + date-sort is the closest honest approximation of "recent".

## Verification
cargo fmt/check/clippy and `gen:ipc:check` pass; mapping unit tests + the exhaustiveness guard updated.

### DIFF
--- a/apps/desktop/src-tauri/src/scraping/boards/aggregator/mod.rs
+++ b/apps/desktop/src-tauri/src/scraping/boards/aggregator/mod.rs
@@ -70,22 +70,28 @@ where
 
 /// Map a UI date-filter token to Adzuna's `max_days_old` integer (whole days).
 ///
-// ponytail: Adzuna's recency granularity is whole days, so all sub-day options
-// collapse to 1 day (API ceiling). No filter / unrecognized token caps at 30 days
-// so the aggregator never surfaces postings older than a month.
+// ponytail: Adzuna's recency granularity is whole days — it can't do sub-day. A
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
🎉 This is included in version 0.124.0.

PR #583, opened 2026-07-08
chore: batch safe dep + tooling bumps (pnpm, harden-runner, upload-sarif, setup-uv)

saeedkolivand:
## Batch: safe dependency + tooling bumps

Groups the verified-safe Dependabot bumps with the pnpm tooling bump into a single PR (one CI cycle instead of four, since the repo's strict up-to-date merge policy serializes individual Dependabot merges).

### Included
- **pnpm** `11.1.3 → 11.10.0` (`packageManager` in `package.json`)
- **step-security/harden-runner** `2.19.4 → 2.20.0` — supersedes #571
- **github/codeql-action/upload-sarif** `4.36.2 → 4.37.0` — supersedes #572 (standalone use in the `semgrep`/`scorecard` jobs; independent of `init`/`analyze`)
- **astral-sh/setup-uv** `8.2.0 → 8.3.2` — supersedes #574

All three action bumps are the exact Dependabot commits (SHA-pinned), applied as one squashed commit. Each was independently checked (diff + upstream release notes + in-repo usage) and found to touch only workflow YAML.

### Deliberately excluded (not safe as lone bumps)
- **#570** codeql-action/analyze `4.37.0` — would leave `init` at 4.36.2 in the same `codeql` job → deterministic version-mismatch failure. Needs a coordinated `init`+`analyze` bump.
- **#577** sha1 `0.11.0` — breaks the non-Windows build (RustCrypto digest 0.11 vs hmac/pbkdf2 0.12 on digest 0.10). Needs a coordinated RustCrypto-family bump.
- **#578** typst `0.15.0` — lone bump breaks the build; the typst-* crates must move in lockstep + code migration.
- **#579** cbc `0.2.1` — breaks the non-Windows build (cipher 0.5 vs aes 0.8 on cipher 0.4). Needs an aes 0.9 + code change.
- **#581** conventional-changelog-conventionalcommits `10.2.1` — reintroduces the empty-release-notes regression; closed, staying on v9.

PR #584, opened 2026-07-08
chore: ci + tooling bumps (codeql-action 4.37.0, pnpm single-source, turbo 2.10.4)

saeedkolivand:
## CI + tooling bumps

Three low-risk tooling/CI changes batched into one PR (one CI cycle).

### 1. codeql-action `init` + `analyze` → 4.37.0 — supersedes #570
`#570` bumped only `analyze`, which would leave `init` at 4.36.2 in the same `codeql` job → deterministic version-mismatch failure. This bumps `init` **and** `analyze` to the 4.37.0 SHA (`99df26d4…`), matching `upload-sarif` (already bumped in #583). The whole codeql-action family is now consistent at 4.37.0.

### 2. pnpm version — single source of truth
The pnpm version was pinned in **6 places** (`package.json` `packageManager` + 5 `pnpm/action-setup` `version:` inputs), and `pnpm/action-setup@v6` errors when they disagree — which is what broke CI when `packageManager` was bumped to 11.10.0 in #583.

Fix: drop the redundant `version:` inputs from the shared `setup-node-pnpm` composite action and the 4 `release.yml` steps, so `pnpm/action-setup` reads `packageManager` from `package.json` as the **single source**. Future pnpm bumps are now a one-line `package.json` edit. (Verified every `pnpm/action-setup` call runs after `actions/checkout`, so `package.json` is always present.)

### 3. turbo `2.10.3 → 2.10.4`
Patch bump (`package.json` + `@turbo/*` platform binaries in the lockfile). `turbo.json` already uses the v2 `tasks` schema — no migration.

### DIFF
--- a/.github/actions/setup-node-pnpm/action.yml
+++ b/.github/actions/setup-node-pnpm/action.yml
@@ -12,8 +12,6 @@ runs:
   steps:
     - name: 📦 Setup pnpm
       uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
-      with:
-        version: 11.10.0
 
     - name: 🟢 Setup Node.js
       uses: actions/setup-node@v6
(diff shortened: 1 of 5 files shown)


saeedkolivand:
@claude review

saeedkolivand:
🎉 This is included in version 0.124.1.

PR #594, opened 2026-07-10
feat: selectable letter layouts inheriting the resume template palette (templates pr 5/6)

saeedkolivand:
## Summary

PR 5 of the template-overhaul series (after #590–#593): **Letter layouts** — the first time cover letters have selectable designs.

- **Concept:** a *Letter layout* owns **arrangement only**; colors and fonts keep inheriting from the selected résumé template (the pair stays visually coordinated). Canonical term per the design glossary — deliberately not "letter template".
- **Three layouts:** `classic` (the existing letter — byte-identical default), `refined` (large name + role top-left, right-aligned contact, full-width rule, always-on JOB REFERENCE line from the subject, spaced signature area — "Olivia Wilson" reference), `banded` (angled pale accent band on page 1 only, small-caps name in the template's name font, right-aligned stacked contact, short accent-rule footer — "Belinda Davidson" reference).
- **Market conventions still own semantics:** DIN 5008 essentials (A4, date-top-right, Betreff, German salutation/sign-off) hold across all three layouts — layouts gate structural elements on the conventions data, never on the layout id. Verified by per-market render tests + out-of-band renders.
- **Wire:** `letterLayoutId` (`#[serde(default)]` → classic; graceful unknown-value fallback mirrors TemplateId). Old clients unaffected.
- **UI:** `LetterLayoutPicker` (radiogroup) surfaces in the wizard (cover + combined targets), the done-panel cover tab, and the tailor-flow cover tab — one shared per-export state, preview and export always agree.

## Verification highlights

- All 3 layouts render valid PDFs in US + DE with correct reading order (out-of-band render since the host can't execute cargo tests; CI authoritative).
- Banded's band tint proven **precisely attributable and page-1-only** (SVG fill diff vs classic, absent on page 2), and its geometry set from **measured** header extents (38.86mm) rather than estimates — 44mm band with verified clearance at A4 and US Letter.
- Palette inheritance proven: Regent's burgundy reaches the letter output; Regent ≠ Swiss bytes for both new layouts.
- Classic default-path regression: identical source string compiled; default chain TS-undefined → omitted → serde default tested.

## Review status (pre-PR agent gates)

`resume-export-expert` — no blocking findings; MEDIUM (picker unreachable in the primary résumé+cover flow) fixed in-diff with tests; LOW polish (dead bindings, band geometry, comment accuracy, US label dedup) all applied. Conventions verdict: refined's always-on reference line is convention-safe (US emits no subject unless the content carries one).

[...]

PR #596, opened 2026-07-10
fix: classic cover letters render the recipient address block again

saeedkolivand:
## Summary

Fixes a day-one bug (shipped 2026-06-02): the **Classic cover-letter layout never rendered the recipient/inside-address block** — for any of the 16 markets. The template gated emission on order tokens (`after-date`/`""`) while the shared conventions fixture supplies placement tokens (`left`/`top-right`), passed verbatim — the vocabularies never overlapped. Recipient lines were parsed, serialized, and silently dropped. DE/AT/CH letters violated DIN 5008 (the Anschriftfeld is mandatory); every market lost the inside address. The refined/banded layouts (#594) were unaffected (unconditional emission) — Classic is the default, so most users were.

**Fix (2 lines of production code):** the template now emits the block for anything except `before-date` (that branch is preserved untouched); the misleading Rust doc-comment claiming a `before-date | after-date` contract is corrected. The fixture stays untouched — it's the shared TS↔Rust source of truth and its placement vocabulary serves the prompt layer. FR's true top-right recipient layout remains unimplemented (renders left-aligned — strict improvement over rendering nothing; possible follow-up).

## Why tests never caught it

Parser tests asserted `recipient_lines` **in the model**; render tests asserted salutation/subject/body/sign-off — nothing asserted recipient content in the extracted text. Now: position-aware regression assertions in the US + DE Classic render tests (recipient-unique tokens, ordered before the salutation/Betreff to prove the inside-address occurrence), and the refined/banded recipient rendering is pinned.

## Verification

Real renders through `render_letter_pdf` + pdf_extract (example-binary; host cannot execute test binaries — CI authoritative): US `street_line=true, acme_before_salutation=true`; DE `musterfirma=true, anna_weber=true, recipient_before_betreff=true`; both single-page, salutations intact. cargo check/clippy/fmt clean.

### DIFF
--- a/apps/desktop/src-tauri/src/export/typst_engine/letter.rs
+++ b/apps/desktop/src-tauri/src/export/typst_engine/letter.rs
@@ -83,11 +83,11 @@ pub(super) struct LetterOpts {
     /// `"top"` | `"bottom"`.
     pub sender_position: String,
     /// Horizontal-placement hint from the shared fixture (`"left"` | `"top-right"`),
-    /// passed through verbatim — NOT an order token. The Classic template treats
-    /// anything except an explicit `"before-date"` as after-date placement, so the
-    /// inside address renders for every market. (fr's true top-right recipient
-    /// layout is not implemented — it renders left-aligned like the rest; a real
-    /// top-right variant is a possible follow-up.)
+    /// passed through verbatim. Not currently read by any `.typ` template — every
+    /// letter layout (Classic/Refined/Banded) renders the recipient block
+    /// unconditionally, left-aligned, right after the date. (fr's true top-right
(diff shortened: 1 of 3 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Both LOWs addressed by deletion per your suggestion: the unreachable before-date branch, the dead no-op else-if, and the now-unused recip-pos binding are removed — emit-recipient-block() runs unconditionally at a single point (matching refined/banded), which also structurally eliminates the latent top-right+before-date fall-through. The recipient_position field stays in the data contract (documented as unread, reserved for a future FR top-right variant). Re-render-verified US + DE after the deletion.

saeedkolivand:

PR #602, opened 2026-07-11
feat: canonical location model with honest per-board location handling

saeedkolivand:
## Summary

PR F of the job-search trust program (audit 2026-07-10) — the one refactor the audit sanctioned: location was a free-text string that 23 boards freelanced over, with 17+ silently ignoring it. Now the app knows which boards honor location, approximates honestly for the rest, and says so. Builds on #597-#601.

## What changed

- **`LocationSpec`** (city/region/country/lat/lon/radius) derives once via an accessor from the geo fields the renderer already sends — zero churn on 54 construction sites, automatic back-compat (no location → everything byte-identical).
- **`Scraper::supports_location()`** — author-verified catalog: only the aggregator (`where` + market routing), LinkedIn (geoId/radius), and Arbeitsagentur (`wo`) consume location server-side; the other 20 boards are honestly flagged `false`, exposed to the renderer as `supportsLocation` on the board catalog.
- **Central conservative filter** (`scraping/engine/location_filter.rs`): for non-supporting boards, drops a posting only on a clear place mismatch — diacritic folding (base + DIN-5007-2) so Muenchen/München/Munchen always match, a curated exonym table (Munich/München, Cologne/Köln, Nuremberg/Nürnberg) with the boundary documented and tested, remote/unknown-location rows always kept. Filtered items are excluded from the item cap (`KeepItemFn` runs before the counter), so boards keep paginating until enough *matching* rows arrive — no silent under-delivery.
- **Nothing reads clean by accident:** every non-supporting board notes `location-filtered:<n>` — including n=0 — whenever a location was requested, rendered as chips ("N off-location results hidden" / "location filtered locally"); the board picker warns up front (`LocationFilterNote` in both the scrape form and the autopilot wizard). A run with location + non-supporting boards deliberately no longer collapses to "all ok".

## Review trail (internal fleet)

- scraping-applier-expert: REQUEST-CHANGES → 2 HIGH fixed (cap/filter ordering — boundary-pinned test; diacritic false-drops — folding + exonym table) → APPROVE
- frontend-reviewer: APPROVE-WITH-FIXES → contrast floor, defaultValue fallback, and the trust-story gap (unconditional n=0 notes) closed
- tauri-security-reviewer: PASS (linear matching — no ReDoS; count-only notes — no PII; zero new IPC/dep surface)
- pr-reviewer ×2: opus PASS (verified all 23 boards stream==return, so the kept-set path can't zero out results); sonnet PASS 1🟡 → integration render tests added

## Test plan

[...]

PR #597, opened 2026-07-10
fix: represent board fetch failures instead of silent zero results

saeedkolivand:
## Summary

PR A of the job-search trust program (audit 2026-07-10, `.claude/scratch/job-search-trust-audit.md`): the scrape fleet's #1 trust killer was `fetch_json` collapsing every non-2xx response and schema-drift parse failure into `Ok(None)`, which nearly every board rendered as a successful zero-result run — a blocked, rotted, or mistyped-slug board was indistinguishable from "no jobs".

## What changed

- `scraping/http/mod.rs` — `fetch_json` returns `AppResult<T>` (was `AppResult<Option<T>>`): non-2xx → `Err(Provider("HTTP <status>"))`, schema drift → `Err(Parse(static message))` with serde detail kept log-side only. The `Option` removal compile-forces every call site to handle failure honestly.
- All ~20 `fetch_json` board call sites propagate errors into `BoardScrapeSummary.error` instead of mapping to empty success; `germantechjobs`/`wwr`/`berlinstartupjobs` (`fetch_text`) err on non-200 (fixes the documented germantechjobs `/rss` 403 silent zero).
- ATS boards: shared `ats_all_fetches_failed` gate in `scraping/boards/common.rs` wired into all 10 ATS boards — an all-companies-failed run now returns `Err` (closes the wrong-company-slug silent zero); lever/recruitee/smartrecruiters gained the `successful_fetches`/`first_fetch_error` tracking they lacked.
- Paginated boards (themuse/arbeitnow/arbeitsagentur): shared `should_propagate_page_error` — page-0 failure propagates, later-page failure keeps the partial harvest; user-cancel breaks cleanly instead of surfacing as a board error.
- Adzuna/JSearch raw HTTP failures carry their provider prefix so aggregator errors stay attributable.
- Tests: `fetch_json` contract tests (403 carries status, Parse message pinned static + no-body-leak guard), truth-table tests for both shared fns, engine containment test (failing board's error lands in its summary, siblings' jobs survive, genuine-empty stays distinct).
- Docs: `docs/knowledge/scraping-domain.md` synced (error contract + stale aggregator file refs).

## Review trail (internal fleet)

- scraping-applier-expert: APPROVE-WITH-FIXES → provider-prefix fix applied
- testing-reviewer: REQUEST-CHANGES → pure-fn extraction + truth-table tests → APPROVE
[...]

PR #603, opened 2026-07-11
fix: live-verified board hygiene with honest failure reasons

saeedkolivand:
## Summary

PR G of the job-search trust program (audit 2026-07-10): the never-verified recon boards got a live-shape verification pass, the one board that could only ever silently return zero is no longer promised to users, and LinkedIn's silent soft-block is detected. Builds on #597-#602.

## What changed

- **Live verification (curl, one-shot, 2026-07-11):** themuse (runs on every keyword search), bamboohr, and rippling shapes confirmed field-by-field with dated module notes. **Breezy was functionally dead** — `location.state` drifted from string to object, atomically failing every tenant's deserialize — fixed with a tolerant untagged parse + per-row handling, PLUS a total-drift guard: a response whose rows ALL fail to parse records a fetch failure, never a silent empty success. Pinpoint has no findable public slug — documented unverifiable; PR A's error plumbing covers every failure mode.
- **Comeet hidden** (`listed() → false`): its response shape was speculative and the live endpoint 400s — the code stays registered (saved targets still dispatch), only the user-facing promise is removed.
- **LinkedIn soft-block detection:** empirically established that page 0 always pads with talent-pool cards (even nonsense keywords) and paging past the end returns 400 — so a 200 page-0 with zero cards is never a genuine empty. It now errs honestly ("may be rate-limiting / require login / changed layout"), after the geoId-strip retry, cancellation-guarded. Plus country-biased geoId selection (alias table incl. Adzuna-allowlist parity) and an in-process geoId cache.
- **All-slugs-invalid becomes a distinct error** ("all N company slug(s) invalid — check the company names in the jobs search form") across the 7 slug-validating boards, via a shared `ats_finish_search` that encodes cancellation-wins in one tested place — a mid-list cancel can never be misreported as user misconfiguration.

## Review trail (internal fleet)

- scraping-applier-expert: REQUEST-CHANGES → cancellation-guard HIGH fixed across all 7 boards + country aliases → resolved
- tauri-security-reviewer: PASS-WITH-ADVISORIES (no slug values in error strings; geoId cache in-memory only; serde untagged bounded by response caps)
- pr-reviewer ×2 (opus + sonnet independently converged): total-row-drift silent zero → fetch-failure guard added; wrong-screen copy → reworded; cancel-guard consolidated into ats_finish_search with shared tests

## Test plan

- CI runs the Rust suite (host fault blocks local cargo test; fmt/check/clippy clean). Live shapes captured as fixtures; standalone example verified the breezy untagged parse + geoId bias.
- Manual: scrape with a breezy company (e.g. a real tenant) → jobs return; enter an invalid company name for an ATS board → distinct "slugs invalid" chip, not silent zero.

Part 7/8 of the trust program — next: PR H (scheduler retry + robustness tail).

### DIFF
--- a/apps/desktop/src-tauri/src/scraping/boards/bamboohr/mod.rs
+++ b/apps/desktop/src-tauri/src/scraping/boards/bamboohr/mod.rs
@@ -5,9 +5,12 @@
 //! board with `"needs-company"` when `input.companies` is empty.
 //!
 //! Endpoint reconnaissance ported from santifer/career-ops (MIT), `providers/bamboohr.mjs`.
+//! Live-verified 2026-07-11 against slug `fivetran` (4 jobs): response is
+//! `{meta, result[{id (string), jobOpeningName, location{city, state (string)},
+//! isRemote}]}` — shape confirmed, no drift.
 use super::super::http::fetch_json;
 use super::super::types::{BoardSearchInput, JobPosting, ScrapeContext, Scraper, ScraperMode};
-use super::common::{ats_all_fetches_failed, is_valid_dns_label_slug, normalize_companies};
+use super::common::{ats_board_failure, is_valid_dns_label_slug, normalize_companies};
 use async_trait::async_trait;
 use serde::Deserialize;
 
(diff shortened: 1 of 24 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #607, opened 2026-07-11
chore: ast-grep tier-0 rule pack and gitleaks secret scan

saeedkolivand:
PR 3 of 6 — review-system hardening program. Deterministic invariants move out of prose/regex into versioned, testable structural rules.

## What

- **NEW \`.claude/review-rules/*.yml\`** — five ast-grep rules, every one validated against the real repo before landing:
  - \`rust-env-var-outside-platform\` (error, also matches \`var_os\`) — 0 repo hits
  - \`rust-reqwest-construct-outside-net\` (error) — **construction-only** (\`Client::new/builder\`, \`ClientBuilder::new\`), aligned to \`tests/architecture.rs\` R5. The naive "type mention" form had 2 false positives in \`scraping/\` — which also exposed that the gate's legacy regex had the same latent false-block; fixed here too.
  - \`rust-result-string-outside-error\` (error) — 0 repo hits
  - \`ts-gctime-infinity\` + \`tsx-gctime-infinity\` (warning; one rule per language — a tsx rule never matches .ts files) — ignore test files (3 test QueryClients legitimately pin gcTime)
- **Deliberately dropped** two planned rules after the repo scan: unwrap-in-LazyLock-init (72 pre-existing hits — known-good literal regexes in statics are house style; "fallible/externally-influenced" is reviewer judgment) and \`json!(ident)\` in commands (40 hits — struct-vs-contract-shaped-local isn't mechanically decidable). Both stay prose-only in the review checklists. Zero-false-positive is the acceptance bar for blocking rules.
- **Stop gate Tier 0**: primary = \`pnpm exec ast-grep scan --json=compact\` over the changed files (repo-root \`sgconfig.yml\` auto-discovery — \`-c\` from elsewhere silently breaks \`files:\` globs; forward-slash paths required). error→HIGH conf 1.0, warning→MEDIUM advisory; PR 1's added-lines gating still applies (pre-existing hits in touched files stay non-blocking). Introduced HIGH hits short-circuit to a **zero-LLM deterministic block**. Legacy regexes remain only as fallback when the binary is missing (logged as \`sg_fallback\`).
- **pr-reviewer secret scan**: prefers \`gitleaks\` (\`detect --log-opts\` + \`protect --staged\`) when on PATH, grep fallback with an install hint otherwise (no downloader script — gitleaks has no npm package; CI enforcement comes in PR 6).
- Tooling: \`@ast-grep/cli\` devDependency (+ build approval in \`pnpm-workspace.yaml\`), \`scan:rules\` script, docs note in DEVELOPMENT.md.

## Verification

- \`pnpm scan:rules\` clean repo-wide (the FP audit).
- All five rules proven to FIRE on positive fixtures (a never-firing rule scans clean forever — worse than no rule).
- Live canary through the real gate: untracked \`std::env::var\` file → \`tier0-block\` via the sg rule in 1.4s, \`model: "none"\`, no \`sg_fallback\` in metrics — zero LLM spend.
- Gate harness 21/21 (test repos exercise the regex-fallback path); \`check:agent-system\` green.

### DIFF
--- a/.claude/agents/pr-reviewer.md
+++ b/.claude/agents/pr-reviewer.md
@@ -34,7 +34,7 @@ pasted their outputs into your prompt, use those — do NOT re-run them.** Other
 - **Cross-OS / `#[cfg]` blind spot:** full doctrine in `.claude/review-checklists/rust.md` — a same-host cargo run silently excludes other-OS `#[cfg]` code; treat touched OS-gated code as **UNVERIFIED** (require a cross-target `cargo check --target …` or hand-review the gated bodies). Never report 🟢/PASS on cfg-gated code a same-host build can't compile.
 - If `packages/shared/**` IPC contracts/schemas changed: `rtk pnpm gen:ipc:check` (must be clean) and confirm `mock-client.ts` mirrors any new method.
 - **Targeted tests**: run the test files covering the touched code (not the full suite) — `rtk pnpm --filter @ajh/desktop test <paths>` / the owning package filter.
-- **Secret scan**: grep the diff for hardcoded secrets/keys/tokens (API keys, `adzuna` app_id/app_key literals, private keys, `Authorization:` bearer literals). Any committed secret is 🔴.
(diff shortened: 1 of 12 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #609, opened 2026-07-11
chore: pre-push ai review gate with cargo fault mitigation

saeedkolivand:
PR 5 of 6 — review-system hardening program. The user-chosen enforcement point: nothing reaches GitHub unreviewed — and the reason `--no-verify` became a daily habit is fixed in the same PR (a gate behind a bypassed hook is theater).

## What

- **NEW \`scripts/pre-push-review.mjs\`** — AI review of the exact push range (\`PREPUSH_RANGES\` accumulated in the hook's stdin ref loop; stdin can't be re-read). Layers, cheapest first:
  1. **cache fast-path** — hunks the Stop gate already reviewed clean pass in <1s (in the normal workflow this makes pre-push near-free)
  2. **ledger lane** — unresolved open HIGH/CRITICAL findings from previous reviews surface without any LLM call
  3. **ast-grep deterministic scan** — error-severity findings **exit 1 from day one**, even in warn mode
  4. **one sonnet schema-1 review** — **RATCHET (user decision): \`REVIEW_MODE=warn\` default** — prints loudly, exits 0, logs \`would_block\`; flip to \`block\` after \`/review-stats\` shows a clean FP rate (~2 weeks). Evidence: day-one full blocking → ~40% dev bypass within a week.
  - Fail-open on infra (binary missing / 180s timeout / double parse-fail) — CI is the backstop. \`REVIEW_SKIP=1\` skips but logs an **audited** metrics entry.
- **\`.husky/pre-push\`**: range accumulation; the AI review as the final step; and the **cargo-test entrypoint-fault fix** — \`STATUS_ENTRYPOINT_NOT_FOUND\`/\`0xc0000139\` is auto-detected (or \`AJH_SKIP_CARGO_TEST=1\`) and skips ONLY cargo test with a loud warning while check/clippy/fmt/deny still run. Any other test failure still aborts the push.
- \`docs/DEVELOPMENT.md\`: the three layers, both escape hatches, and why \`--no-verify\` is now the wrong tool (it hides the audit trail).

## Verification (all against the real script/hook, not mocks)

- Canary commit with \`std::env::var\` outside \`platform/\` → script exits 1 with the ast-grep finding, \`tier0-block\` metric
- \`--log-skip\` → audited \`skipped:true\` metric; empty range → warned fail-open; \`sh -n\` clean
- Nested \`claude -p\` verified working from hook context (earlier \`llm-unavailable\` was the 120s timeout on a large diff — pre-push now gets 180s, no Stop-hook constraint applies)
- **Acceptance test: this PR's own push ran through the FULL hook with no \`--no-verify\`** — pre-push AI review made a live sonnet pass and logged \`clean\`
- \`check:agent-system\` green

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -68,6 +68,7 @@ export TURBO_FORCE=1
 # diff against the merge-base with origin/main. Any failure → FULL gate.
 ZERO=0000000000000000000000000000000000000000
 CHANGED=""
+RANGES=""
 HAVE_RANGE=0
 while read -r local_ref local_sha remote_ref remote_sha; do
   if [ -z "${local_sha:-}" ] || [ "$local_sha" = "$ZERO" ]; then
@@ -85,6 +86,7 @@ while read -r local_ref local_sha remote_ref remote_sha; do
     range="$remote_sha..$local_sha"
   fi
   files=$(git diff --name-only "$range" 2>/dev/null || true)
+  RANGES="$RANGES $range"
   CHANGED="$CHANGED
 $files"
   HAVE_RANGE=1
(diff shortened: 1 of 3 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Review findings addressed: warn mode now continues past stale open blockers into layers 2-3 (live-tested: seeded ledger blocker → block mode exit 1; warn mode re-emits AND completes a fresh sonnet review of the range). Both doc errors fixed (REVIEW_MODE is an env var; layers are sequential/short-circuiting). Skipped with rationale: cargo-test output buffering (the capture is what makes fault detection reliable; trade-off accepted), shell-quoting symmetry (matches the gate; both can move to safer arg passing later), ADR-0008 forward reference (the ADR lands in PR 6, the program's closing PR — intentional). Your working-tree divergence note was a runner-side artifact again — local tree is clean.

saeedkolivand:

PR #613, opened 2026-07-11
chore: raise pre-push llm review timeout to 300s

saeedkolivand:
One-line calibration from the first /review-stats-class signal: the pre-push LLM layer had a 3-for-3 `llm-unavailable` streak. A timed diagnostic showed a small 2-file sonnet review takes ~170s in this environment (nested `claude -p` startup + model latency), so the 180s budget failed open on essentially every real diff — the layer was silently theater, exactly the failure mode the metrics log exists to catch. 300s gives the measured latency real headroom; in warn-mode the cost is latency only, never a block. The deterministic ast-grep layer and cache fast-path are unaffected (they carry the sub-second paths).

### DIFF
--- a/scripts/pre-push-review.mjs
+++ b/scripts/pre-push-review.mjs
@@ -201,8 +201,11 @@ ${learnings ? `\n## Known repo false positives — do NOT re-raise any of these\
 ${diff}
 \`\`\`
 `;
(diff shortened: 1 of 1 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #618, opened 2026-07-11
feat: add jooble as a byo-key aggregator fallback provider

saeedkolivand:
## What

First of the P2 sourcing-depth PRs. Adds **Jooble** (jooble.org/api) as a new bring-your-own-key aggregator provider — **~67-country coverage vs Adzuna's ~19 markets**, directly patching the unsupported-country zero-jobs case.

- Wired as a **third fallback tier**: fires only when Adzuna AND JSearch come up unconfigured/empty/err, so it never hammers Jooble's undocumented quota on searches the primary chain already answers. JSearch `Ok(empty)` short-circuits before Jooble — symmetric with Adzuna's existing empty-means-stop rule (**both edges regression-tested**).
- New `JoobleProvider` modeled on JSearch; key from the `ai:jooble-key` keyring slot, degrades a read error to `None` (optional-key policy).
- Routes through `fetch_json` — inherits the trust-program's failure-representable fetch, 429/503 backoff, per-host rate limiting, and cancellation. Errors prefixed `jooble:`; `external_id` prefixed `jooble-` for dedup.
- **Security**: Jooble puts the API key in the URL *path*, so the error-log line would have leaked it on a bad-key 403 → added opt-in `FetchOptions.redact_path` + `safe_log_url()` that strips the path (default `false` = byte-identical for the 20+ existing callers). Key percent-encoded; query/location go in the POST body.
- Provider-slot contract + a settings `AggregatorKeyField` (en+de) with a "get a free key" link.

## Review

**scraping-applier-expert + tauri-security-reviewer: no HIGH/CRITICAL.** Security verified the credential-in-log fix airtight on every response path; domain confirmed the fallback wiring and trust-program non-regression.

**Note (needs a live key to confirm — deferred):** Jooble's real rate-limit/quota response shape is unverified. The failure-representable path handles a real 403/429 regardless; the one consistent-with-Adzuna/JSearch risk is if Jooble signals quota as `200 OK` empty rather than non-2xx (would short-circuit silently, same as the other providers' genuine-empty case). Worth a spot-check when you add a key.

## Validation

- 18 Jooble unit tests + full aggregator module (106) + http module — green; clippy clean; whole-graph typecheck + `@ajh/desktop` tests pass; `gen:ipc:check` clean.

### DIFF
--- a/apps/desktop/src-tauri/src/ipc_contracts/provider_slots.rs
+++ b/apps/desktop/src-tauri/src/ipc_contracts/provider_slots.rs
@@ -5,6 +5,7 @@
 pub const ADZUNA_APP_ID: &str = "adzuna-app-id";
 pub const ADZUNA_APP_KEY: &str = "adzuna-app-key";
 pub const JSEARCH_KEY: &str = "jsearch-key";
+pub const JOOBLE_KEY: &str = "jooble-key";
 pub const APIFY_TOKEN: &str = "apify-token";
 pub const COMEET_COMPANY_UID: &str = "comeet-company-uid";
 pub const COMEET_API_TOKEN: &str = "comeet-api-token";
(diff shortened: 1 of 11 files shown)


saeedkolivand:
@claude review

saeedkolivand:
@claude review

saeedkolivand:

PR #619, opened 2026-07-11
fix: parse jooble timezone-less timestamps instead of dropping posted_at

saeedkolivand:
## What

Fast-follow to #618. Verified against the **live Jooble API** (with a real key): the `updated` field arrives as `2026-05-15T00:00:00.0000000` — **no timezone offset** — which `chrono::DateTime::parse_from_rfc3339` rejects. So `posted_at` was silently `None` for every real Jooble job (no dates, broken date-filtering). The unit-test fixtures used a well-formed RFC3339 timestamp and never caught it.

- `map_jooble_job` now tries RFC3339 first (in case any entry carries an offset), then falls back to a naive datetime assumed UTC, tolerant of Jooble's 7-digit-fraction/no-TZ form.
- Adzuna/JSearch parsing untouched.

## Tests

3 new cases beside the existing `map_jooble_job` mapping tests: the real no-TZ shape (`posted_at` = the correct UTC millis, computed via chrono not a magic number), a TZ-bearing RFC3339 shape (still parses via the first branch), and a malformed string (`None`, no panic). Full aggregator module: 113/113; clippy + `fmt --check` clean.

## Note

#618 merged before this fix landed, so it ships as a fast-follow. This is a live-key finding — the kind fixtures can't surface — so it's worth a spot-check that dates now populate when you next run a Jooble-fallback search.

### DIFF
--- a/apps/desktop/src-tauri/src/scraping/boards/aggregator/providers.rs
+++ b/apps/desktop/src-tauri/src/scraping/boards/aggregator/providers.rs
@@ -693,11 +693,22 @@ pub(super) fn map_jooble_job(j: JoobleJob, now: i64) -> Option<JobPosting> {
         source: "aggregator".to_string(),
         description: j.snippet.map(|d| html_to_markdown(&d)),
         requirements: None,
-        posted_at: j
-            .updated
-            .as_deref()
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #621, opened 2026-07-11
feat: route the ats seed into company-scoped boards (engine-side) — DO NOT MERGE until disclosure decision

saeedkolivand:
## ⚠️ Owner decision required before merge (not a bug — a product call)

This finishes P2 sourcing depth by wiring the merged `ats_seed` directory into discovery. It works and is correctness-clean, but it **changes unattended autopilot behavior**, so I'm leaving it open for you.

**The decision:** today, selecting an ATS board (Greenhouse/Lever/Ashby/…) in an autopilot workflow silently does nothing (skipped as `needs-company` forever). After this PR, that same selection starts querying the **curated seed companies** (Stripe, Airbnb, OpenAI, Bosch, N26, …) on every scheduled run — and the wizard has **no UI telling the user which companies are queried or letting them opt out** short of deselecting the board. You approved "autopilot gets ATS coverage," so the intent is right; the open question is the UX:
- **(a) merge as-is** — ATS boards just start working in autopilot with the curated list; or
- **(b) add a wizard disclosure first** ("this board searches curated companies: …") as a fast-follow before merge; or
- **(c) exclude ATS boards from the autopilot picker** and keep the seed for manual search only.

I'd lean (a) or (b). Your call.

## What it does

Engine now auto-populates a company-scoped ATS board's `companies` from `ats_seed::by_ats(scraper.id())` when the board is selected, requires a company, and the run supplied none. Seeds 7 of the 11 company-scoped boards.

- **Option (b) routing, engine-local**: no `BoardSearchInput` contract change, no scraper changes, no renderer changes. `run_boards` gains one borrowed `&HashMap` override param and stays seed-agnostic.
- Keyed on `scraper.id()`; **explicit user companies always win** (seed applies only when the list is empty); **unseeded requires-company boards still skip** (no regression).
- **Bounded fetch volume**: Σ(selected boards' slugs), capped by `MAX_BOARDS_PER_BATCH=6` — not 59×N. Rotted slugs 404 harmlessly → `CompletedWithErrors`, never silent-empty.
- Applies to **both manual search and autopilot** (unconditional — the recommended default; a strict improvement for manual, still gated by selecting an ATS board + leaving companies blank). Say if you want it autopilot-only.

## Footprint note (reviewer, informational)

**SmartRecruiters** does 1 list + up to 100 detail fetches *per company*; with its 4 seeded companies a run can fire ~400 requests to one host — correctly throttled by the 30 req/60s limiter, but that's ~13 throttled minutes for that board's slice of an unattended run. Not a defect; worth knowing.

## Validation

scraping-applier-expert review: **no CRITICAL, no correctness blockers** (skip gate, explicit-input precedence, trust-program non-regression, fetch bound, per-board seed consumption all verified). 782 scraping tests + clippy + `fmt --check` green. 3 new engine routing tests + per-board validator/casing tests.

### DIFF
--- a/apps/desktop/src-tauri/src/autopilot_helpers/mod.rs
+++ b/apps/desktop/src-tauri/src/autopilot_helpers/mod.rs
@@ -52,8 +52,12 @@ pub async fn autopilot_scrape(
         latitude: None,
         longitude: None,
         radius_km: None,
-        // Autopilot has no per-company target; ATS company slugs are a manual
-        // search affordance, so this stays empty (a no-op for every board).
+        // Autopilot has no per-company target, so it passes no explicit
+        // companies here. Company-scoped ATS boards (greenhouse, lever, ashby,
+        // smartrecruiters, recruitee, personio, workable) don't no-op on that —
+        // the engine (`scraping/engine/mod.rs`) falls back to the curated
+        // `ats_seed` directory for them when the list is empty. Non-company
+        // boards are unaffected.
         companies: Vec::new(),
     };
 
--- a/apps/desktop/src-tauri/src/scraping/boards/ashby/test.rs
+++ b/apps/desktop/src-tauri/src/scraping/boards/ashby/test.rs
@@ -89,6 +89,23 @@ fn normalize_cap_exact_boundary() {
     assert_eq!(result.len(), 50);
 }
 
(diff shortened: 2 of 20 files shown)


saeedkolivand:

PR #626, opened 2026-07-13
test: add en/de translation key and placeholder parity guard

saeedkolivand:
## What

Adds a real en/de translation **parity guard** and fixes 5 per-feature i18n tests that were silently failing to catch missing German keys.

## Why

The renderer inits i18n with `fallbackLng: 'en'` (`packages/translations/src/index.ts:56`), so a missing German key resolves to the **English** value, not the raw key. The existing per-feature `*.i18n.test.ts` guards tried to catch gaps with `expect(getFixedT(lng)(key)).not.toBe(key)` — which **passes on the fallback English text**, so an untranslated key ships silently. For a DACH-focused app that's a real blind spot.

## The guard

New `apps/desktop/src/renderer/i18n/translations-parity.test.ts` reads the raw resource trees via `getResourceBundle('en'|'de','translation')` (bypasses `fallbackLng`):

- **Key-set parity both directions** — flattens each locale to dot-path leaf keys (recursing objects **and arrays**; plural-suffixed `_one`/`_other` keys compare as-is) and asserts every en key exists in de and vice versa, **listing offenders** on failure.
- **Placeholder parity** — per shared key, asserts the `{{name}}` interpolation-variable *sets* match (catches a translator dropping `{{count}}`).
- **Exhaustive leaf classification** — `number`/`boolean`/`null` leaves are captured via `String(value)`, so a cross-locale *type* mismatch (`"count": 5` vs `"count": "5"`) can't escape the diff.
- **Self-checks** — synthetic missing-key / placeholder-mismatch / numeric-leaf inputs prove the guard actually bites (not a re-implementation — exercises the same exported `flatten`/`keysOnlyIn`/`extractPlaceholders`).

The 5 per-feature tests (`InterviewPracticePanel`, `BoardSummaryChips`, `LocationFilterNote`, `SeededCompaniesNote`, `AutopilotCard`) get **honest docstrings** (they verify resolution + interpolation, not gap-detection) and real `i18n.exists(key,{fallbackLng:false})` checks so they independently catch a gap in their own scope too.

## Result

Current en/de resources **pass clean** (2600+ keys, all placeholders match) — the blind spot existed but hadn't let a gap through yet; now one can't.

## Review notes (pre-PR)

`testing-reviewer` audited the harness and verified the guard genuinely bites (structural, array-length, and placeholder mismatches all caught; `fallbackLng:false` confirmed honored by i18next `exists()`). It caught a CI-blocking `tsc` TS2537 (invalid recursive indexed-access type) that the `rtk`/turbo typecheck wrapper had masked — fixed and re-verified with `tsc --noEmit` directly (exit 0).

## Tests

`@ajh/desktop`: **2311 pass** (+9). `tsc --noEmit` clean (verified directly). eslint clean (no disables — rule 15).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved translation coverage checks to detect missing English or German strings instead of silently relying on fallback translations.
  * Added validation that translated messages preserve required interpolation placeholders.
  * Strengthened checks for pluralized labels, hints, disclosures, and interview-practice content across supported locales.

### DIFF
--- a/apps/desktop/src/renderer/components/interview/InterviewPracticePanel.i18n.test.ts
+++ b/apps/desktop/src/renderer/components/interview/InterviewPracticePanel.i18n.test.ts
@@ -8,7 +8,7 @@
  * missing in de would still resolve here (to the English string), so this
  * file does NOT catch locale gaps — that's owned by the global
  * `i18n/translations-parity.test.ts`, which reads the raw resource trees
- * directly per-locale via `i18n.exists(key, { lng, fallbackLng: false })`.
+ * directly per-locale via `i18n.getResourceBundle` + `flatten`.
  * Mirrors AutopilotCard.i18n.test.ts.
  */
 
--- a/apps/desktop/src/renderer/components/interview/InterviewPracticePanel.i18n.test.ts
+++ b/apps/desktop/src/renderer/components/interview/InterviewPracticePanel.i18n.test.ts
@@ -1,8 +1,15 @@
 /**
- * en/de parity for the interview-practice keys (mock Q&A + STAR feedback).
+ * Resolution check for the interview-practice keys (mock Q&A + STAR feedback).
  * Uses the REAL @ajh/translations instance (not the identity mock the
- * component test uses) so a key present in only one locale fails here instead
- * of shipping the raw key string as UI copy. Mirrors AutopilotCard.i18n.test.ts.
+ * component test uses) so this verifies each key resolves to real, non-empty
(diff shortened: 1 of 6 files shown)


saeedkolivand:
@claude review

saeedkolivand:
@coderabbitai review

saeedkolivand:
@coderabbitai review

saeedkolivand:

PR #640, opened 2026-07-14
feat: source the settings changelog from the bundled changelog file

saeedkolivand:
## Summary

User-requested: the Settings changelog now reads from the repo's own `CHANGELOG.md`, bundled at compile time, instead of the GitHub releases API.

- `updater_changelog` no longer fetches `api.github.com/.../releases` — the semantic-release-generated `CHANGELOG.md` is embedded via `include_str!` and parsed by a small no-regex heading/body parser into the byte-identical `{releases:[...]}` IPC shape (capped at the newest 15). **Renderer diff: zero** — the contract, client, hook, and update-section component are untouched (the renderer never read the one field GitHub provided that the file doesn't, the release title).
- Works fully offline, no rate limits, and removes a second, ADR-0005-undocumented GitHub egress path (the on-launch updater version check remains the only GitHub call).
- **No version lag**: verified empirically that semantic-release commits `CHANGELOG.md` before tagging (`git show v0.126.0:CHANGELOG.md` leads with 0.126.0) and the build job checks out by tag (`release.yml:215-219`), so the shipped binary's bundled changelog always includes its own version.
- Reviewed by rust-backend-architect: no blocking findings; parser verified against all 38 real headings (byte-offset slicing is multibyte-safe, CRLF tolerated, preamble discarded); `reqwest` correctly retained (25 other consumers); `include_str!` path is compile-time source-relative (cwd-independent; this crate is never `cargo publish`ed).

## Test plan

- `cargo test --lib updater::` 21 passed (heading variants, body isolation, malformed/empty tolerance, plus two tests against the real bundled file: non-empty, newest-first via numeric semver compare); full crate 2270; architecture 11/11; exact CI clippy + fmt clean.
- Renderer suite 2312 passed; whole-graph typecheck + lint:strict clean; full pre-push gate suite passed on push.

### DIFF
--- a/apps/desktop/src-tauri/src/updater/mod.rs
+++ b/apps/desktop/src-tauri/src/updater/mod.rs
@@ -21,9 +21,7 @@ use parking_lot::Mutex;
 /// The Update object is stored across commands so the download URL and signature
 /// are never re-fetched, avoiding race conditions and unnecessary network calls.
 use std::sync::Arc;
-use std::time::Duration;
 
-use serde::Deserialize;
 use serde_json::{json, Value};
 use tauri::{AppHandle, Manager};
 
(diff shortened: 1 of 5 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #642, opened 2026-07-14
ci: grant actions write so claude review runs can cache

saeedkolivand:
## Summary

Every claude-review run logs "Cache reservation failed: cache write denied: token has no writable scopes" (e.g. run 29369668267) and re-downloads instead of caching: the jobs grant contents/pull-requests/id-token but GitHub's cache service requires `actions: write` on the job token for cache writes. Adds that scope to both jobs with an explanatory comment. Concurrency block untouched (documented cancel-in-progress gotcha). Only workflow using claude-code-action (verified).

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -72,6 +72,7 @@ jobs:
       contents: read
       pull-requests: write
       id-token: write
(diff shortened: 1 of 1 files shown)


saeedkolivand:
@coderabbitai review

saeedkolivand:

PR #674, opened 2026-07-15
fix: bring hmac keyinit trait into scope for the 0.13 api

saeedkolivand:
**Urgent — `main` doesn't compile.** hmac 0.13 (#672) moved `Hmac::new_from_slice` from an inherent method to the `KeyInit` trait, so `cargo check` fails in the extension_bridge pairing handshake.

**Fix:** one import — `use hmac::{Hmac, KeyInit, Mac};` in `handshake.rs`. Zero logic change: same MAC key material, message construction, and constant-time `verify_slice`; the known-answer-vector + tamper/reject tests pass unchanged.

**Also checked (no changes needed):** sha2 0.11 (#669) and tokio-tungstenite 0.30 (#670) — both major bumps, but every use site was already API-compatible (sha2 already imports `Digest`; tungstenite's `Utf8Bytes` derefs to `str` so `.to_string()` still works). aes-gcm 0.11 was fixed separately in #662.

### Verified (from `apps/desktop/src-tauri`)
`cargo check --all-targets --all-features --locked` clean · `clippy -- -D warnings` clean · `cargo test` 2417 passed · `cargo fmt --check` clean.

Third `main` breakage from a merged Dependabot **major** without a green Rust gate (after #660 TS 7 and #659 aes-gcm).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #675, opened 2026-07-15
fix: restore rewrite-mode review fixes lost in #649

saeedkolivand:
## Summary

Restores the **PR 11 rewrite-mode review fixes** that were lost when #649 merged
before they landed. #649 shipped the rewrite feature itself; this re-applies the
three review findings (all already resolved on that PR's threads, all three lenses
+ claude + CodeRabbit) that were sitting on a not-yet-merged commit at merge time.

## What lands

- **Value-drift guard on Accept/Restore** (`answer-fill.ts` — `CHANGED_SINCE_PICK`):
  `replaceFilledField` now compares the picked field's CURRENT text to the pick-time
  snapshot and refuses if the user edited it since — never clobbers an unseen manual
  edit with a rewrite of stale text. Extends the "never write an unexpected field"
  discipline to "never overwrite unexpected content".
- **Picker `Number('') === 0` fix** (`popup.ts`): resetting the picker to the
  placeholder silently re-picked index 0; now guarded so an empty selection resolves
  to `undefined`, not the first field. Adds `expectedValue` tracking on the rewrite
  target so the next Accept/Restore compares against the right baseline.
- **Validate-before-charge ordering** (`answer_assist.rs` — `validate_rewrite_fields`):
  rewrite inputs are validated as a pure step BEFORE the `ai_research` limiter is
  acquired, so a malformed rewrite request can't consume the daily charge. Adds the
  `assist_prompt_for_mode` pure mode→prompt seam.

## Why a separate PR

The fixes lived on `d6c244c4`, which hadn't merged when #649 landed (a fix-push kept
aborting on the pre-push sync-merge loop). Rather than force anything onto `main`, this
cherry-picks that commit onto current `main` (which already has the #674 hmac-import
repair), so `main` ends with the rewrite feature **and** its review fixes.

## Review trail

Fully reviewed already on #649: three lenses (extension-reviewer, tauri-security-reviewer,
ai-provider-expert), claude, and all CodeRabbit threads resolved. No new logic — this is
the exact reviewed diff, re-applied. Rust `extension_bridge` + `agent` suites, extension
vitest (collector/locate/replace + rewrite-UI + preset-parity), `lint:strict`, typecheck,
and both browser builds green.

### DIFF
--- a/apps/desktop/src-tauri/src/extension_bridge/answer_assist.rs
+++ b/apps/desktop/src-tauri/src/extension_bridge/answer_assist.rs
@@ -313,6 +313,50 @@ fn resolve_rewrite_instruction(preset: Option<&str>, instruction: &str) -> AppRe
     Ok(instruction.to_string())
 }
 
+/// The system prompt + max-token cap [`resolve_answer_assist`] passes to
+/// [`super::stream::compose_draft_stream`] for `mode` — draft always selects
+/// [`ANSWER_ASSIST_SYSTEM`]/[`ANSWER_ASSIST_MAX_TOKENS`], rewrite always
+/// selects [`super::answer_rewrite::REWRITE_SYSTEM`] (same token cap — no
+/// in-app precedent to size a distinct one for rewrite, see
+/// `ANSWER_ASSIST_MAX_TOKENS`'s own doc). A PURE function (no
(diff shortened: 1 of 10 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #679, opened 2026-07-15
fix(shared): map swedish to iso 639-1 'sv' not iso 639-3 'swe'

### LINKED ISSUE #676: Swedish language code mapped to ISO 639-3 'swe' instead of ISO 639-1 'sv'
In `packages/shared/src/language-detection.ts`, the Swedish entry in `LANGUAGE_MAP` returns `'swe'` (ISO 639-3) while every other language returns its ISO 639-1 code. `LANGUAGE_NAMES` is then keyed on `'swe'` to compensate, but the rest of the pipeline (analysis prompt, metadata, mismatch checks) expects ISO 639-1 codes, so a Swedish resume/job ad silently misroutes.

**Impact:** Swedish resumes never match Swedish job ads for the mismatch check, and downstream locale-aware prompts (`resumeConventions`, target-language routing) receive an unexpected code.

**Fix:** map `swe -> 'sv'` and key `LANGUAGE_NAMES` on `'sv'`.

thejesh23:
Fixes #676

Every other entry in `LANGUAGE_MAP` returns an ISO 639-1 code, but Swedish returned `'swe'` (ISO 639-3). `LANGUAGE_NAMES` was keyed on `'swe'` to hide the mismatch, but the rest of the pipeline expects ISO 639-1, so Swedish text silently misrouted.

- `LANGUAGE_MAP.swe` -> `'sv'`
- `LANGUAGE_NAMES` key `'swe'` -> `'sv'`


saeedkolivand:
Merged, thanks! `sv` is the right ISO 639-1 code, and this also quietly fixed the display name — it was keyed under `swe`, so nothing was actually resolving "Swedish". Good catch.

saeedkolivand:

PR #680, opened 2026-07-15
fix(analyze): don't flag language mismatch when a language is unknown

### LINKED ISSUE #677: validateAndRepair reports false language mismatch when a language is 'unknown' or missing
In `packages/prompts/src/analyze/validate.ts` (line 119):

```ts
mismatch: langs.mismatch === true || langs.resume !== langs.jobAd,
```

When the model omits `mismatch` (or emits it as anything other than `true`) and either language is `'unknown'` or missing, the fallback flips `mismatch` to `true`. Example: `{resume: 'en', jobAd: 'unknown'}` yields `mismatch: true`, and `{resume: 'en'}` (jobAd undefined) also yields `true`.

**Impact:** the UI shows a language-mismatch warning to users whose job ad language simply could not be detected. This diverges from the sibling implementation in `packages/shared/src/language-detection.ts` `detectLanguages()`, which correctly gates mismatch on both codes being known.

**Fix:** mirror the shared helper — only report mismatch when both codes are known and differ.

thejesh23:
Fixes #677

`validateAndRepair`'s mismatch fallback compared `langs.resume` to `langs.jobAd` directly, so a missing or `'unknown'` value on either side flipped `mismatch` to `true` (e.g. `{resume: 'en', jobAd: 'unknown'}` reported a mismatch).

Now mirrors `detectLanguages()` in `@ajh/shared`: mismatch only when both codes are present, non-`'unknown'`, and different.

### DIFF
--- a/packages/prompts/src/analyze/validate.ts
+++ b/packages/prompts/src/analyze/validate.ts
@@ -116,7 +116,16 @@ export function validateAndRepair(raw: string): AnalysisResult | null {
     detectedLanguages: {
       resume: ensureString(langs.resume, 'unknown'),
       jobAd: ensureString(langs.jobAd, 'unknown'),
-      mismatch: langs.mismatch === true || langs.resume !== langs.jobAd,
+      // Only flag mismatch when both languages are actually known — matches
+      // detectLanguages() in @ajh/shared. Otherwise an 'unknown' or missing
+      // side would falsely trip the mismatch warning.
+      mismatch:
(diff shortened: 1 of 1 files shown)


saeedkolivand:
This one was a real papercut — a false "language mismatch" warning any time one side came back `unknown`. Your fix matches how `detectLanguages` already treats it, which is exactly what I'd want. Appreciate you tracking it down. Merged!

saeedkolivand:

PR #681, opened 2026-07-15
fix(prompts): count zero words for empty or whitespace-only resume

### LINKED ISSUE #678: getResumeStats over-counts words on empty or whitespace-padded input
In `packages/prompts/src/context-manager/multi-pass.ts` (line 77):

```ts
const words = resume.split(/\s+/).length;
```

`String.prototype.split` returns empty strings for leading/trailing separators, and `['']` for an empty input, so:
- `''` -> reported as **1** word
- `'  hello world  '` -> reported as **4** words

**Impact:** minor — resume stats surfaced in the UI (`words`) are inflated by up to 2 phantom entries and never bottom out at 0 for an empty document.

**Fix:** trim first and filter empty tokens, e.g.:

```ts
const words = resume.trim() ? resume.trim().split(/\s+/).length : 0;
```

thejesh23:
Fixes #678

`resume.split(/\s+/).length` returned `1` for an empty string and inflated counts by 1-2 for leading/trailing whitespace, so the `words` field in `getResumeStats` was systematically wrong on the edges. Trim first, bail out on empty.

### DIFF
--- a/packages/prompts/src/context-manager/multi-pass.ts
+++ b/packages/prompts/src/context-manager/multi-pass.ts
@@ -74,7 +74,10 @@ export function getResumeStats(resume: string) {
   const pages = estimatePages(resume);
   const sections = detectSections(resume);
   const chars = resume.length;
-  const words = resume.split(/\s+/).length;
+  // `.split(/\s+/)` yields `['']` for empty input and empty tokens for
+  // leading/trailing whitespace, so trim first and bail out on empty.
+  const trimmed = resume.trim();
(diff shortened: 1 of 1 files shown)


saeedkolivand:
Ha, the eternal `''.split(/\s+/)` returning `['']` trap. Trimming first and short-circuiting on empty is spot on. Thanks for the clean fix and the clear repro in the issue — merged. And thanks for the three PRs today, genuinely helpful stuff. Hope to see more!

saeedkolivand:

PR #682, opened 2026-07-15
fix: own the active ai provider config in the backend so the renderer can't set base_url

saeedkolivand:
## Summary

Moves the **active AI generation provider config** (`activeProvider` + per-provider `{model, baseUrl}`) to a **backend-owned `AiConfigStore`** (SQLite, mirrors `EmbeddingConfig`), making the backend the single source of truth. Closes task #16 — the "twice-incurred" (actually **three-fold**) debt where the renderer's Zustand prefs were threaded into the backend via generation calls, the extension bridge, **and** autopilot.

### Why — a real trust boundary, not just cleanup

The renderer used to inject `provider/model/base_url` into every generation call and the backend **trusted** it. An XSS'd renderer could set `provider=openai` + `base_url=https://attacker/` and exfiltrate the stored API key + full prompt on every call over an **unguarded** egress path — and the extension bridge was worse, **persisting** the attacker `base_url` to disk. This flips ownership so `base_url` can only originate from an explicit user Settings action, validated on write.

### What changed

- **New `AiConfigStore`** (dedicated SQLite `DataStore`) — single source of truth for `{activeProvider, providers:{model,baseUrl}}`, wired into the GDPR reset registry + backup bundle (holds no secrets — keys stay in the keychain).
- **Compile-time security lock** — `provider` + `base_url` removed from `AiGenerateRequest` (Rust struct **and** shared Zod schema). Generation, research/salary, `generate_pipeline`, the extension bridge (`resolve_answer_assist`), and autopilot all resolve routing server-side via the new `Completer::from_active`.
- **Write-time base_url validation** — `net::ssrf::validate_provider_base_url`: http(s) only, cloud-metadata `169.254.169.254` blocked, **provenance not IP-filtering** (localhost/LAN gateways like Ollama/LM Studio stay allowed). `base_url` is only persisted for `openai-compatible` (the only kind that honors it) on both the writer and seed/import paths.
- **Extension bridge de-dup** — `extension_bridge_set_ai_assist_enabled` shrinks to a bare boolean (billable-consent opt-in unchanged, ADR-0011); old opt-in files stay honored and drop the stale `base_url` on next write.
- **Renderer flip** — the Zustand AI slice becomes a backend-backed `useActiveConfig` React Query hook with a boot prefetch; the settings UI writes via `ai_set_active_provider`/`ai_set_provider_settings` (switch-vs-edit split preserved); one-time post-hydration seed from existing prefs. `effort` + `modelLimits` stay renderer-side (not routing/egress).

### ⚠️ Behavior change

A **scheduled autopilot run now follows the currently-active provider**, not the one pinned when the schedule was created (owner-approved).

## Review trail

Three independent critics, **all approved, zero in-scope blockers**:
- **rust-backend-architect** — store mirrors the `DataStore` pattern, no `Mutex`-across-`await` on the hot path, `from_active` reproduces the resolve steps + fail-closed re-validate, reset/backup wired.
- **tauri-security-reviewer** — invariant fully enforced for its scope; residual renderer influence bounded/by-design. **Approval conditions met:** docs scoped to the flipped commands (module + ssrf overclaims corrected), and the pre-existing `agent_run` base_url path tracked as the immediate next PR.
- **frontend-reviewer** — query/prefetch/seed/dead-code-removal all verified; the one diff-introduced cold-boot caption flash (`StepAction`) is fixed + tested.

## Deferred follow-up (tracked)

`agent_run` ("prep this application") still resolves via a request-supplied `base_url` (pre-existing; not touched here) — the **same exploit class**, to be flipped next (`docs/NEXT_ISSUES.md #5`). **ADR-0012** (`ai-provider-base-url-provenance`) records the decision and deliberately **scopes** the closure claim to the flipped commands — it does not claim base_url is closed for *all* generation until that lands. (Also: NEXT_ISSUES #6/#7 track two LOW cleanups; ADR-0011 gets a superseded-consequence note.)

## Test plan

`cargo test` (2440+) + `clippy --all-features` + `pnpm typecheck` + `lint:strict` + `pnpm test` (3740+) + `gen:ipc:check` all green. New coverage: store round-trip/validation (reject non-http(s) + metadata addr, allow LAN), native-provider base_url dropped to NULL (writer + seed/import), single-shot seed, bridge opt-in back-compat, cold-boot caption guard. Honest gap (disclosed): no `tauri::test` mock-app harness for end-to-end `from_active` — locked by the compile-time field removal + store/validator unit tests.


saeedkolivand:
@claude review

saeedkolivand:
@coderabbitai review

saeedkolivand:
@claude review

saeedkolivand:

PR #693, opened 2026-07-16
fix: track bridge connections with a refcount and push status changes to settings

saeedkolivand:
## Summary

**Fixes the extension connection status in Settings** (user-reported): after pairing, the pill didn't flip to Connected until the next 30-second poll, and there was no way to refresh it manually. Digging in surfaced a real state bug underneath: `BridgeState.connected` was a single boolean that every socket set true on auth and false when *its* socket closed — so with two browsers paired (which the bridge fully supports: one shared token, per-socket HMAC), whichever disconnected last won. Chrome's MV3 service worker idle-closing its socket would show "not connected" while Firefox was still live.

## What changed

- **Connection refcount** — `AtomicUsize` instead of the boolean: increment only on a completed HMAC handshake (`FrameDecision::AuthOk`), decrement on teardown only if that connection had authenticated (per-connection flag, saturating arithmetic). `is_connected()` = count > 0, public shape unchanged. The original code's own comment ("not a refcount — promote if concurrent sockets become real") predicted exactly this.
- **Push event** — `extensionBridge:changed` (`{connected}`) emitted only on real 0→1/→0 transitions; contract + generated Rust constant in lockstep (`gen:ipc:check` pins it). The renderer treats it purely as a re-read signal — a new `useExtensionBridgeEvents` (mounted once in the root layout beside the existing event bridges) invalidates the status query; the 30s poll stays as fallback.
- **Refresh button** beside the status pill (shared `RefreshButton` primitive, same pattern as the update section).
- **Copy** (en/de): documents that multiple browsers can pair with the same token and that Regenerate token disconnects all of them.

## Review trail

extension-reviewer **PASS** (traced all five teardown paths — no double-increment, no unauthenticated decrement, no listener leak, no invalidate→emit loop; protocol untouched: the event is desktop-internal, zero extension files changed). tauri-security-reviewer **PASS** (refcount is informational only — the HMAC session remains the sole authorization boundary; unpaired clients can't move the count; payload carries only a bool; one waived LOW: no debounce on connect-churn events, amplification bounded to a cheap local read behind the pairing-token trust boundary).

## Test plan

New Rust tests: two-sockets-one-closes stays connected, unauthenticated close never decrements, saturating decrement, stable event name. Renderer: listener wiring/unsubscribe, invalidation, refresh button. Suites: bridge cargo 296 · shared 252 · desktop whole-graph 3779 · clippy `--all-features` · fmt · `gen:ipc:check` · typecheck · lint:strict — all green.

## Summary by CodeRabbit

* **New Features**
  * Added live extension connection status updates in Settings.
  * Added a refresh control for manually updating connection information.
  * Added support for checking whether a job has already been applied, including application details when available.
  * Clarified that one pairing token can connect multiple browsers.

* **Bug Fixes**
  * Connection status now remains accurate when multiple browsers are connected or disconnected.

* **Documentation**
  * Updated extension pairing guidance and connection behavior descriptions.


saeedkolivand:
@claude review

saeedkolivand:
@claude review

saeedkolivand:

PR #692, opened 2026-07-16
docs: refresh the status tracker, de-drift the knowledge base, retire finished audit worklists

saeedkolivand:
## Summary

**Docs-hygiene sweep** (user-requested, four asks in one PR): make the status tracker true, verify-then-retire the finished audit worklists, and de-drift the entire knowledge base. Every change here is backed by per-item verification — 40 legacy worklist items and ~75 knowledge-doc findings were each checked against current code/git by dedicated verification agents before any edit.

## What's here

- **ARCHITECTURE_STATUS.md refreshed**: 5 shipped features moved out of Planned/Backlog (salary answers #548/549, interviewer questions #383, humanize #563, 12-template gallery #590-594, auto-track Layer A #687); 3 missing sections added (Applications & Tracker, Email-Confirmation Watching 🚧, Browser Extension with its 8 shipped features); TypeScript row corrected to the per-importer 7/6 split; the stale hybrid-search warning and agent provider-snapshot note fixed; 9 verified-open remnants folded into the backlog.
- **Superseded worklists deleted** (user-confirmed): `AUDIT_TRIAGE.md` (15/20 items shipped, rest folded), `UX_AUDIT.md` (verified across all 13 action rows), `architecture-analysis.md` (4/6 shipped, 2 folded). All ~46 items carry evidence in the verification trail; git history preserves the files; 5 inbound references repointed. `AUDIT_REPORT.md` stays as the frozen historical ledger, with 6 requirement rows annotated closed/superseded per the verification.
- **Knowledge base de-drifted**: 19 docs corrected (+284/−213) — headline: `anti-abuse-limits.md` described a limiter that doesn't exist (RateLimiter/Semaphore/DashMap/H13) and now documents the real `Limiter`/`ConcurrencyGuard`/`RATE_LIMITED` machinery; `persistence.md`'s store table matched reality (AiConfigStore/EmailWatchStore/SpendStore in, backup-excluded stores out); `extension-domain.md` got the HMAC-v2 auth model and six-state phase list (email-watch lines deliberately untouched — the Layer C branch owns them); drift-prone hardcoded counts removed per the no-literals rule.
- **10 decision-records** re-pinned (moved files, renamed modules, stale counts/claims — pins only, no decision rewrites).
- **DEVELOPMENT.md** gains the codegraph section beside graphify (structural vs semantic routing).

## Verification trail

[...]

PR #694, opened 2026-07-16
ci: disable setup-node caching in the badge publisher (zizmor cache-poisoning)

saeedkolivand:
## Summary

zizmor v1.27.0 (pulled unpinned by the Security workflow) raises a new HIGH `cache-poisoning` finding: `.github/workflows/downloads-badge.yml`'s `setup-node@v6` step enables package-manager caching by default inside a workflow that holds `contents: write` and pushes to the badges branch — a poisoned cache could influence published artifacts. Every PR's 🛡️ zizmor check went red starting today (first seen on #692); the check is advisory, so nothing was blocked, but the finding is legitimate defense-in-depth.

## Fix

Two lines: `package-manager-cache: false` on the setup-node step (input name verified against the pinned SHA's `action.yml`) plus a comment citing the audit. Verified with a local zizmor 1.27.0 before/after run — the HIGH clears; the only remaining finding is the pre-existing low-confidence `artipacked` note on checkout, which that workflow needs (persisted credentials to push the badge).


saeedkolivand:

PR #687, opened 2026-07-16
feat: auto-track sent applications on a detected form submit (opt-in)

saeedkolivand:
## Summary

**Auto-track sent applications, Layer A** — an opt-in that marks a job **applied** the moment you submit its application form, so you don't have to remember to. Built entirely on the existing extension gesture + `applied.check`/`status.update` verbs; **no new permissions.**

## How it works

The extension only touches a page's DOM on a user gesture (`activeTab` + `scripting` — there's no persistent content script). So Layer A leverages that: after you invoke the extension on an application page (autofill / fill answers / import) **and** have enabled the new opt-in, a pure-DOM `submit-watch` arms a capture-phase `submit` + apply-button listener. On a detected submit it:
- re-checks the opt-in → `applied.check {url}` →
- **tracked & `saved`** → auto `status.update {to:'applied', auto:true}` (a desktop notification confirms it),
- **already applied** → silent no-op,
- **not tracked** → an action-badge nudge to import (never auto-creates).

It **never** blocks or alters the submit (observe-only), reads only the page URL (no form values), and never writes without a matching `saved` job.

## Consent & safety (both reviewers PASS, no HIGH/CRITICAL)

- **New "Auto-track sent applications" opt-in** — default **OFF**, desktop-enforced, cleared on factory reset. Sits beside the autofill/AI-assist toggles.
- **The auto write is gated server-side:** `handle_status_update` refuses `auto:true` when the opt-in is off — a client that lies can't force it. The manual popup "Mark as applied" stays unflagged and ungated exactly as before.
- **tauri-security-reviewer** settled the design: the enforced boundary is the server-side `saved→applied` transition allowlist + exact-URL CAS + HMAC-auth; a client-attested gesture/opt-in would be theater against a compromised extension, so it's kept as defense-in-depth only. No new write power vs. PR 2. The ungated device-local opt-in *read* is the right boundary.
- **No new permissions / manifest changes**; Firefox `data_collection_permissions` stays `['none']`; `submit-watch.js` is an import-free classic script (no SW crash risk).

## Honest limit (why #23 exists)
Only catches a submit **after** you've used the extension on that page, and full-page-nav submits are best-effort (SPA/Easy-Apply are reliable). Layer C (#23, local-only email-confirmation parsing) is the complement. Documented in ADR-0009 + `extension-domain.md`.

## Review trail
[...]

PR #696, opened 2026-07-17
feat: watch the inbox for application confirmations and notify to confirm (auto-track layer c)

saeedkolivand:
## Summary

**Auto-track Layer C goes live** — the watcher half of task #23, completing what #689's foundation started. Every ~15 minutes (or on "Check now"), the app reads your Gmail inbox **locally over IMAP**, fingerprints application-confirmation emails (English + German), fuzzy-matches them against your `saved` applications, and raises a Notification Center card asking you to confirm — **it never writes a status by itself**. This closes Layer A's honest gap: applications submitted outside the extension now get caught by the inbox signal.

## How it works

`email_watch_scheduler` (a separate L2 module mirroring the autopilot scheduler: startup grace, interval tick, failure backoff capped at 2h, a RunGuard so a manual check and a scheduled tick can never interleave, 60s check-now rate limit) → the IMAP read path (fresh `EXAMINE`, `UID SEARCH` scoped above the watermark when the mailbox generation is unchanged, headers first, `BODY.PEEK` for fingerprint hits only — **the watcher never marks your real messages read**) → the EN+DE parser (`mail-parser`, subject fingerprint families incl. dative German forms and "Eingangsbestätigung"; sender-domain hints boost but never gate) → the token-Jaccard company/title matcher against `saved` applications → stamp the `seen` table, then notify with a route to the matched application.

## Correctness story (the hard-won part)

The review pipeline earned its keep on this one. Beyond the three domain critics (security PASS — content-containment traced to every sink, read-only verified at the protocol level; match-quality PASS — recall broadened, rejection-parity pinned; rust-arch HIGH — the RunGuard, fixed), the three-pass `/review` ensemble caught two silent-failure HIGHs that everything else missed:

- **Mailbox renumber suppression**: a UIDVALIDITY change reset the watermark but kept stale `seen` rows, so reused UIDs would silently swallow real confirmations forever. Now the reset wipes `seen` transactionally, with the regression test.
- **Disconnect-mid-tick resurrection**: the new post-tick writes lacked the `address IS NOT NULL` guard #689's review established — a disconnect landing during the multi-second IMAP round trip could get its wipe resurrected and even fire a notification from the just-disconnected mailbox. All three writes are now guarded and the notify path re-checks the account exists.

Also from review: the UID-scoped search (no more full 30-day refetch per tick), backoff no longer inflated by benign concurrent-run refusals, and a cross-language parity test pinning the rate-limit sentinel.

## Honest limits (ADR-0013 + `docs/knowledge/email-watch-domain.md`)

INBOX only; the employer must actually send a confirmation (Greenhouse's is opt-in); fuzzy matching can pick the wrong role at the same company (confirm-gate is the backstop); **a rejection email fingerprints identically to a confirmation** — harmless under notify+confirm, and pinned in tests as the hard blocker for any future auto-write without a body negative-signal check; Gmail-only v1.

## Test plan

`cargo test` **2538** (email_watch + scheduler suites: renumber wipe, concurrent-clear interleavings, RunGuard exactly-one-proceeds, backoff classification, EN+DE fixtures incl. near-misses and adversarial rejection/interview subjects, parity tests) · clippy `--all-features` · fmt · `cargo deny` · `gen:ipc:check` · typecheck 12/12 · lint clean · `pnpm test` 342 files / 3799. IMAP integration remains manual-smoke by design (no fake-IMAP harness; documented).

Completes task #23. Layer A (#687) + Layer C now cover both sides of "did I apply?".


saeedkolivand:
@claude review

saeedkolivand:
@coderabbitai review

saeedkolivand:
@claude review

saeedkolivand:

PR #699, opened 2026-07-17
feat: use dom_smoothie readability pass in scrape_url generic extraction fallback

saeedkolivand:
## What

On job pages without JSON-LD (or the extension's DOM hint node), `scrape_url`'s generic extraction fell back to a largest-`main`/`article`-block guess that let nav/footer/boilerplate through. This adds a Mozilla-Readability-grade pass (`dom_smoothie`, MIT) as the **last-resort rung**, keeping the largest-block guess as the final floor.

## How

- `readability_content_text()` in `scraping/scrape_url/mod.rs` runs ONLY after the higher-precision rungs (extension hint node, JSON-LD `JobPosting`) miss — those are unchanged.
- Guarded by `is_probably_readable()` before `parse()`; any error / false pre-check / empty output → `None` → falls through to the unchanged `main_content_text`.
- `TextMode::Markdown` keeps the output shape identical to the existing fallbacks (no new format contract; `JobDetailPane` already renders markdown).
- `max_elements_to_parse: 4000` bounds parsing of hostile/huge HTML (trips → `TooManyElements` → clean fall-through) — defense-in-depth flagged by both reviewers.

## Dependency

`dom_smoothie 0.18` (MIT) + its transitives — all already in the `deny.toml` allowlist; no GPL/AGPL pulled in; blast radius is 5 crates (verified against `Cargo.lock`; the rest already ship via existing deps).

## Tests

- `parse_from_html_readability_skips_in_article_decoy_block` — a decoy nested INSIDE `<article>` (which the old selector-scoped guess would NOT have excluded) is dropped, proving the readability differentiator; verified it fails without the change.
- `readability_content_text_returns_none_for_sparse_page` — sparse posting fails `is_probably_readable()` → clean fall-through.
- Full crate `cargo test --lib`: 2546 passed; `cargo clippy --lib` clean.

## Reviews

`scraping-applier-expert` + `tauri-security-reviewer` both approved — no HIGH/CRITICAL. Supply-chain clean; parse element-capped per both reviewers' advisory.

### DIFF
--- a/apps/desktop/src-tauri/Cargo.lock
+++ b/apps/desktop/src-tauri/Cargo.lock
@@ -101,6 +101,7 @@ dependencies = [
  "criterion",
  "dbus-secret-service-keyring-store",
  "docx-rs",
+ "dom_smoothie",
  "encoding_rs",
  "feed-rs",
  "flate2",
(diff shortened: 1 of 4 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks for the review — addressed the LOWs and consciously deferred the MEDIUM. Pushed `adeea744`.

**Fixed**
- 🔵 **Logs** (`scrape_url/mod.rs`): both `log::warn!` in `readability_content_text` → `log::debug!`, now logging **only the host** (via the file's existing `Url::parse(...).host_str()` convention) — no raw scraped URL, and no over-signalling on best-effort enrichment that always has a clean fallback.
- 🔵 **Supply-chain wording** (`Cargo.toml`): reworded so it states precisely that the crate's **MIT license** is permitted by `deny.toml`'s `[licenses] allow` list (enforced by CI `cargo deny check`), not a per-crate allowlist.

**Deferred (tracked in-code with `// TODO(perf):`)**
- 🟡 **`spawn_blocking`**: agreed it's worth doing, but out of scope here. `parse_from_html` is *already* fully synchronous CPU work (every rung calls `Html::parse_document`) and is shared with the extension-bridge Scan path, so the correct fix wraps the whole generic-HTML fallback in `spawn_blocking` at its async call sites — a broader refactor rather than a half-measure that special-cases only this one pass.
- 🔵 **"readability wins even when thinner than the floor"**: left as the documented, intentional design (readability is higher-precision; `is_probably_readable()` gates the thin cases).

`cargo check` / `cargo test --lib scraping::scrape_url::` (86 passed) / `cargo clippy --lib` all green.

saeedkolivand:

PR #707, opened 2026-07-18
fix: hold the gl experience behind a pre-launch gate

saeedkolivand:
## What

**Recommend merging this ahead of the normal review cycle — it affects live production.**

The Pages pipeline has served the new app's index.html since P0 (#702), which means the capability gate has been mounting the in-progress GL skeleton for capable desktop visitors since P1 (#703) — placeholder beats and all. Per the ruling that production flips only when everything is done:

- `gateVerdict()` now returns `gl:false` (reason `pre-launch`) unless the build is dev or the URL carries `?gl=1`
- Every production visitor gets the legacy page exactly as it behaves today (semantic layer + legacy.ts)
- The GL work stays previewable live at `aijobhunter.app/?gl=1`
- The P7 flip removes this guard (documented in the phase task + code comment)

## Verification

typecheck, eslint --max-warnings 0, static-export build + passthrough merge, ASCII scan — all green. One file changed.

### DIFF
--- a/apps/landing/src/engine/gate.ts
+++ b/apps/landing/src/engine/gate.ts
@@ -16,8 +16,18 @@ export interface GateVerdict {
 
 let cached: GateVerdict | null = null;
 
+// Pre-launch guard: the GL experience is still being built (phases P4-P7,
+// docs/adr/0014). Until the production flip, GL mounts only in dev or with an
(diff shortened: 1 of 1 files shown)


saeedkolivand:

PR #725, opened 2026-07-19
fix(referral): pass ISO code instead of display name to locale

### LINKED ISSUE #724: fix: referral drafts always generated in English for non-English resumes
## Describe the bug

`useReferralDraft` passes `detectLanguages(resume, '').resumeName` (a display name like `\"Deutsch\"`, `\"Français\"`) into `generateReferral`'s `locale` field. `locale` is routed through `safeLocale()` (`apps/desktop/src/renderer/lib/generate/locales.ts:50`) which only recognises ISO 639-1 codes and silently clamps any unrecognised value to `'en'`.

Net effect: every non-English resume produces an English referral draft, regardless of the user's language.

## Where

- `apps/desktop/src/renderer/features/documents/components/TailorFlow/ReferralModal/useReferralDraft.ts:101`
- `apps/desktop/src/renderer/features/documents/components/TailorFlow/ReferralModal/useReferralDraft.ts:150`

## Steps to reproduce

1. Load a résumé written in a non-English language (e.g. German).
2. Open the Referral Modal and pick any channel.
3. Generate a draft — it comes back in English.

## Expected

The draft is written in the résumé's language (the intent the surrounding comment describes).

## Fix

Pass the ISO code (`.resume`) instead of the display name (`.resumeName`).

thejesh23:
Closes #724.

## Summary

\`useReferralDraft\` was passing the language *display name* (\`\"Deutsch\"\`) into \`locale\`, which \`safeLocale()\` doesn't recognise and clamps to \`'en'\`. Every non-English résumé produced an English referral draft. Fix: pass \`.resume\` (ISO 639-1 code) instead of \`.resumeName\`.

## Changes

- \`useReferralDraft.ts\`: swap \`.resumeName\` → \`.resume\` at both call sites.

## Type of change

- [x] \`fix\` — Bug fix

## Testing

- [x] \`pnpm exec vitest run useReferralDraft.test.ts\` — **26/26 pass**
- [ ] Ran full \`pnpm test\` (skipped — a pre-existing zustand-persist / localStorage issue causes ~86 unrelated tests to fail on \`main\` in a fresh clone; my change doesn't touch that surface).
- Manual end-to-end verification not run (no local Ollama setup).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved referral generation language detection by using the correct resume language code.
  * Prevented supported locales from incorrectly defaulting to English during referral drafting and refinement.

### DIFF
--- a/apps/desktop/src/renderer/features/documents/components/TailorFlow/ReferralModal/useReferralDraft.ts
+++ b/apps/desktop/src/renderer/features/documents/components/TailorFlow/ReferralModal/useReferralDraft.ts
@@ -96,9 +96,9 @@ export function useReferralDraft({
         format: channel,
         charLimit: channel === 'connection_note' ? CONNECTION_NOTE_LIMIT : undefined,
         model,
-        // Write the message in the résumé's language — same client-side detection
-        // the cover-letter/metadata path uses; `safeLocale` clamps it downstream.
-        locale: detectLanguages(resume, '').resumeName,
+        // Write the message in the résumé's language — pass the ISO 639-1 code
+        // (not the display name) so `safeLocale` downstream doesn't collapse it to 'en'.
+        locale: detectLanguages(resume, '').resume,
         onToken: (tok) => setDraft((prev) => prev + tok),
         signal: controller.signal,
       });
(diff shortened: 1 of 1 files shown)


saeedkolivand:

PR #731, opened 2026-07-19
fix(analyze): drop false language mismatch when a language is empty string

### LINKED ISSUE #730: fix: analyze mismatch flag lights up when a language field is an empty string
## Describe the bug

\`validateAndRepair\` in \`packages/prompts/src/analyze/validate.ts\` coerces an empty-string language field to \`'unknown'\` for the output (lines 117–118) but then decides mismatch on the ORIGINAL raw value (lines 124–127). Empty string is a string, and \`\"\" !== 'unknown'\`, so a model reply like \`{ resume: \"\", jobAd: \"fr\" }\` produces:

\`\`\`
{ resume: 'unknown', jobAd: 'fr', mismatch: true }
\`\`\`

The output flags a mismatch while simultaneously reporting the resume language as \`'unknown'\` — self-contradictory, and the amber mismatch banner lights up in the UI.

## Expected

Mismatch is only flagged when both languages are actually known — the behaviour the surrounding comment (lines 119–121) already describes.

## Fix

Compute mismatch on the coerced values, not the raw ones.

thejesh23:
Closes #730.

## Summary

\`validateAndRepair\` coerced an empty-string language to \`'unknown'\` in the output but then decided \`mismatch\` on the raw pre-coerce value, producing self-contradictory output like \`{ resume: 'unknown', jobAd: 'fr', mismatch: true }\`. Compute mismatch on the coerced values so the guard the comment describes actually applies.

## Changes

- \`packages/prompts/src/analyze/validate.ts\`: coerce \`resume\` / \`jobAd\` once and reuse them in the mismatch expression.
- \`packages/prompts/src/analyze/analyze.test.ts\`: new regression covering the empty-string case.

## Type of change

- [x] \`fix\` — Bug fix

## Testing

- [x] \`pnpm --filter @ajh/prompts test\` — **505/505 pass** (includes the new regression).
- [ ] Ran full \`pnpm test\` (skipped — same pre-existing zustand-persist / localStorage vitest env issue that fails ~115 unrelated tests on unmodified \`main\` in a fresh clone; CI on GitHub Actions runs the full suite).

### DIFF
--- a/packages/prompts/src/analyze/analyze.test.ts
+++ b/packages/prompts/src/analyze/analyze.test.ts
@@ -232,6 +232,18 @@ describe('validateAndRepair', () => {
     expect(validateAndRepair('totally not json')).toBeNull();
   });
 
+  it('does not flag mismatch when a language field is an empty string', () => {
+    // The empty-string resume must coerce to 'unknown', and the mismatch check
+    // must run on the coerced value — otherwise the badge lights up alongside a
+    // resume language displayed as "unknown", which is self-contradictory.
+    const result = validateAndRepair(
+      JSON.stringify({ detectedLanguages: { resume: '', jobAd: 'fr' } })
+    ) as AnalysisResult;
+    expect(result.detectedLanguages.resume).toBe('unknown');
+    expect(result.detectedLanguages.jobAd).toBe('fr');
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #764, opened 2026-07-21
fix(extraction): do not emit bare w:p docx paragraphs twice

### LINKED ISSUE #763: fix(extraction): DOCX with bare <w:p> paragraphs yields duplicated + boundary-mashed text
## Describe the bug

`parse_document` in `apps/desktop/src-tauri/src/extraction/docx.rs:96` splits the document XML on **both** paragraph spellings unconditionally:

```rust
for para_xml in xml.split("<w:p ").chain(xml.split("<w:p>").skip(1)) {
```

Word writes paragraphs with attributes (`<w:p w:rsidR="…">`), but Google Docs, LibreOffice and `python-docx` write a **bare** `<w:p>`. For a bare-`<w:p>` document:

- `xml.split("<w:p ")` (note the trailing space) matches nothing, so it yields **the whole document as one element**, which `parse_paragraph` mines into a single blob with no paragraph boundaries;
- the chained `xml.split("<w:p>").skip(1)` *also* yields every paragraph.

So all résumé text is emitted **twice** — once mashed together without separators, once correctly per paragraph. MS Word documents escape the bug only because their `w:rsidR` attributes make the `<w:p>`-split branch empty.

## Steps to reproduce

Extract a DOCX whose `word/document.xml` body is:

```xml
<w:p><w:r><w:t>Jane Doe</w:t></w:r></w:p><w:p><w:r><w:t>Engineer</w:t></w:r></w:p>
```

## Expected behaviour

```
Jane Doe
Engineer
```

## Actual behaviour

```
Jane DoeEngineer
Jane Doe
Engineer
```

The duplicated + boundary-less copy is fed to embedding/matching and to the AI prompts, so the résumé text the app reasons about is wrong for every Google Docs / LibreOffice / python-docx export.

## Environment

- **OS**: macOS 15 (platform-independent — pure parsing logic)
- **Ollama version**: n/a (no AI involved)
- **Model in use**: n/a

## Additional context

The existing unit test can't catch this: it asserts with `contains()`, which stays true when the text is duplicated.

A fix is ready in a PR that scans for `<w:p` followed by `' '` or `'>'` instead of splitting on one spelling, so mixed documents are also handled correctly.

thejesh23:
## Summary

`parse_document` emitted every paragraph **twice** — once mashed together with no boundaries, once per paragraph — for any DOCX that uses the bare `<w:p>` tag (Google Docs, LibreOffice, python-docx). Fixes #763.

## Type of change

- [x] `fix` — Bug fix

## Changes

- `paragraph_slices()` now scans for the `<w:p` tag and requires the next byte to be `' '` or `'>'`, instead of `xml.split("<w:p ").chain(xml.split("<w:p>").skip(1))`.
  - The old chain was unconditional: for a bare-`<w:p>` document, `split("<w:p ")` (trailing space) matched nothing and returned **the whole document as a single element**, which `parse_paragraph` mined into one boundary-less blob — and the chained `split("<w:p>").skip(1)` then emitted each paragraph again.
  - MS Word documents escaped the bug only because their `w:rsidR` attributes leave the `<w:p>`-split branch empty.
  - Requiring `' '`/`'>'` after `<w:p` is what keeps `<w:pPr>` / `<w:pict>` from being treated as paragraph starts.
- Documents that **mix** both spellings are now handled correctly (the old code could not).
- Four regression tests in `docx.rs`.

Before → after for `<w:p><w:r><w:t>Jane Doe</w:t></w:r></w:p><w:p><w:r><w:t>Engineer</w:t></w:r></w:p>`:

```
- "Jane DoeEngineer\nJane Doe\nEngineer"
+ "Jane Doe\nEngineer"
```

## Testing

- [x] Ran `cargo test --lib extraction` locally — 49 passed, 0 failed (includes the pre-existing `docx_extracts_paragraphs` / `docx_resolves_hyperlink` tests).
- [x] Ran `cargo test --lib` on the base commit for a baseline — 2689 passed, 0 failed.
- [x] `cargo fmt` clean.

The new tests assert the **exact** extracted string rather than using `contains()`; the existing `contains()`-based coverage stays green even when the text is duplicated, which is why this went unnoticed.

## Checklist

- [x] New public APIs are documented (`paragraph_slices` is private, documented inline)

### DIFF
--- a/apps/desktop/src-tauri/src/extraction/docx.rs
+++ b/apps/desktop/src-tauri/src/extraction/docx.rs
@@ -93,7 +93,7 @@ fn parse_document(xml: &str, rels: &HashMap<String, String>) -> (String, Vec<Lin
     let mut links: Vec<Link> = Vec::new();
 
     // Split on paragraph boundaries.
-    for para_xml in xml.split("<w:p ").chain(xml.split("<w:p>").skip(1)) {
+    for para_xml in paragraph_slices(xml) {
         let para_text = parse_paragraph(para_xml, rels, &mut links);
         let trimmed = para_text.trim().to_string();
         if !trimmed.is_empty() {
@@ -102,6 +102,38 @@ fn parse_document(xml: &str, rels: &HashMap<String, String>) -> (String, Vec<Lin
     }
 
     (paragraphs.join("\n"), links)
+}
+
+/// One slice per `<w:p …>` / `<w:p>` paragraph, each running up to the next
+/// paragraph start (any leading preamble before the first `<w:p` is dropped).
+///
+/// Writers disagree on the spelling: MS Word emits `<w:p w:rsidR="…">` while
+/// Google Docs / LibreOffice / python-docx emit a bare `<w:p>`, and a document
+/// can contain both. Splitting on one spelling and chaining the other emitted
+/// every paragraph TWICE for the bare form (the `"<w:p "` split matched nothing
+/// and returned the whole document as a single boundary-less element), so scan
+/// for the tag instead. The byte after `<w:p` must be `' '` or `'>'` — that is
+/// what keeps `<w:pPr>`/`<w:pict>` from being mistaken for a paragraph.
+fn paragraph_slices(xml: &str) -> Vec<&str> {
+    let bytes = xml.as_bytes();
+    let mut starts: Vec<usize> = Vec::new();
+    let mut from = 0usize;
+    while let Some(rel) = xml[from..].find("<w:p") {
+        let at = from + rel;
+        let after = at + "<w:p".len();
+        if matches!(bytes.get(after), Some(b' ') | Some(b'>')) {
+            starts.push(at);
+        }
+        from = after;
+    }
+    starts
+        .iter()
+        .enumerate()
+        .map(|(i, &start)| {
+            let end = starts.get(i + 1).copied().unwrap_or(xml.len());
+            &xml[start..end]
+        })
+        .collect()
 }
 
 fn parse_paragraph(xml: &str, rels: &HashMap<String, String>, links: &mut Vec<Link>) -> String {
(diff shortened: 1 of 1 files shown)


saeedkolivand:
@claude review

thejesh23:
Thanks for the review — addressed in `fcb8cc7`.

**1. `<w:p/>` self-closing empty paragraphs** — verified your read rather than taking it on trust, and it holds: the byte after `<w:p` is `/`, so no slice opens and the element folds into the previous one. It carries no runs, so it contributes no text either way, and the old chained splits ignored it too — behaviour is unchanged, not newly introduced. Pinned it with your suggested case plus a line in the doc comment:

```rust
#[test]
fn self_closing_empty_paragraph_neither_splits_nor_duplicates() {
    let xml = body(&format!("{}<w:p/>{}", para("<w:p>", "A"), para("<w:p>", "B")));
    let (text, _) = parse_document(&xml, &HashMap::new());
    assert_eq!(text, "A\nB");
}
```

**2. Dead block in `parse_paragraph` (`:143-148`)** — agreed it's a real cleanup candidate, and agreed it's out of scope here. Left untouched so this PR stays a single-behaviour fix. Happy to send it as a separate `refactor` PR if you'd like it.

**Nit** — taken; `xml.as_bytes().get(after)` inline, local binding dropped.

Re-verified locally after the change:

- `cargo test --lib extraction` — **50 passed**, 0 failed
- `cargo clippy --lib --all-features -- -D warnings` — clean
- `cargo fmt --check` — clean

thejesh23:
`🧪 Tests` went red on `fcb8cc7`, but **every test passed** — it's an unrelated flaky unhandled error, and I don't have re-run rights here. Details so you can judge:

```
Test Files  369 passed (369)
      Tests  4021 passed (4021)
     Errors  1 error

⎯⎯⎯ Uncaught Exception ⎯⎯⎯
 ❯ resolveUpdatePriority  react-dom-client.development.js:1308:7
 ❯ dispatchSetState       react-dom-client.development.js:9126:14
 ❯ Timeout._onTimeout     apps/desktop/src/renderer/features/settings/components/preferences/JobLocationPreferences/index.tsx:127:44
This error originated in ".../JobLocationPreferences.test.tsx"
```

Rust was clean too: `2718 tests run: 2718 passed, 26 skipped`.

The unhandled error makes vitest exit 1, so `steps.js-tests.outcome == 'failure'` trips the `❌ Check for Test Failures` gate even though nothing failed.

**Why it's not this PR:** the diff is one Rust file (`extraction/docx.rs`). The **first** CI run on this same PR passed `🧪 Tests`; the only change since is a Rust test plus a comment. Same JS code, different result.

**Root cause** — `JobLocationPreferences/index.tsx:127`:

```ts
onBlur={() => setTimeout(() => setShowSuggestions(false), 150)}
```

The 150 ms timer is never cleared, so if the test file finishes within that window the callback fires after jsdom is torn down and React's `resolveUpdatePriority` touches `window`. That's why it's intermittent — it landed on 4 of the 20 PRs I opened today (#764, #770, #776, #794), all on unrelated diffs.

Could you re-run the job? And if you'd like, I'm happy to send a separate PR storing the timer id and clearing it on unmount (plus the same for the `setTimeout(updateDropdownPosition, 0)` on the line above) — just say the word and I'll open an issue for it first.

saeedkolivand:
Thanks for this one. Nice catch on the doubled bare paragraphs, that kind of duplication is easy to miss in extraction output. Merged.

saeedkolivand:

PR #766, opened 2026-07-21
fix(extraction): stop rtf hex escape from slicing a str mid-character

### LINKED ISSUE #765: fix(extraction): RTF hex escape slices a str by byte index and panics on a non-char boundary
## Describe the bug

`rtf_to_text` (`apps/desktop/src-tauri/src/extraction/rtf.rs:157`) decodes the `\'xx` hex escape by slicing the **`str`** with byte indices:

```rust
b'\'' => {
    // \'xx hex byte (Windows-1252).
    if i + 3 < n {
        if let Ok(v) = u8::from_str_radix(&rtf[i + 2..i + 4], 16) {
```

`rtf` is a `Cow<str>` (from `String::from_utf8_lossy`) and `i` is a **byte** index. Slicing a `str` by a byte range panics unless both ends land on UTF-8 char boundaries, and the only guard here is a length check (`i + 3 < n`) — never a boundary check. So when a multi-byte scalar starts at `i + 3`, the end index `i + 4` falls **inside** that character and the process panics:

```
byte index 4 is not a char boundary; it is inside '€' (bytes 3..6) of `\'€`
```

That aborts extraction for the whole document. The parser is explicitly written to tolerate malformed input everywhere else, so a malformed/unusual RTF should degrade, not panic.

## Steps to reproduce

Import an RTF résumé containing a `\'` escape immediately followed by a multi-byte character, e.g. the 3-byte `€` (`E2 82 AC`):

```
{\rtf1\ansi \'€}
```

## Expected behaviour

The malformed escape is skipped (or best-effort decoded) and the rest of the document still extracts.

## Actual behaviour

Panic: `byte index N is not a char boundary`, aborting extraction of that résumé.

## Environment

- **OS**: macOS 15 (platform-independent — pure parsing logic)
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

The loop already keeps a `bytes` slice (`let bytes = rtf.as_bytes()`), so the fix is to read the two hex digits out of `bytes` instead of re-slicing the `str`. A PR is ready.

thejesh23:
## Summary

`rtf_to_text` decoded the `\'xx` hex escape by slicing the **`str`** with byte indices, which panics whenever the escape is followed by a multi-byte character — aborting extraction of the whole résumé. Fixes #765.

## Type of change

- [x] `fix` — Bug fix

## Changes

- Read the two hex digits out of the existing `bytes` slice instead of re-slicing `rtf`:
  - `rtf` is a `Cow<str>` and `i` is a **byte** index, so `&rtf[i + 2..i + 4]` panics unless both ends are UTF-8 char boundaries. The only guard was a length check (`i + 3 < n`), which cannot rule that out.
  - Reproduced standalone with the pre-fix expression verbatim:
    ```
    thread 'main' panicked at repro.rs:8:36:
    end byte index 20 is not a char boundary; it is inside '€' (bytes 18..21 of string)
    ```
- Digits are validated as ASCII hex before parsing, so a malformed escape can no longer be mis-parsed (`from_str_radix` otherwise accepts e.g. a leading `+`).
- A malformed escape now advances by 2 (dropping just the `\'`) rather than by 4, so the following character's bytes are not cut in half. The truncated-at-end-of-input branch (`i = n`) is unchanged.
- Three regression tests.

## Testing

- [x] `cargo test --lib extraction::rtf` — 8 passed, 0 failed (includes the pre-existing `decodes_unicode_and_hex`, `recovers_hyperlink_fields`, `full_extract_sets_rtf_source_and_appends_links`).
- [x] `cargo fmt` clean.

`well_formed_hex_escape_still_decodes` pins the unchanged happy path (`\'e9` → `é`).

## Checklist


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Slicing UTF-8 on a byte offset is the kind of panic that only shows up when someone feeds in a resume with accented characters, good that it's fixed before anyone hit it in the wild. Merged.

saeedkolivand:

PR #768, opened 2026-07-21
fix(scraping): advertise only gzip on the linkedin client

### LINKED ISSUE #767: fix(scraping): LinkedIn client advertises br/deflate but can only decode gzip
## Describe the bug

The LinkedIn HTTP client advertises three content encodings but can only decode one:

- `apps/desktop/src-tauri/src/scraping/linkedin/client/mod.rs:58` sends
  ```rust
  headers.insert(reqwest::header::ACCEPT_ENCODING, HeaderValue::from_static("gzip, deflate, br"));
  ```
- `reqwest` is declared with `default-features = false` and **no** `gzip` / `brotli` / `deflate` feature (`Cargo.toml:94`), so it performs no automatic decompression.
- `get_html` (`:169`) decodes manually, and only for the gzip magic bytes `1f 8b`:
  ```rust
  let body = if bytes.starts_with(&[0x1f, 0x8b]) { /* GzDecoder */ } else {
      String::from_utf8(bytes.to_vec())
          .map_err(|e| AppError::Parse(format!("response was not valid UTF-8: {e}")))?
  };
  ```

So if LinkedIn's edge honours the advertised `br` (or `deflate`), the still-compressed bytes fall into the `String::from_utf8` branch, fail, and the request returns `AppError::Parse("response was not valid UTF-8")`.

## Steps to reproduce

1. Run a scrape that includes the LinkedIn board.
2. Have the LinkedIn edge answer the guest endpoint with `Content-Encoding: br` (it is advertised as acceptable, so this is entirely up to the server).

## Expected behaviour

The response is decoded and the board returns jobs.

## Actual behaviour

`get_html` returns `Err(AppError::Parse("response was not valid UTF-8: …"))`. That propagates out of `search_guest` on page 0, so the LinkedIn board yields **nothing** for the whole run — and the error text points at UTF-8, not at compression, so it reads like a LinkedIn parsing change rather than a client bug.

## Environment

- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

The client should advertise only what it can actually decode. A PR is ready that sends `Accept-Encoding: gzip` and adds a test pinning the header to the decoder's real capability.

thejesh23:
## Summary

The LinkedIn client asked for `gzip, deflate, br` but can only decode gzip, so a `br`/`deflate` response failed the whole board with a misleading `"response was not valid UTF-8"`. Fixes #767.

## Type of change

- [x] `fix` — Bug fix

## Changes

- `Accept-Encoding` is now `gzip` — exactly what `get_html` can decode.
  - `reqwest` is declared `default-features = false` with no `gzip`/`brotli`/`deflate` feature (`Cargo.toml:94`), so it performs **no** automatic decompression.
  - `get_html` (`:169`) decodes manually and only on the gzip magic bytes `1f 8b`; anything else goes to `String::from_utf8(bytes)`, which fails on compressed data and returns `AppError::Parse("response was not valid UTF-8")`.
  - That error propagates out of `search_guest` on page 0, so the LinkedIn board returns nothing for the run — and the message points at UTF-8 rather than at compression.
- New test `accept_encoding_advertises_only_what_get_html_can_decode` pins the header to the decoder's real capability, so re-adding `br` without also adding a decoder fails the suite.

This is the minimal correct fix. The alternative — enabling reqwest's `brotli`/`gzip`/`deflate` features and deleting the manual `GzDecoder` block — is a larger dependency change and is left to the maintainer's judgement.

## Testing

- [x] `cargo test --lib scraping::linkedin` — 35 passed, 0 failed.
- [x] `cargo fmt` clean.

## Checklist


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Advertising encodings we can't decode was always going to bite eventually, gzip-only is the right call for the linkedin client. Merged.

saeedkolivand:

PR #770, opened 2026-07-21
fix(scraping): read the remoteok id per json variant

### LINKED ISSUE #769: fix(scraping): RemoteOK string id is JSON-stringified, embedding quotes in the posting id
## Describe the bug

`apps/desktop/src-tauri/src/scraping/boards/remoteok/mod.rs:98` normalises the RemoteOK `id` with `Value::to_string()`:

```rust
let id_str = id.map(|i| i.to_string()).unwrap_or_else(|| "".to_string());
```

`id` is an `Option<serde_json::Value>` because RemoteOK sends a **number** for some rows and a **string** for others. `Value::to_string()` is the JSON *serialisation*, so `Value::String("1234567")` becomes `"\"1234567\""` — **with the quotes**. Only the `Number` variant happens to serialise to clean digits.

Those quotes then travel into the posting identity:

```rust
id: format!("{}:{}", self.id(), id_str),   // remoteok:"1234567"
external_id: Some(id_str.clone()),         // "1234567"
```

## Steps to reproduce

1. Have `https://remoteok.com/api` return a row whose `id` is a JSON string, e.g. `{"id":"1234567","position":"Rust Engineer", …}`.
2. Run a scrape including the RemoteOK board.
3. Inspect the resulting posting's `id` / `externalId`.

## Expected behaviour

`id == "remoteok:1234567"`, `externalId == "1234567"`.

## Actual behaviour

`id == "remoteok:\"1234567\""`, `externalId == "\"1234567\""` — literal quote characters embedded in the key.

That key is the stable PostingsCache key, so it is what dedup and interaction tracking match on. A row that arrives as a number on one run and a string on another (or vice versa) yields two different identities for the same job.

## Environment

- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

The existing unit tests only exercise a numeric id (`serde_json::json!(123)`), which is exactly the variant that escapes the bug. A PR is ready that reads the value per variant and adds coverage for both.

thejesh23:
## Summary

RemoteOK sends `id` as a JSON number for some rows and a JSON **string** for others. `Value::to_string()` JSON-serialises, so the string variant arrived with literal quotes and corrupted the posting id / `external_id` — the stable PostingsCache key. Fixes #769.

## Type of change

- [x] `fix` — Bug fix

## Changes

- New private helper `normalize_id()` reads the value per variant (`as_str` → `as_i64` → `as_u64`) instead of `Value::to_string()`.
  - `Value::String("1234567").to_string()` is `"\"1234567\""`; only the `Number` variant serialised to clean digits.
  - The quotes flowed into `id: format!("{}:{}", self.id(), id_str)` → `remoteok:"1234567"` and `external_id` → `"1234567"`, so the same job could take two identities depending on which type the API happened to send.
- Anything that is neither a string nor an integer still yields `""`, so the caller's existing `id_str.is_empty()` guard drops the row exactly as before.
- Test covering the number, string, `None`, `null` and object cases.

## Testing

- [x] `cargo test --lib remoteok` — 8 passed, 0 failed, 1 ignored (the `live_search_returns_results` network test).
- [x] `cargo fmt` clean.

The pre-existing tests only used a numeric id (`serde_json::json!(123)`) — the one variant that escapes the bug.

## Checklist

### DIFF
--- a/apps/desktop/src-tauri/src/scraping/boards/remoteok/mod.rs
+++ b/apps/desktop/src-tauri/src/scraping/boards/remoteok/mod.rs
@@ -95,7 +95,7 @@ impl Scraper for RemoteOkScraper {
                     RemoteOkItem::Legend { .. } => continue, // skip legend entry
                 };
 
-            let id_str = id.map(|i| i.to_string()).unwrap_or_else(|| "".to_string());
+            let id_str = normalize_id(id);
             if id_str.is_empty() || position.is_none() {
                 continue;
             }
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

thejesh23:

```
Test Files  369 passed (369)
      Tests  4021 passed (4021)
     Errors  1 error

```

saeedkolivand:
Thanks. Handling both json variants of the remoteok id beats hoping their api stays consistent, which it clearly doesn't. Merged.

saeedkolivand:

PR #772, opened 2026-07-21
fix(applications): make update_fields read-modify-write atomic

### LINKED ISSUE #771: fix(applications): update_fields read-modify-writes across two locks and clobbers concurrent commits
## Describe the bug

`ApplicationStore::update_fields` (`apps/desktop/src-tauri/src/applications/mod.rs:990`) reads the row and writes it back across **two separate lock acquisitions**:

```rust
let existing = self.get(id)?;          // takes the mutex, then RELEASES it
let app = Application { …patched fields…, ..existing };
self.write_row(&app)                   // re-takes the mutex
```

`write_row_conn` is a full-column upsert (`ON CONFLICT DO UPDATE SET` on every column), so it re-persists **all** columns from the snapshot taken before the lock was released. Any writer that commits in that gap is silently overwritten — including columns `update_fields` was never asked to touch.

`upsert_internal` (`:617`) already had this exact race fixed, with the comment *"the old separately-released lookup lock left a gap where a concurrent `merge_answers` commit got clobbered"* and a regression test (`upsert_and_merge_answers_race_never_loses_an_update`). `update_fields` never got the same treatment.

## Steps to reproduce

Two independent IPC commands writing the same Application concurrently — e.g. the UI's `applications_update` and the extension bridge's `answers.save`:

1. UI `applications_update` reads the row: `answers = [A]`.
2. Extension `merge_answers` commits `answers = [A, B]`.
3. UI `update_fields` writes its stale snapshot back: `answers = [A]`.

Answer **B** is gone. The same window reverts a concurrent `set_status` change.

Reproduced deterministically in-repo by porting the sibling race test: two threads, one appending a traceable answer per iteration via `merge_answers`, the other patching `notes` via `update_fields`. Against the current code it fails on **every** run:

```
merge_answers entry 1 was lost to a concurrent update_fields
test result: FAILED. 0 passed; 1 failed
```

## Expected behaviour

A patch of `notes` (or any other field) must not roll back another writer's committed change.

## Actual behaviour

The concurrent commit is clobbered — the answer/status is lost with no error surfaced anywhere.

## Environment

- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready that folds the read + write into one transaction, mirroring `upsert_internal`, and adds the race test above.

thejesh23:
## Summary

`update_fields` read the row under the mutex, released it, then re-took it and re-persisted **every** column from the now-stale snapshot — clobbering any writer that committed in the gap. Fixes #771.

## Type of change

- [x] `fix` — Bug fix

## Changes

- `update_fields` now holds **one** lock across its lookup and its write, using the connection-scoped `row_by_id_conn` / `write_row_conn` instead of the self-locking `get` / `write_row`.
  - Every writer shares a single `Connection` behind this mutex, so not releasing it between the read and the write is what prevents the lost update.
  - `write_row_conn` is a full-column upsert, so the stale snapshot rewrote columns `update_fields` was never asked to touch (`answers`, `status`, `applied_at`, …).
  - `upsert_internal` (`:617`) already carries the same guarantee (it uses a transaction because it makes several writes; this path makes one).
- `write_row` is removed — `update_fields` was its only caller, and leaving it would fail `clippy -D warnings` as dead code.
- New regression test `update_fields_and_merge_answers_race_never_loses_an_answer`, modelled on the existing `upsert_and_merge_answers_race_never_loses_an_update`.

Net effect on `applications/mod.rs` is **-3 LOC** (1395). That matters here: the file is 1398 on `main` against R8's 1400 hard cap, and my first attempt at this fix used a transaction and pushed it to 1407, which CI correctly rejected.

## Testing

- [x] `cargo test --lib` — **2690 passed**, 0 failed

The new test is a genuine reproducer: against `main`'s `update_fields` it fails with

```
merge_answers entry N was lost to a concurrent update_fields
```

Being a thread race it isn't deterministic — measured **2–3 failures per 3 runs** against the old code, and **5/5 passes** with the fix, so it should not be flaky in the passing direction. (An earlier revision of this description said "fails on every run"; that was an overstatement from a smaller sample.)

## Checklist


saeedkolivand:
@claude review

thejesh23:
CI caught two real problems with my first version — both now fixed in `ce34a99`, and both were my fault, not flakes.

**1. `🔬 Quality Checks` — R8 (no oversized modules)**

```
R8 FAILED — module exceeds the hard LOC cap — split it before it grows further
  src/applications/mod.rs:1407  1407 LOC > hard cap 1400
```

`applications/mod.rs` is **1398 LOC on `main`** — two lines under the cap — and my transaction version added 9.

**2. `🔬 Quality Checks` — clippy**

```
error: method `write_row` is never used
```

Replacing the `self.write_row(&app)` call left `write_row` itself dead.

**The fix for both is the same simplification.** `update_fields` does one read and one write, and every writer shares a single `Connection` behind this mutex — so *holding the lock across both* is what actually prevents the lost update. The transaction only added multi-statement atomicity that a lone upsert already has. So:

```rust
let conn = self.conn.lock();
let existing = Self::row_by_id_conn(&conn, id)?
    .ok_or_else(|| AppError::Validation(format!("application not found: {id}")))?;
…
Self::write_row_conn(&conn, &app)
```

and `write_row` is deleted as dead. Net **-3 LOC** on the file (**1395**), so it no longer sits against the cap.

**Re-verified locally:**

- `cargo test --lib` — **2690 passed**, 0 failed
- `cargo test --test architecture` — **11 passed** (R8 included)
- `cargo clippy --all-targets --all-features --locked -- -D warnings` — clean
- `cargo fmt --check` — clean
- the race test: **5/5 passes** with the fix; against `main`'s `update_fields` it fails with `merge_answers entry N was lost to a concurrent update_fields`

**Correction to my PR description:** I wrote that the race test "fails on every run" against the old code. Re-measuring, it's **2–3 failures out of 3 runs** — it is a real reproducer but, being a thread race, not a deterministic one. Sorry for the overstatement; I've left the rest of the description accurate.

**Worth flagging separately:** `applications/mod.rs` sits 2 lines under the R8 hard cap on `main`, so essentially any addition there trips CI. Happy to open an issue (or a PR) splitting the store internals out if that's useful.

saeedkolivand:
Thanks. Read-modify-write races on update_fields were quietly dropping fields, atomic is the only sane version of that path. Merged.

saeedkolivand:

PR #774, opened 2026-07-21
fix(documents): validate an import bundle before clearing the store

### LINKED ISSUE #773: fix(documents): import clears the store before validating rows, so a bad bundle destroys the library
## Describe the bug

`DocumentStore::import` (`apps/desktop/src-tauri/src/documents/mod.rs:1061`) **clears the store first and validates the rows afterwards**:

```rust
fn import(&self, data: &serde_json::Value) -> AppResult<usize> {
    let items = data.as_array().ok_or("documents: expected an array")?;
    self.clear_all();                                   // ← wipes everything
    for item in items {
        let record: DocumentRecord =
            serde_json::from_value(item.clone()).map_err(|e| e.to_string())?;   // ← can fail here
        self.insert(&record)?;
        …
    }
```

`clear_all()` deletes `documents`, `vectors`, `posting_vectors` **and** `match_scores`. There is no enclosing transaction, so if row *k* fails to deserialize — a hand-edited bundle, a bundle from a newer schema, a corrupted file — the function returns `Err` with the originals already gone and rows `0..k` half-inserted.

Every sibling store validates up-front for exactly this reason. `ai_generations::import` even says so:

```rust
// Deserialize EVERY record before mutating the store, so a malformed row
// aborts the import without having cleared the table.
```

`applications`, `dedup` and `discovered` do the same. `documents` is the only outlier — and it is the one holding the résumés and their embeddings.

## Steps to reproduce

1. Have a document library with at least one document (and its embedding).
2. Restore a bundle whose 2nd (or later) document row is malformed.

## Expected behaviour

The import fails and the existing library is untouched.

## Actual behaviour

The import fails **and** the entire prior document library + embeddings + match scores are gone. Verified with a test on current `main`:

```
assertion `left == right` failed: the prior library must survive a failed import
  left: 0
 right: 1
```

## Environment

- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready that deserializes every row (and its optional vector) before calling `clear_all()`, matching the sibling stores, plus a regression test.

thejesh23:
## Summary

`DocumentStore::import` called `clear_all()` **before** deserializing the rows, so a malformed row partway through a restore destroyed the user's entire document library, embeddings and match scores — and still returned `Err`, leaving nothing to restore from. Fixes #773.

## Type of change

- [x] `fix` — Bug fix

## Changes

- `import` now deserializes **every** `DocumentRecord` (and its optional embedding vector) before touching the store, then clears and repopulates — the same shape `ai_generations::import` already uses:
  > *"Deserialize EVERY record before mutating the store, so a malformed row aborts the import without having cleared the table."*

  `applications`, `dedup` and `discovered` all do this too; `documents` was the only outlier, and it is the store holding the résumés and their embeddings.
- The vector/`vectorSpace` reconstruction moved into a documented helper, `parse_exported_vector()`, so the import loop stays readable. Behaviour is unchanged, including the legacy-export fallback to Ollama/`nomic-embed-text`.
- Regression test `import_of_a_malformed_bundle_leaves_the_existing_library_intact`.

**Scope note:** this closes the reported failure — a bad *row* can no longer destroy the library. It does not wrap the clear+insert loop itself in a transaction, so an SQL-level failure mid-loop could still leave a partial state. Doing that properly means connection-scoped variants of `insert`/`set_default` (only `upsert_vector` has one today), which felt like a bigger change than this bug report warrants; happy to follow up if you'd prefer the full transactional version.

## Testing

- [x] `cargo test --lib documents::` — 71 passed, 0 failed (includes the pre-existing `test_data_store_export_import_round_trip`, which covers the happy path and the vector round trip).
- [x] `cargo fmt` clean.

With the fix reverted, the new test fails exactly as reported:

```
assertion `left == right` failed: the prior library must survive a failed import
  left: 0
 right: 1
```

## Checklist

### DIFF
--- a/apps/desktop/src-tauri/src/documents/mod.rs
+++ b/apps/desktop/src-tauri/src/documents/mod.rs
@@ -1031,6 +1031,48 @@ pub fn make_doc_id() -> String {
     format!("doc-{}-{}", now_ms(), &Uuid::new_v4().to_string()[..8])
 }
 
+/// Read an exported document's optional embedding vector out of its JSON row.
+///
+/// Legacy exports carry no `vectorSpace` — they predate cloud embeddings and were
+/// all Ollama/nomic-embed-text, which is what the fallback records.
+fn parse_exported_vector(item: &serde_json::Value) -> Option<EmbeddingVector> {
+    let values: Vec<f64> = item
+        .get("vector")?
+        .as_array()?
+        .iter()
+        .filter_map(|v| v.as_f64())
+        .collect();
+    if values.is_empty() {
+        return None;
+    }
+    let dim = values.len();
+    let space = item
+        .get("vectorSpace")
+        .map(|s| EmbeddingSpace {
+            provider: s
+                .get("provider")
+                .and_then(|v| v.as_str())
+                .unwrap_or("ollama")
+                .to_string(),
+            model: s
+                .get("model")
+                .and_then(|v| v.as_str())
+                .unwrap_or("nomic-embed-text")
+                .to_string(),
+            dim: s
+                .get("dim")
+                .and_then(|v| v.as_u64())
+                .map(|d| d as usize)
+                .unwrap_or(dim),
+        })
+        .unwrap_or_else(|| EmbeddingSpace {
+            provider: "ollama".to_string(),
+            model: "nomic-embed-text".to_string(),
+            dim,
+        });
+    Some(EmbeddingVector { values, space })
+}
+
 impl DataStore for DocumentStore {
     fn key(&self) -> &'static str {
         "documents"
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

thejesh23:
`🧪 Tests` is red here on an unrelated flaky test, not on anything in this PR — this diff is Rust-only (`documents/mod.rs` + its test module), and the failure is in a renderer service test:

```
FAIL |renderer| src/renderer/services/use-extension-bridge/use-extension-bridge.test.ts
  > useSetExtensionAiAssistSetting > caches the returned setting so a re-read reflects it immediately
AssertionError: expected undefined to deeply equal { enabled: true }
  ❯ use-extension-bridge.test.ts:213:71
```

The 1019 ms duration on that one case (its siblings run in 2–54 ms) points at a `waitFor` hitting its 1 s timeout under CI load rather than a logic failure — the assertion sees the cache before it settles.

Rust was clean on this run: `2714 tests run: 2714 passed`.

Could you re-run the job? Happy to look at hardening that test separately if it's recurring for you.

saeedkolivand:
Thanks. Validating the bundle before wiping the store is the difference between a failed import and lost documents, glad this landed. Merged.

saeedkolivand:

PR #776, opened 2026-07-21
fix(postings): write interactions atomically and stop swallowing write errors

### LINKED ISSUE #775: fix(postings): interactions save is non-atomic and discards write errors
## Describe the bug

`InteractionStore::save` (`apps/desktop/src-tauri/src/postings/mod.rs:415`) persists the interactions file like this:

```rust
let records = self.records();
if let Ok(json) = serde_json::to_string_pretty(&records) {
    std::fs::write(&self.data_file, json).ok();
}
```

Two problems:

1. **Non-atomic.** `fs::write` truncates the target and then writes into it. A crash (or a power loss) between the truncate and the final byte leaves `interactions.json` truncated — i.e. corrupt — and the store's own corrupt-file handling then has to kick in on next launch.
2. **The error is discarded.** `.ok()` swallows any write failure — read-only volume, disk full, permissions. `upsert` and `import_bundle` both call `save()` and return normally, so the caller reports success while the record exists only in memory and is silently lost at the next restart.

The serialization failure is swallowed the same way: `if let Ok(json)` quietly skips the write.

## Steps to reproduce

Either of:

- Make the data directory read-only, then mark a job as viewed/applied. The UI shows the interaction; restart the app and it's gone, with nothing logged.
- Kill the process while `save` is mid-write; `interactions.json` is left truncated.

## Expected behaviour

The file is replaced atomically, and a write failure is at minimum logged rather than reported as success.

## Environment

- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready that writes `interactions.json.tmp` and renames it over the target (the same `with_extension` trick the corrupt-file backup path at `:371` already uses), and logs failures instead of discarding them.

`save()` is called from `upsert`/`clear_all`/`import_bundle`, none of which return a `Result`, so the PR logs rather than changing those public signatures — happy to make it propagate instead if you'd prefer that.

thejesh23:
## Summary

`InteractionStore::save` truncated `interactions.json` in place and threw away the write `Result`, so a failed write looked like a success and a crash mid-write left the file truncated. Fixes #775.

## Type of change

- [x] `fix` — Bug fix

## Changes

- `save` now writes `interactions.json.tmp` and `fs::rename`s it over the target, so the file is replaced **atomically** — the same `with_extension` pattern the corrupt-file backup path at `:371` already uses.
- Write, rename and serialization failures are now logged (`log::error!`) instead of being discarded by `.ok()` / `if let Ok(json)`. On any failure the temp file is cleaned up and the previous `interactions.json` is left intact.
- Regression test `save_replaces_the_file_atomically_and_leaves_no_temp_file`.

**On propagating instead of logging:** `save()` is called from `upsert`, `clear_all` and `import_bundle`, none of which return a `Result`, so making it propagate would change three public signatures and their call sites. This PR keeps the change contained and at least makes the failure visible; glad to do the propagating version if you prefer it.

## Testing

- [x] `cargo test --lib postings::` — 31 passed, 0 failed.
- [x] `cargo fmt` clean.

The test asserts the file is written and parses completely, that no `.tmp` file is left behind, and that a freshly-opened store hydrates both records from it.

## Checklist

### DIFF
--- a/apps/desktop/src-tauri/src/postings/mod.rs
+++ b/apps/desktop/src-tauri/src/postings/mod.rs
@@ -411,8 +411,39 @@ impl InteractionStore {
             return;
         }
         let records = self.records();
-        if let Ok(json) = serde_json::to_string_pretty(&records) {
-            std::fs::write(&self.data_file, json).ok();
+        let json = match serde_json::to_string_pretty(&records) {
+            Ok(json) => json,
+            Err(e) => {
+                log::error!(
+                    "[postings] save skipped: could not serialize {} interaction(s): {e}",
+                    records.len()
+                );
+                return;
+            }
+        };
+        // Write-then-rename, so the file is replaced atomically. `fs::write`
+        // truncates the target in place, so a crash mid-write left
+        // `interactions.json` truncated — and the discarded `Result` meant a
+        // failed write (disk full, read-only volume) still looked like a success
+        // to `upsert`/`import_bundle`, with the record living only in memory and
+        // silently lost on the next restart.
+        let tmp = self.data_file.with_extension("json.tmp");
+        if let Err(e) = std::fs::write(&tmp, &json) {
+            log::error!(
+                "[postings] failed to write {}: {e} — interaction NOT persisted",
+                tmp.display()
+            );
+            std::fs::remove_file(&tmp).ok();
+            return;
+        }
+        if let Err(e) = std::fs::rename(&tmp, &self.data_file) {
+            log::error!(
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

thejesh23:

```
Test Files  369 passed (369)
      Tests  4021 passed (4021)
     Errors  1 error

```

saeedkolivand:
Thanks. Swallowed write errors on interactions meant we could lose state and never know, atomic writes plus surfaced errors closes that hole. Merged.

saeedkolivand:

PR #778, opened 2026-07-21
fix(autopilot): honour a cancel on the scrape error path

### LINKED ISSUE #777: fix(autopilot): a run stopped during the scrape is recorded Failed and auto-retried
## Describe the bug

Stopping an autopilot run during the scrape records it as **Failed** and then auto-retries it ~12 minutes later.

`autopilot_run` (`apps/desktop/src-tauri/src/commands/autopilot.rs:257`) handles the scrape result like this:

```rust
let (postings, summaries) = match autopilot_scrape(&engine, &target, &job_id, &app).await {
    Ok(out) => out,
    Err(e) => {
        engine.unregister_token(&job_id).await;
        store(&app).lock().fail_run_without_summaries(&autopilot_id);
        crate::commands::jobs::job_fail(&app, &job_id, e.to_string());
        span.end(false);
        return json!({ "error": e, "jobId": job_id });   // ← no cancel check
    }
};
```

But an early cancel *is* an `Err`. `ScraperEngine` (`scraping/engine/mod.rs:936`) reports it that way deliberately:

```rust
// F6 — return Err only when the user cancelled AND no items were recovered.
if parent.is_cancelled() && !any_recovered_items {
    return Err(anyhow::anyhow!("scrape cancelled"));
}
```

The only cancel handler in the command (`:418`, `Completed` + `job_cancel` + `{cancelled:true}`) sits on the **Ok** path and is never reached in this case. Two consequences:

1. The run is persisted **Failed** and `job_fail` fires, so the UI shows a failure for something the user deliberately stopped.
2. The returned payload is `{ "error": … }`, and `outcome_failed` (`autopilot_scheduler.rs:307`) returns `true` for any payload carrying `error` — so `run_with_single_retry` re-runs the scrape the user just stopped.

`outcome_failed` explicitly does **not** want this — its own comment says *"A user stop or a de-duplicated double-invoke is not a failure"* and it returns `false` for `{cancelled: true}`. The `Err` arm just never produces that shape.

## Steps to reproduce

1. Start a scheduled autopilot run.
2. Press Stop before any job has streamed in (so `any_recovered_items` is still false).

## Expected behaviour

Run recorded as cancelled/Completed, `job_cancel` fired, no retry.

## Actual behaviour

Run recorded **Failed**, `job_fail` fired, and the scheduler retries the stopped scrape after the backoff.

## Environment

- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready that checks `cancel_token.is_cancelled()` first in the `Err` arm and mirrors the existing Ok-path cancel handler.

thejesh23:
## Summary

The scrape `Err` arm of `autopilot_run` had no cancel check, so a run the user stopped mid-scrape was persisted **Failed** and then auto-retried. Fixes #777.

## Type of change

- [x] `fix` — Bug fix

## Changes

- The `Err` arm now checks `cancel_token.is_cancelled()` **first** and mirrors the existing Ok-path cancel handler (`:418`): `set_run_status_clearing_summaries(… Completed)`, `job_cancel`, and a `{ "cancelled": true }` payload.
- Why the `Err` arm needs it: an early Stop *is* an `Err`, by design —
  ```rust
  // F6 — return Err only when the user cancelled AND no items were recovered.
  if parent.is_cancelled() && !any_recovered_items {
      return Err(anyhow::anyhow!("scrape cancelled"));
  }
  ```
  (`scraping/engine/mod.rs:936`) — while the only cancel handler lived on the Ok path.
- The `{ "error": … }` payload it used to return makes `outcome_failed` (`autopilot_scheduler.rs:307`) return `true`, so `run_with_single_retry` re-ran the scrape the user had just stopped. `outcome_failed` already returns `false` for `{cancelled: true}` — *"A user stop or a de-duplicated double-invoke is not a failure"* — the `Err` arm simply never produced that shape.

## Testing

- [x] `cargo test --lib autopilot` — 150 passed, 0 failed (includes `outcome_failed_only_true_for_failed_runs`, which already pins `{cancelled:true}` → no retry).
- [x] `cargo fmt` clean.
- [ ] Added tests for new logic — see below.

**No new unit test**, and I'd rather say so than pretend: `autopilot_run` is a `#[tauri::command]` that needs a live `AppHandle`, the `ScraperEngine` state and the store, so there is no existing harness to drive it from a test — none of the command's other branches are unit-tested either. The behaviour it now produces (`{cancelled: true}` → not a failure → no retry) *is* covered on the scheduler side by `outcome_failed_only_true_for_failed_runs`. If you'd like a test here, I'm happy to add one behind whatever seam you'd prefer.

## Checklist


saeedkolivand:
@claude review

saeedkolivand:
Thanks. A user hitting stop and then watching the run get retried as failed was a bad look, checking cancel before the failure path is the right ordering. The duplicated cancel handler the review flagged goes on the cleanup list for after this batch lands. Merged.

saeedkolivand:

PR #780, opened 2026-07-21
fix(spend): recognise a port-less ipv6 loopback as local

### LINKED ISSUE #779: fix(spend): port-less IPv6 loopback is not recognised as local, so a free call is billed
## Describe the bug

`is_localhost_url` (`apps/desktop/src-tauri/src/spend/mod.rs:324`) strips the port by splitting on the **last** colon:

```rust
let host = host_and_port
    .rsplit_once(':')
    .map_or(host_and_port, |(h, _)| h);
matches!(host, "localhost" | "127.0.0.1" | "0.0.0.0" | "::1" | "[::1]")
```

An IPv6 host contains colons of its own, so when there is **no** port the last colon is an internal one:

- `http://[::1]/v1` → `host_and_port = "[::1]"` → `rsplit_once(':')` → host `"[:"` → matches nothing.
- the bare `"::1"` arm is dead for the same reason (`"::1".rsplit_once(':')` → `":"`).

Only the ported form `http://[::1]:1234/v1` works — which is exactly the one the existing test covers.

Because `is_free_call` is then false, a **local, free** call to an `openai-compatible` server is billed. And since a local model name is not in `RATES`, `estimate_cost` falls back to `DEFAULT_RATE` ($3/$15 per 1M), so the user sees a fabricated dollar figure for a call that cost nothing.

## Steps to reproduce

1. Point the `openai-compatible` provider at a port-less IPv6 loopback base URL, e.g. `http://[::1]/v1` (LM Studio / llama.cpp / vLLM behind a proxy on the default port).
2. Run any AI action.
3. Check the spend view.

## Expected behaviour

Cost `$0.00` — same as `http://[::1]:1234/v1`, `http://localhost:1234/v1` or `http://127.0.0.1:8080`.

## Actual behaviour

The call is treated as paid and billed at the `DEFAULT_RATE` fallback.

Confirmed against current `main` by adding the port-less form to the existing test:

```
thread 'spend::test::is_localhost_url_recognizes_common_local_forms' panicked at src/spend/test.rs:107:9:
http://[::1]/v1 should be recognized as local
```

## Environment

- **Ollama version**: n/a (local OpenAI-compatible server)
- **Model in use**: any local model

## Additional context

A PR is ready that strips the port structurally — everything up to and including `]` for a bracketed host, whole for a bare IPv6 — and extends the existing test with the port-less forms.

thejesh23:
## Summary

`is_localhost_url` stripped the port by splitting on the **last** colon, which mangles an IPv6 host that has no port — so `http://[::1]/v1` was treated as a remote paid endpoint and billed at the `DEFAULT_RATE` fallback. Fixes #779.

## Type of change

- [x] `fix` — Bug fix

## Changes

- The port is now stripped structurally:
  - bracketed IPv6 → the host is everything up to and including `]`;
  - bare IPv6 (more than one colon, no bracket) → kept whole, since a port cannot be appended to it unambiguously;
  - otherwise → split on the single colon as before.
- This makes both previously-dead match arms reachable: `http://[::1]/v1` (and `[::1]`, `::1`) now resolve to a loopback host. The ported form `http://[::1]:1234/v1` and every IPv4/hostname form behave exactly as before.
- Extended `is_localhost_url_recognizes_common_local_forms` with the port-less spellings, plus an `is_free_call` case.

Impact: a local (free) `openai-compatible` call was billed, and because a local model name isn't in `RATES`, `estimate_cost` used the `DEFAULT_RATE` $3/$15-per-1M fallback — a fabricated figure for a call that cost nothing.

## Testing

- [x] `cargo test --lib spend::` — 19 passed, 0 failed.
- [x] `cargo fmt` clean.

With the fix reverted, the extended test fails on exactly the reported input:

```
thread 'spend::test::is_localhost_url_recognizes_common_local_forms' panicked at src/spend/test.rs:107:9:
http://[::1]/v1 should be recognized as local
```

The pre-existing negative cases (`https://openrouter.ai/api/v1`, `http://my-localhost-lookalike.example.com/v1`) still pass, so nothing remote became free.

## Checklist


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Billing a fabricated rate for a server running on the same machine was a silly bug, the bracket-first parsing is much closer to how the url actually reads. The rarer loopback spellings the review mentioned go on the post-batch list. Merged.

saeedkolivand:

PR #782, opened 2026-07-21
fix(applications): reject a bad-typed next_action_at instead of clearing it

### LINKED ISSUE #781: fix(applications): a bad-typed nextActionAt silently clears the reminder instead of erroring
## Describe the bug

`applications_update` (`apps/desktop/src-tauri/src/commands/applications.rs:135`) maps the inbound `nextActionAt` like this:

```rust
let next_action_at: Option<Option<u64>> = req.next_action_at.map(|v| match v {
    serde_json::Value::Null => None,
    other => other.as_u64(),
});
```

The store's patch shape is `Option<Option<u64>>`, where the inner `None` means **"clear the reminder"**. `Value::as_u64()` also returns `None` for anything that isn't a `u64` — a negative number, a fractional number, an oversized one, a string, an object.

So a caller passing a bad-typed number produces exactly the same `Some(None)` as an explicit `null`: instead of setting (or rejecting) the reminder, the handler **silently clears** it. No error is returned and nothing is logged.

Every other bad input on this same handler *is* rejected — `validate_recipient_email` returns `AppError::Validation`, and `reject_oversized_job_description` rejects an oversized JD explicitly. `nextActionAt` is the odd one out.

The schema doesn't catch it either: `packages/shared/src/schemas/index.ts:390` is `z.number().int().nullable().optional()` with no `.min(0)` — and a direct IPC caller bypasses the renderer's Zod validation anyway, which is precisely why the other fields are re-validated server-side.

## Steps to reproduce

Invoke the `applications_update` IPC command with `nextActionAt: -1` (or `1.5`, or `"1767225600000"`) on an application that already has a reminder set.

## Expected behaviour

A `Validation` error — or at minimum, the reminder left alone.

## Actual behaviour

The existing reminder is cleared and the call reports success.

## Environment

- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready that pulls the mapping into a pure `parse_next_action_at` helper (matching `validate_recipient_email`'s shape in the same file), rejects non-`u64` values with `AppError::Validation`, and unit-tests it.

The PR is Rust-only; adding `.min(0)` to the Zod schema would also be reasonable, but that touches the IPC codegen so I left it for you to decide.

thejesh23:
## Summary

`applications_update` mapped `nextActionAt` through `Value::as_u64()`, whose `None` is indistinguishable from an explicit `null` — so a bad-typed value silently **cleared** the reminder instead of setting or rejecting it. Fixes #781.

## Type of change

- [x] `fix` — Bug fix

## Changes

- New pure helper `parse_next_action_at()`, shaped like the existing `validate_recipient_email()` in the same file:
  - absent → `Ok(None)` (leave unchanged),
  - explicit `null` → `Ok(Some(None))` (clear),
  - a `u64` → `Ok(Some(Some(ms)))` (set),
  - anything else → `Err(AppError::Validation)`.
- `applications_update` returns that error through the same `span.end_with` + `{ "error": e }` path it already uses for `validate_recipient_email`.
- Two unit tests covering the accepted mappings (including `0`, a legitimate epoch value) and the rejected ones (`-1`, `1.5`, oversized, string, object, array).

**Scope note:** Rust-only. `packages/shared/src/schemas/index.ts:390` is `z.number().int().nullable().optional()` with no `.min(0)`, so tightening the schema is also reasonable — but that runs through the IPC codegen, so I left that call to you. The server-side guard is the real trust boundary either way, since a direct IPC caller bypasses the renderer's Zod validation (the same reasoning the JD and recipient-email guards already follow).

## Testing

- [x] `cargo test --lib commands::applications` — 17 passed, 0 failed.
- [x] `cargo fmt` clean.

## Checklist


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Silently clearing a reminder because the payload had the wrong type was a nasty one, rejecting loudly is much better behaviour. The zod min(0) tightening from the review is on the post-batch list. Merged.

saeedkolivand:

PR #786, opened 2026-07-21
fix(extension): only auto-track a submit that came from the application form

### LINKED ISSUE #785: fix(extension): auto-track marks a job applied on any form submit or a bare Apply click
## Describe the bug

The auto-track submit watcher (`apps/extension/src/lib/submit-watch.ts`) fires on signals that don't mean "the application was sent", and the background then auto-advances a `saved` application to `applied` (`background.ts:531`/`:1027`).

Two independent triggers:

**1. Any form submit on the page.**

```ts
// Real form submit — any submit path that fires the native event.
doc.addEventListener('submit', () => fire(), true);
```

This is a capture-phase listener on the whole document, so it fires for **every** `<form>`: the site search box, a filter, a newsletter signup, a login form. The watcher then posts nothing but `location.href`, so the background cannot tell those apart from the real thing.

**2. A bare "Apply" control that only OPENS the form.**

```ts
const APPLY_TEXT_RE = /apply|submit application|send application|finish/i;
```

`apply` is unanchored and matches "Apply", "Apply now", "Easy Apply" — on a job-listing page those are almost always the button that opens the application (often into a modal), not the one that sends it. `isApplyControl` accepts any `role="button"` too, so an SPA "Apply now" link qualifies.

## Steps to reproduce

Either of:

1. Import a job from the extension (→ `saved`, which arms the watcher — opening the popup auto-runs `answersSuggest`, a gesture kind, so it arms readily), then press Enter in the page's own search box. The application flips to `applied`.
2. Import a job, then click a `role="button"` labelled "Apply now" that opens the application modal. The application flips to `applied` before anything has been filled in.

## Expected behaviour

`applied` is only set when the application was actually submitted.

## Actual behaviour

The application is marked `applied` on an unrelated form submit, or on merely opening the application form. The user's tracker now says they applied to a job they didn't.

## Environment

- **OS**: macOS 15 (platform-independent — pure DOM logic)
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

The module's own doc frames auto-track as observe-only and careful, but nothing here distinguishes a real submit from any other one. A PR is ready that:

- gates the `submit` listener on the target form actually looking like an application form (≥3 visible fillable fields, or a file input);
- keeps the broad text pattern for a control **inside** such a form (where the structure corroborates it), and requires an explicit send verb ("submit/send application", "finish") for a control with no form around it.

That keeps Easy-Apply/SPA flows working while dropping both false positives.

thejesh23:
## Summary

The submit watcher fired on **any** form submit and on a bare "Apply" control, so a search-box submit or merely opening the application modal auto-marked the job `applied`. Fixes #785.

## Type of change

- [x] `fix` — Bug fix

## Changes

- **Submit path is scoped to the application form.** `looksLikeApplicationForm()` requires a visible file input (the résumé upload — decisive on its own) or ≥3 visible fillable fields. A site search box, a newsletter signup and a login form all fall below that; a real application form does not.
- **Click path is corroborated by structure.** A control **inside** an application-looking form keeps the existing broad `APPLY_TEXT_RE`. A control with **no** form around it (pure SPA / `role="button"`) now needs `STRICT_APPLY_TEXT_RE` — an explicit send verb (`submit/send [your|my|the] application`, `finish`) — because out there a bare "Apply now" is the button that *opens* the application.
- Documented the direction as an invariant in the module header: this signal auto-advances a `saved` application to `applied`, so a false positive silently misreports what the user did, while a false negative just leaves them to mark it by hand.

**Known remaining gap (pre-existing, not introduced here):** a *"Save draft"* `type="submit"` button inside the real application form implicitly submits that form, so the submit listener still fires for it. Distinguishing that needs the submitter's identity (`SubmitEvent.submitter`) rather than the form, which felt like a separate change — happy to do it here if you'd like.

## Testing

- [x] `pnpm exec vitest run src/lib/submit-watch.test.ts src/lib/auto-track.test.ts src/lib/autofill.test.ts` — 97 passed, 0 failed.

Two new tests, both of which fail against the current code:

```
× does NOT post on a search / newsletter form submit
× does NOT post on a bare "Apply now" that is not inside an application form
```

**Existing tests I had to adjust, and why** — all of them were fixtures that no longer describe a real application page, not changes of intent:

| Test | Change |
| --- | --- |
| `posts once on a real form submit` | form had no fields at all; added the résumé/name/email fields |
| `OBSERVES ONLY — never preventDefault` | same fixture fix |
| `posts on a click of an apply-style submit button` | `<button>Apply now</button>` had no form; wrapped it in the application form (renamed to say so) |
| `resolves the control when the click lands on a child` | same fixture fix |
| `does NOT fire on … "Save draft"` | kept it **formless** so it still exercises the click heuristic alone (inside a form, a submit button implicitly submits — see the gap above) |
| `does NOT fire on a hidden apply button` | text changed to "Submit application" so `isHidden` is genuinely the only thing keeping it quiet |

Unchanged and still passing: the `role="button"` Easy-Apply case, the `input[type=submit] value="Finish"` case, and the fire-once guard.

## Checklist

- [x] New public APIs are documented (both new helpers are private, documented inline)

### DIFF
--- a/apps/extension/src/lib/submit-watch.test.ts
+++ b/apps/extension/src/lib/submit-watch.test.ts
@@ -18,13 +18,23 @@ function setBody(html: string): void {
   document.body.innerHTML = html;
 }
 
+/** The fields that make a `<form>` read as a real application form rather than
+ *  a search box / newsletter signup (see `looksLikeApplicationForm`). */
+const APPLICATION_FIELDS = `
+  <input name="first_name" />
+  <input name="email" type="email" />
+  <input name="resume" type="file" />
+`;
+
 afterEach(() => {
   document.body.innerHTML = '';
 });
 
 describe('armSubmitWatch — real form submit', () => {
   it('posts once on a real form submit', () => {
-    setBody(`<form id="f"><button type="submit">Submit application</button></form>`);
+    setBody(
+      `<form id="f">${APPLICATION_FIELDS}<button type="submit">Submit application</button></form>`
+    );
     const post = vi.fn();
     armSubmitWatch(document, post);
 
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Marking a job applied because some unrelated form fired a submit was quietly eroding trust in the tracker, and under-report-over-over-report is the right invariant. The hidden file input case from the review goes on the follow-up list. Merged.

saeedkolivand:

PR #788, opened 2026-07-21
fix(extension): reply via the response callback so popup requests work on chrome

### LINKED ISSUE #787: fix(extension): background onMessage returns a Promise, so every popup action fails on Chrome
## Describe the bug

On Chrome, every popup request/response action fails with **"No response from the extension background."**

`apps/extension/src/background.ts:1031` replies to a popup request by **returning a Promise** from the `onMessage` listener:

```ts
browser.runtime.onMessage.addListener(
  (message: unknown, sender: Browser.runtime.MessageSender): Promise<PopupResponse> | undefined => {
    if (isSubmitDetected(message)) { …; return undefined; }
    return handleRequest(message as PopupRequest);
  }
);
```

That is the **Firefox** `browser.*` convention. Chromium's `chrome.runtime.onMessage` has never supported it — the listener must call `sendResponse` and return the **literal** `true` to keep the channel open. A returned Promise is truthy but not `true`, so Chrome closes the channel immediately with no reply.

`@wxt-dev/browser` does not bridge that difference. Its entire runtime is:

```js
// node_modules/@wxt-dev/browser/src/index.mjs
export const browser = globalThis.browser?.runtime?.id ? globalThis.browser : globalThis.chrome;
```

No wrapper, no `webextension-polyfill` dependency (`package.json` deps are `@types/filesystem` and `@types/har-format` only). On Chromium `globalThis.browser` is undefined, so `browser === chrome` and the raw Chrome semantics apply.

Downstream, `popup.ts`'s `send` sees `undefined` and produces the error message:

```ts
const res = (await browser.runtime.sendMessage(req)) as PopupResponse | undefined;
if (!res) return { ok: false, error: 'No response from the extension background.' };
```

## Steps to reproduce

1. Load the extension in Chrome.
2. Open the popup on any job page and press any button — Import, Fill, Save answers, Mark applied, Check fit.

## Expected behaviour

The action completes and its result is rendered.

## Actual behaviour

Every one of them returns "No response from the extension background."

Note the connection pill can still read **Connected**: one-way `background → popup` pushes (`browser.runtime.sendMessage` from the background) are unaffected, so only the request/response path is broken. Firefox is unaffected.

## Environment

- **OS**: any
- **Browser**: Chrome / any Chromium (Firefox is fine)
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready. `sendResponse` + `return true` is the shape **both** engines accept, so the fix is correct on Firefox too.

thejesh23:
## Summary

The background `onMessage` listener replied by returning a Promise, which Chromium does not support — so every popup request/response action returned "No response from the extension background." Fixes #787.

## Type of change

- [x] `fix` — Bug fix

## Changes

- The listener now takes `sendResponse`, calls it with the result, and returns the **literal** `true`.
- A rejection from `handleRequest` also calls `sendResponse` (with `{ok:false, error}`) instead of leaving the port open until it times out — which the popup would surface as the same "No response" message.
- `background.test.ts`'s `send()` helper now drives the listener the way the browser does (passing a `sendResponse` callback) and **asserts the listener returned literal `true`**, so this can't regress silently.

Why this is a Chrome-only failure, and why the package doesn't save us — `@wxt-dev/browser`'s entire runtime is a pass-through:

```js
// node_modules/@wxt-dev/browser/src/index.mjs
export const browser = globalThis.browser?.runtime?.id ? globalThis.browser : globalThis.chrome;
```

No wrapper and no `webextension-polyfill` dependency, so on Chromium `browser === chrome` and a Promise return (truthy, but not `true`) closes the channel immediately. Firefox's native `browser.*` accepts the Promise form, which is why this never showed up there.

`sendResponse` + `return true` is accepted by **both** engines, so this is safe on Firefox as well.

## Testing

- [x] `pnpm exec vitest run src/background.test.ts` — 65 passed, 0 failed.
- [x] Added tests for new logic (the harness assertion).

With the fix reverted, the new harness assertion catches it across the board:

```
Error: listener must return literal true to keep the channel open, got [object Promise]
⎯ Failed Tests 62 ⎯
```

The three that still passed are the ones that don't go through the request/response path — which is exactly the asymmetry that made this hard to notice in the first place.

## Checklist

### DIFF
--- a/apps/extension/src/background.test.ts
+++ b/apps/extension/src/background.test.ts
@@ -97,15 +97,38 @@ const setBadgeTextMock = vi.mocked(browser.action.setBadgeText);
 const listener = vi.mocked(browser.runtime.onMessage.addListener).mock.calls[0]?.[0] as
   | ((
       message: unknown,
-      sender: Browser.runtime.MessageSender
-    ) => Promise<PopupResponse> | undefined)
+      sender: Browser.runtime.MessageSender,
+      sendResponse: (response?: PopupResponse) => void
+    ) => true | undefined)
   | undefined;
 
+/**
+ * Drive the registered listener the way the browser does: hand it a
+ * `sendResponse` callback and resolve on the reply.
+ *
+ * The `kept === true` assertion is the point — Chromium keeps the message
+ * channel open for an async reply ONLY for a literal `true` return. Anything
+ * else (including a returned Promise, which is truthy but not `true`) closes it
+ * immediately and `sendMessage` resolves `undefined`, which the popup reports as
+ * "No response from the extension background."
+ */
 function send(req: PopupRequest): Promise<PopupResponse> {
   if (!listener) throw new Error('onMessage listener not registered');
-  return listener(req, {
-    id: EXTENSION_ID,
-  } as Browser.runtime.MessageSender) as Promise<PopupResponse>;
+  return new Promise<PopupResponse>((resolve, reject) => {
+    const kept = listener(
+      req,
+      { id: EXTENSION_ID } as Browser.runtime.MessageSender,
+      (response?: PopupResponse) => {
+        if (response) resolve(response);
+        else reject(new Error('listener called sendResponse with no response'));
+      }
+    );
+    if (kept !== true) {
+      reject(
+        new Error(`listener must return literal true to keep the channel open, got ${String(kept)}`)
+      );
+    }
+  });
 }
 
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Chrome never delivering the promise-style reply to the popup explains a bunch of weird one-off reports, the callback path is the boring correct fix. Merged.

saeedkolivand:

PR #790, opened 2026-07-21
fix(shared): map the iso codes franc actually emits for zh, ar and no

### LINKED ISSUE #789: fix(shared): Chinese/Arabic/Norwegian always detect as unknown (dead franc map keys)
## Describe the bug

Chinese, Arabic and Norwegian always detect as `'unknown'`.

`packages/shared/src/language-detection.ts:19,22,28` keys the franc → ISO 639-1 map on the **macrolanguage** codes:

```ts
const LANGUAGE_MAP: Record<string, string> = {
  …
  zho: 'zh',
  ara: 'ar',
  nor: 'no',
  …
};
```

franc (v6) never emits those. It returns the **individual** ISO 639-3 code — `cmn` for Mandarin, `arb` for Standard Arabic, `nob`/`nno` for Norwegian Bokmål/Nynorsk. So those three entries are dead and `detectLanguage` falls through to `?? 'unknown'`.

Verified against the installed franc 6.2.0:

```
Chinese    -> cmn
Arabic     -> arb
Norwegian  -> nob
German     -> deu   ← the mapped codes that do work
```

Downstream consequences:

- `detectLanguages(...).mismatch` can never flag a Chinese/Arabic résumé against an English ad (the mismatch check requires both sides known), so the "résumé and job ad are in different languages" warning never appears for those users.
- Referral drafts collapse `'unknown'` → `'en'`, so a Chinese or Arabic user silently gets an English draft.
- `zh` is a first-class `LocaleSchema` output and `CJK_LANGUAGES` lists it, so the rest of the codebase expects this to work.

This is the same class of bug as the `swe → sv` fix in commit #679; these three were just missed.

## Steps to reproduce

```ts
import { detectLanguage } from '@ajh/shared';
detectLanguage('我们正在寻找一位经验丰富的软件工程师加入我们的团队。…'); // → 'unknown'
```

## Expected behaviour

`'zh'` / `'ar'` / `'no'`.

## Actual behaviour

`'unknown'` for all three.

## Environment

- **OS**: any
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

The existing tests only cover `en`/`de`/`fr` — all three of which map from codes franc really does emit — so the dead keys went unnoticed. A PR is ready with the corrected keys (`cmn`, `arb`, `nob`, plus `nno`) and tests using real text in each script.

thejesh23:
## Summary

`LANGUAGE_MAP` keyed Chinese/Arabic/Norwegian on the **macrolanguage** ISO 639-3 codes (`zho`/`ara`/`nor`), which franc never emits — so all three always resolved to `'unknown'`. Fixes #789.

## Type of change

- [x] `fix` — Bug fix

## Changes

- `zho → cmn`, `ara → arb`, `nor → nob`, plus `nno` (Nynorsk) also mapping to `no`.
- Added a doc comment on the map recording *why* — franc emits the individual code, never the macrolanguage one — so the next entry doesn't repeat it.
- Test with real Chinese, Arabic and Norwegian text.

Verified against the installed franc 6.2.0:

```
Chinese    -> cmn
Arabic     -> arb
Norwegian  -> nob
German     -> deu     ← one of the codes that already mapped correctly
```

Same class as the `swe → sv` fix in commit #679; these three were just missed. The existing tests only covered `en`/`de`/`fr`, all of which happen to use codes franc really does emit, so nothing caught it.

Downstream effect of the fix: `detectLanguages(...).mismatch` can now flag a Chinese/Arabic résumé against an English ad (it requires both sides known), and referral drafts stop collapsing `'unknown'` → `'en'` for those users.

## Testing

- [x] `pnpm exec vitest run src/language-detection.test.ts` (in `packages/shared`) — 13 passed, 0 failed.
- [x] `pnpm --filter @ajh/shared typecheck` — clean.
- [x] `pnpm exec prettier --check` — clean.

With the fix reverted:

```
× detects Chinese, Arabic and Norwegian (franc emits cmn/arb/nob)
  → expected 'unknown' to be 'zh'
```

## Checklist

### DIFF
--- a/packages/shared/src/language-detection.test.ts
+++ b/packages/shared/src/language-detection.test.ts
@@ -14,6 +14,15 @@ const GERMAN =
   'Erfahrener Softwareentwickler mit fundierten Kenntnissen in der Entwicklung skalierbarer Webanwendungen und verteilter Backend-Systeme für große Unternehmen.';
 const FRENCH =
   "Ingénieur logiciel expérimenté possédant une solide expérience dans la création d'applications web évolutives et de systèmes backend distribués pour de grandes organisations.";
+// franc emits the INDIVIDUAL ISO 639-3 code for these — cmn / arb / nob — never
+// the macrolanguage zho / ara / nor, so mapping on the macrolanguage code left
+// the entry dead and every one of these resolved to 'unknown'.
+const CHINESE =
+  '我们正在寻找一位经验丰富的软件工程师加入我们的团队。您将负责设计和开发高质量的后端服务，与产品经理和设计师紧密合作，推动项目按时交付。';
+const ARABIC =
+  'نبحث عن مهندس برمجيات ذي خبرة للانضمام إلى فريقنا. ستكون مسؤولاً عن تصميم وتطوير خدمات خلفية عالية الجودة والعمل بشكل وثيق مع مديري المنتجات والمصممين لتسليم المشاريع في الوقت المحدد.';
+const NORWEGIAN =
+  'Vi ser etter en erfaren programvareutvikler som vil bli med i teamet vårt. Du vil ha ansvaret for å designe og utvikle backend-tjenester av høy kvalitet, og samarbeide tett med produktsjefer og designere.';
 
 describe('detectLanguage', () => {
   it('returns "unknown" for empty or very short text', () => {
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Mapping keys franc never actually emits meant chinese, arabic and norwegian all silently fell through to english. Matching the codes it really returns fixes it at the root. Merged.

saeedkolivand:

PR #792, opened 2026-07-21
fix(prompts): stop the italic strip from eating bold markup

### LINKED ISSUE #791: fix(prompts): extractPlainText downgrades and merges **bold** keyword markup
## Describe the bug

`extractPlainText` (`packages/prompts/src/generate/text/text.ts:17`) corrupts `**bold**` markup in every generated document.

```ts
.replace(/\*\*\*(.+?)\*\*\*/g, '**$1**') // triple → double (preserve bold)
.replace(/\*([^*]+)\*/g, '$1')           // single italic → plain
```

The italic pass is not bold-aware, and it does two different kinds of damage:

1. **Downgrade.** In `**bold**` the leftmost match is the *inner* pair `*bold*`, so the result is `*bold*` — the bold is silently demoted to italic markup.
2. **Merge.** `[^*]+` matches spaces and commas, so two adjacent bold spans pair up **across** each other and swallow the text between them.

Reproduced in `node` with the exact pipeline:

```
"**bold**"                                  → "*bold*"
"Skills: **Python**, **Go**, **Kubernetes**" → "Skills: *Python, Go, Kubernetes*"
```

That second one is the damaging case: the commas and words between the keywords end up *inside* a single emphasis span.

This matters because the prompts explicitly instruct the model to emit 2–3 `**keyword**` bolds per bullet, and `extractPlainText` runs on every generated résumé, cover letter, email and answer — so it fires on essentially every document the app produces.

## Steps to reproduce

```ts
import { extractPlainText } from '@ajh/prompts';
extractPlainText('Skills: **Python**, **Go**, **Kubernetes**');
// → 'Skills: *Python, Go, Kubernetes*'
```

## Expected behaviour

`**…**` runs are preserved (the triple→double pass says so in its own comment: *"preserve bold"*); only genuine single-`*` italics are stripped.

## Actual behaviour

Bold is downgraded, and adjacent bold spans merge into one, absorbing the text between them.

## Environment

- **OS**: any
- **Ollama version**: n/a (the bug is in post-processing, after generation)
- **Model in use**: any

## Additional context

The existing test enshrines the single-span downgrade (`expect(extractPlainText('***strong***')).toBe('*strong*')` with the comment *"The single-italic pass also unwraps the inner pair of a bold run"*), so the downgrade looks intentional there — but the multi-keyword merge is unambiguously wrong and untested.

A PR is ready that guards the italic pass against `*` adjacent to another `*`, updates that one assertion, and adds coverage for the merge case.

thejesh23:
## Summary

The italic-strip pass in `extractPlainText` was not bold-aware, so it downgraded `**bold**` to `*bold*` and — worse — merged adjacent bold spans, swallowing the text between them. Fixes #791.

## Type of change

- [x] `fix` — Bug fix

## Changes

- The italic strip now matches only a `*` that is **not** part of a `**` run:
  ```diff
  -.replace(/\*([^*]+)\*/g, '$1')
  +.replace(/(?<!\*)\*(?!\*)([^*]+)\*(?!\*)/g, '$1')
  ```
- Before → after, run through the real pipeline:

  | input | before | after |
  | --- | --- | --- |
  | `**bold**` | `*bold*` | `**bold**` |
  | `Skills: **Python**, **Go**, **Kubernetes**` | `Skills: *Python, Go, Kubernetes*` | unchanged |
  | `an *italic* word` | `an italic word` | `an italic word` |
  | `***strong***` | `*strong*` | `**strong**` |
  | `**a** and *b*` | `*a and b*` | `**a** and b` |

- Tests: added coverage for the merge case and for bold+italic on one line.

**One existing assertion changed**, deliberately:

```diff
-// The single-italic pass also unwraps the inner pair of a bold run.
-expect(extractPlainText('***strong***')).toBe('*strong*');
+expect(extractPlainText('***strong***')).toBe('**strong**');
```

That assertion documented the downgrade rather than intending it — the pass immediately above it exists to *"preserve bold"*, which `*strong*` doesn't. The multi-keyword merge on the same code path is unambiguously wrong and was untested.

This runs on every generated résumé, cover letter, email and answer, and the prompts ask the model for 2–3 `**keyword**` bolds per bullet, so it fired on essentially every document.

## Testing

- [x] `pnpm exec vitest run` in `packages/prompts` — 516 passed, 0 failed (15 files).
- [x] `pnpm --filter @ajh/prompts typecheck` — clean.
- [x] `pnpm exec prettier --check` — clean.

With the fix reverted:

```
× reduces emphasis markers … → expected '*strong*' to be '**strong**'
× leaves **bold** keyword markup intact → expected '*bold*' to be '**bold**'
```

## Checklist

### DIFF
--- a/packages/prompts/src/generate/generate.test.ts
+++ b/packages/prompts/src/generate/generate.test.ts
@@ -1298,10 +1298,23 @@ describe('extractPlainText', () => {
     expect(out).toContain('body');
   });
 
-  it('reduces emphasis markers (triple/single asterisks collapse)', () => {
-    // The single-italic pass also unwraps the inner pair of a bold run.
-    expect(extractPlainText('***strong***')).toBe('*strong*');
+  it('reduces emphasis markers (triple collapses to bold, single italic is stripped)', () => {
+    expect(extractPlainText('***strong***')).toBe('**strong**');
     expect(extractPlainText('an *italic* word')).toBe('an italic word');
+  });
+
+  it('leaves **bold** keyword markup intact', () => {
+    // The italic pass used to match the INNER pair of a bold run, downgrading
+    // `**bold**` to `*bold*`; and since `[^*]+` matches spaces and commas, two
+    // adjacent bold spans paired up across each other and swallowed the text
+    // between them. The prompts ask for 2-3 `**keyword**` bolds per bullet, so
+    // this corrupted essentially every generated document.
+    expect(extractPlainText('**bold**')).toBe('**bold**');
+    expect(extractPlainText('Skills: **Python**, **Go**, **Kubernetes**')).toBe(
+      'Skills: **Python**, **Go**, **Kubernetes**'
+    );
+    // Bold and italic in the same line: only the italic is stripped.
+    expect(extractPlainText('**a** and *b*')).toBe('**a** and b');
   });
 
   it('removes a fenced code block entirely (no orphaned backticks or code leak)', () => {
--- a/packages/prompts/src/generate/text/text.ts
+++ b/packages/prompts/src/generate/text/text.ts
@@ -14,7 +14,14 @@ export function extractPlainText(raw: string): string {
       .replace(/<\/?leakage_check>/gi, '') // stray unclosed tags
       .replace(/^#{1,6}\s/gm, '')
       .replace(/\*\*\*(.+?)\*\*\*/g, '**$1**') // triple → double (preserve bold)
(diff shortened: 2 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Bold text losing its markup to a greedy italic strip was subtle, and the fix keeps the regex readable. The pre-existing bullet list edge is noted for later. Merged.

saeedkolivand:

PR #794, opened 2026-07-21
fix(autopilot): derive step patches from the reducer state, not a snapshot

### LINKED ISSUE #793: fix(autopilot): batched step events drop a log entry (patches computed from a stale snapshot)
## Describe the bug

`handleStep` in `apps/desktop/src/renderer/features/autopilot/hooks/useAutopilotRun.ts:36` builds its reducer patches from the **render snapshot** it closed over, not from the reducer's own `prev`:

```ts
const handleStep = useCallback(
  (event: AutopilotStepEvent) => {
    const ev = stepToEvent(event.step);
    if (ev) {
      setRunStates({
        [event.autopilotId]: machineTransition(
          autopilotRunMachine,
          runStates[event.autopilotId] ?? 'idle',   // ← snapshot
          ev
        ),
      });
    }
    setStepLogs({
      [event.autopilotId]: [
        ...(stepLogs[event.autopilotId] ?? []).slice(-49),  // ← snapshot
        { step: event.step, detail: event.detail, ts: Date.now() },
      ],
    });
  },
  [runStates, stepLogs]
);
```

Both reducers are pure `{ ...prev, ...patch }` merges, so the patch value is computed *before* dispatch and is whatever the last render produced. When two step events for the same `autopilotId` are processed in one React batch — no re-render in between — both patches start from the same base, and the second one overwrites the first.

Result: a dropped step-log line, and a skipped state transition (the machine jumps straight to the second event's target instead of walking through both).

## Steps to reproduce

Deliver two step events for the same autopilot in one batch:

```ts
act(() => {
  handleStep({ jobId: 'j1', autopilotId: 'ap', step: 'scrape_start', detail: '' });
  handleStep({ jobId: 'j1', autopilotId: 'ap', step: 'scrape_done',  detail: '' });
});
```

## Expected behaviour

`stepLogs.ap` contains both `scrape_start` and `scrape_done`.

## Actual behaviour

```
AssertionError: expected [ 'scrape_done' ] to deeply equal [ 'scrape_start', 'scrape_done' ]
```

The `scrape_start` entry is gone, and `runStates.ap` reflects only the second transition.

## Environment

- **OS**: any
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

In practice Tauri events usually arrive far enough apart to re-render in between, which narrows the window — but a burst (or React 18's automatic batching under a transition) hits it, and the casualty is silent: a missing line in the run's step log.

A PR is ready that lets the reducers take a functional patch and computes both from `prev`.

thejesh23:
## Summary

`handleStep` computed its reducer patches from the render snapshot rather than the reducer's `prev`, so two step events for one autopilot in the same React batch lost the first one. Fixes #793.

## Type of change

- [x] `fix` — Bug fix

## Changes

- Both reducers now accept a **functional** patch (`Patch<T> = Partial<T> | ((prev: T) => Partial<T>)`), and `handleStep` derives `runStates` / `stepLogs` from `prev`.
- `handleStep`'s `useCallback` deps drop to `[]` — it no longer closes over any state, so the handler identity is stable too.
- Every other call site keeps passing a plain object patch, unchanged (`setRunStates({ [id]: 'scraping' })` etc.); the functional form is only needed where the new value derives from the old one.
- Regression test `keeps both step-log entries when two events for one autopilot land in the same batch`.

## Testing

- [x] `pnpm exec vitest run src/renderer/features/autopilot/hooks/useAutopilotRun.test.ts` — 9 passed, 0 failed (8 pre-existing + 1 new).
- [x] `pnpm exec eslint` / `prettier` on the changed files — clean.

With the fix reverted:

```
× keeps both step-log entries when two events for one autopilot land in the same batch
AssertionError: expected [ 'scrape_done' ] to deeply equal [ 'scrape_start', 'scrape_done' ]
```

**Note on `pnpm typecheck`:** I could not run it — on a fresh clone `apps/desktop` fails to typecheck because `routeTree.gen.ts` doesn't exist until the router plugin has run (`Cannot find module '@/routeTree.gen'`, plus the ~15 route errors that follow from it). That is pre-existing on `main` and unrelated to this change; the test suite runs fine since vitest doesn't typecheck.

## Checklist

- [x] New public APIs are documented (the `Patch<T>` helper type is documented inline)

### DIFF
--- a/apps/desktop/src/renderer/features/autopilot/hooks/useAutopilotRun.test.ts
+++ b/apps/desktop/src/renderer/features/autopilot/hooks/useAutopilotRun.test.ts
@@ -21,6 +21,13 @@ vi.mock('@ajh/translations', () => ({
 
 const runMutateAsync = vi.fn();
 
+/** Captures the step handler the hook registers, so a test can deliver step
+ *  events the way the Tauri event bridge does. */
+const stepEvents = vi.hoisted(() => ({
+  handler: null as
+    ((event: { jobId: string; autopilotId: string; step: string; detail: string }) => void) | null,
+}));
+
 vi.mock('@/services', async (importActual) => {
   const actual = await importActual<Record<string, unknown>>();
   return {
@@ -29,7 +36,9 @@ vi.mock('@/services', async (importActual) => {
     usePauseAutopilot: () => ({ mutateAsync: vi.fn().mockResolvedValue(undefined) }),
     useResumeAutopilot: () => ({ mutateAsync: vi.fn().mockResolvedValue(undefined) }),
     useRemoveAutopilot: () => ({ mutateAsync: vi.fn().mockResolvedValue(undefined) }),
-    useAutopilotStepEvents: () => {},
+    useAutopilotStepEvents: (cb: (event: never) => void) => {
+      stepEvents.handler = cb as unknown as typeof stepEvents.handler;
+    },
   };
 });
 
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

thejesh23:

```
Test Files  369 passed (369)
      Tests  4021 passed (4021)
     Errors  1 error

```

saeedkolivand:
Thanks. Patching steps off a snapshot taken at closure creation is a classic stale-state trap, deriving from the reducer kills the whole class of bug. Merged with the tests check still red, so that failure is on the post-batch list.

saeedkolivand:

PR #796, opened 2026-07-21
fix(jobs): mark a posting applied only after tailor succeeds

### LINKED ISSUE #795: fix(jobs): a failed Tailor still marks the posting Applied (optimistic, never reverted)
## Describe the bug

A **failed** "Prep/Tailor" still marks the posting as Applied.

`handleTailor` (`apps/desktop/src/renderer/features/jobs/hooks/usePostingActions.ts:72`) fires the interaction **before** the save and never rolls it back:

```ts
const handleTailor = async () => {
  void trackInteraction('applied');          // ← optimistic, and persisted
  let res;
  try {
    res = await saveFromPostingMutation.mutateAsync({ … });
  } catch {
    notify.error({ message: t('jobs.tailorError') });
    return;                                   // ← no revert
  }
  if (!res?.id) {
    notify.error({ message: t('jobs.tailorError') });
    return;                                   // ← no revert
  }
  …
};
```

`trackInteraction` is not just local state — it awaits `persistJobMutation.mutateAsync({ job, interactionType })`, so the `applied` interaction is written to the backend. Both failure paths `return` without undoing it.

The sibling `handleSave` in the same file gets this right:

```ts
.then(() => {
  // Only mark saved and notify AFTER the mutation resolves successfully.
  void trackInteraction('bookmarked');
  …
})
```

## Steps to reproduce

1. On the Jobs page, click **Tailor** on a posting while the save backend errors (or returns a payload with no `id`).
2. An error toast appears — and the posting now shows the **Applied** badge, which survives a restart.

## Expected behaviour

A failed Tailor leaves the posting untouched.

## Actual behaviour

The posting is recorded as Applied for a job the user never applied to, contradicting the error toast shown at the same moment.

## Environment

- **OS**: any
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready that moves `trackInteraction('applied')` after the `res?.id` guard, matching `handleSave`, plus tests for both failure paths.

thejesh23:
## Summary

`handleTailor` fired `trackInteraction('applied')` before the save and never reverted it on failure, so a failed Tailor left the posting reading **Applied**. Fixes #795.

## Type of change

- [x] `fix` — Bug fix

## Changes

- `trackInteraction('applied')` moved below the `res?.id` guard, so it runs only after the save actually succeeded — matching the sibling `handleSave`, which already marks `bookmarked` only in its `.then()`.
- Two tests: `saveFromPosting` rejecting, and resolving without an `id`.

The optimistic call was not just local state: `trackInteraction` awaits `persistJobMutation.mutateAsync({ job, interactionType })`, so the interaction was written to the backend and survived a restart, while both failure paths `return`ed without undoing it — the user saw an error toast and an Applied badge at the same time.

## Testing

- [x] `pnpm exec vitest run src/renderer/features/jobs/hooks/usePostingActions.test.ts` — 28 passed, 0 failed (26 pre-existing + 2 new).

With the fix reverted, both new tests fail:

```
× does NOT mark applied when saveFromPosting rejects
× does NOT mark applied when saveFromPosting resolves without an id
AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
```

One existing test was renamed — `tracks applied interaction (optimistic set) before awaiting the mutation` → `tracks the applied interaction once the save has succeeded`. Its assertion is unchanged and still passes; only the title described the old ordering.

**Note on `pnpm typecheck`:** not runnable here — on a fresh clone `apps/desktop` fails typecheck because `routeTree.gen.ts` doesn't exist until the router plugin has run. Pre-existing on `main` and unrelated to this change.

## Checklist

### DIFF
--- a/apps/desktop/src/renderer/features/jobs/hooks/usePostingActions.test.ts
+++ b/apps/desktop/src/renderer/features/jobs/hooks/usePostingActions.test.ts
@@ -327,7 +327,7 @@ describe('usePostingActions — handleView', () => {
 // The two score-presence variants below override via a fresh mock per test.
 
 describe('usePostingActions — handleTailor', () => {
-  it('tracks applied interaction (optimistic set) before awaiting the mutation', async () => {
+  it('tracks the applied interaction once the save has succeeded', async () => {
     const { result } = renderHook(() => usePostingActions(makePosting()));
     await act(async () => {
       await result.current.handleTailor();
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. An applied flag that survives a failed tailor was lying to the user, moving it after the save succeeds is the right ordering. Merged.

saeedkolivand:

PR #798, opened 2026-07-21
fix(monitoring): dedup the activity feed on the real job id

### LINKED ISSUE #797: fix(monitoring): activity feed lists every completed job twice
## Describe the bug

Every completed job appears **twice** in the monitoring Activity feed, permanently.

`useActivityFeed` merges live events on top of the persisted job records and dedupes by job id (`apps/desktop/src/renderer/features/monitoring/hooks/useActivityFeed.ts:56`):

```ts
const liveJobIds = new Set(liveActivity.map((a) => a.id.split('-')[0]));
return [...liveActivity, ...historicalActivity.filter((a) => !liveJobIds.has(a.id))].slice(0, 40);
```

The ids don't line up with that split:

- a job id is itself hyphenated — `format!("job-{}", Uuid::new_v4())` (`db.rs:144`), e.g. `job-3f2504e0-4f89-11d3-…`;
- the live item's id is `` `${event.jobId}-${event.ts}` `` (`:92`), e.g. `job-3f2504e0-…-1700000001500`;
- the historical item's id is the raw job id (`:46`).

So `a.id.split('-')[0]` is **always** the literal string `"job"`. `liveJobIds` is the one-element set `{"job"}`, no historical id ever equals `"job"`, the filter removes nothing, and both rows render. They carry different React keys, so nothing else catches it.

## Steps to reproduce

1. Open Monitoring.
2. Run any job (a scrape, an AI generation) and let it finish.
3. The Activity feed shows the same completed job as two rows, and it stays that way.

## Expected behaviour

One row per job.

## Actual behaviour

Two rows per completed/failed/cancelled job.

## Environment

- **OS**: any
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready that strips only the trailing `-<ts>` (`id.slice(0, id.lastIndexOf('-'))`) to recover the real job id, plus tests for the dedup and for the non-duplicate case.

thejesh23:
## Summary

The Activity feed's live/historical dedup keyed on `a.id.split('-')[0]`, which is always the literal `"job"` — so nothing was ever deduped and every completed job rendered twice. Fixes #797.

## Type of change

- [x] `fix` — Bug fix

## Changes

- Recover the real job id by stripping only the trailing `-<ts>`:
  ```diff
  -const liveJobIds = new Set(liveActivity.map((a) => a.id.split('-')[0]));
  +const liveJobIds = new Set(
  +  liveActivity.map((a) => {
  +    const cut = a.id.lastIndexOf('-');
  +    return cut > 0 ? a.id.slice(0, cut) : a.id;
  +  })
  +);
  ```
  A job id is itself hyphenated (`format!("job-{}", Uuid::new_v4())`, `db.rs:144`) and the live item's id is `` `${event.jobId}-${event.ts}` ``, so splitting on the first `-` yielded `"job"` for every entry — a one-element set no historical id could match.
- New `useActivityFeed.test.ts` covering both directions: a job with a live counterpart is listed once, and a job without one is still kept.

## Testing

- [x] `pnpm exec vitest run src/renderer/features/monitoring/hooks/useActivityFeed.test.ts` — 2 passed, 0 failed.
- [x] `pnpm exec eslint` / `prettier` on the changed files — clean.

With the fix reverted, both fail:

```
× lists a just-completed job once, not twice
   AssertionError: expected [ … ] to have a length of 1 but got 2
× keeps a historical job that has no live counterpart
   AssertionError: expected [ … ] to have a length of 2 but got 3
```

**Note on `pnpm typecheck`:** not runnable here — on a fresh clone `apps/desktop` fails typecheck because `routeTree.gen.ts` doesn't exist until the router plugin has run. Pre-existing on `main` and unrelated to this change.

## Checklist

### DIFF
--- a/apps/desktop/src/renderer/features/monitoring/hooks/useActivityFeed.test.ts
+++ b/apps/desktop/src/renderer/features/monitoring/hooks/useActivityFeed.test.ts
@@ -1,0 +1,92 @@
+/**
+ * useActivityFeed — live/historical merge.
+ *
+ * The feed shows live job events on top and the persisted job records below,
+ * deduped so a job that just finished is not listed twice. The dedup keys the
+ * live item's id back to the job id it was built from.
+ */
+
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { act, renderHook, waitFor } from '@testing-library/react';
+
+// ── services ─────────────────────────────────────────────────────────────────
+
+const mockFetchJob = vi.fn();
+/** The job-event handler the hook registers, so a test can deliver events. */
+const jobEvents = vi.hoisted(() => ({
+  handler: null as ((event: unknown) => void) | null,
+}));
+
+vi.mock('@/services', () => ({
+  fetchJob: (...args: unknown[]) => mockFetchJob(...args),
+  useJobEvents: (cb: (event: unknown) => void) => {
+    jobEvents.handler = cb;
+  },
+}));
+
+import type { JobRecord } from '@/features/monitoring/types';
+
+import { useActivityFeed } from './useActivityFeed';
+
+// ── fixtures ─────────────────────────────────────────────────────────────────
+
+/** Real shape: `format!("job-{}", Uuid::new_v4())` on the Rust side. */
+const JOB_ID = 'job-3f2504e0-4f89-11d3-9a0c-0305e82c3301';
+
+function completedJob(overrides: Partial<JobRecord> = {}): JobRecord {
+  return {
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Deduping on the composite id was never going to hold once timestamps differed, keying on the real job id is the fix. Merged.

saeedkolivand:

PR #802, opened 2026-07-21
fix(generate): prefer the persisted result over a truncated stream buffer

### LINKED ISSUE #801: fix(generate): poll fallback resolves a truncated stream buffer over the persisted result
## Describe the bug

`awaitAiStream`'s job-status poll can resolve a **truncated** document as if it were the finished one.

`apps/desktop/src/renderer/lib/generate/stream-promise.ts:178`:

```ts
} else if (job.status === 'completed') {
  off();
  splitter.flush();
  cleanup();
  resolve(buffer || job.result?.text || '');
}
```

The poll exists *because* streamed deltas can be lost — that is its entire reason for being. But it consults the persisted `job.result.text` only when `buffer` is **empty**. If an *interior* delta was dropped (rather than just the final `done`), `buffer` is a truncated-but-truthy prefix, so `buffer || …` short-circuits and the incomplete text is resolved as the complete document — and then persisted by the caller.

The timeout path in the same file gets the principle right and says so explicitly:

```ts
// On timeout we FAIL the run (never persist truncated output as success)
```

## Steps to reproduce

1. Start a generation.
2. Drop a chunk mid-stream (a dropped `ai.onStream` event), so no `done` chunk ever arrives.
3. The poll ticks, sees `completed`, and resolves the partial buffer.

Reproduced with a fake client driving the real `awaitAiStream`:

```
AssertionError: expected 'Dear hiring manager, ' to be 'Dear hiring manager, I am writing to …'
```

## Expected behaviour

The authoritative persisted result wins over a partial stream.

## Actual behaviour

The truncated prefix is resolved and saved as the finished document.

## Environment

- **OS**: any
- **Ollama version**: n/a
- **Model in use**: any

## Additional context

A PR is ready that resolves whichever of the two is longer, so a partial stream can never beat the persisted result while a missing `result.text` still falls back to what did stream.

thejesh23:
## Summary

The job-status poll resolved `buffer || job.result?.text`, so a stream that lost an *interior* delta produced a truncated-but-truthy prefix that won over the authoritative persisted result. Fixes #801.

## Type of change

- [x] `fix` — Bug fix

## Changes

```diff
-resolve(buffer || job.result?.text || '');
+const persisted = job.result?.text ?? '';
+resolve(persisted.length > buffer.length ? persisted : buffer);
```

- The poll exists *because* deltas can be dropped, so the buffer is precisely the value that can't be trusted to be complete — yet `||` only consulted the persisted text when the buffer was empty (i.e. only when the *final* `done` was lost, not an interior chunk).
- Taking the longer of the two means a partial stream can never beat the persisted result, and a missing/empty `result.text` still falls back to whatever did stream.
- This matches the principle the timeout path in the same file already states: *"On timeout we FAIL the run (never persist truncated output as success)"*.

## Testing

- [x] New `stream-promise.test.ts` — 3 passed, 0 failed: truncated buffer vs. full persisted text; buffer wins when there is no persisted text; and the `done`-chunk path (which never polls) still resolves the buffer directly.

With the fix reverted:

```
× resolves the persisted result when the streamed buffer is a truncated prefix
AssertionError: expected 'Dear hiring manager, ' to be 'Dear hiring manager, I am writing to …'
```

**Two pre-existing environment issues, for transparency** — both reproduce on unmodified `main`, neither is related to this change:
- `src/renderer/lib/generate/generation/generation.test.ts` fails with `TypeError: Cannot read properties of undefined (reading 'setItem')` from zustand's persist middleware — Node 24+ has no global `localStorage` without `--localstorage-file`.
- `apps/desktop` can't be typechecked on a fresh clone (`Cannot find module '@/routeTree.gen'`) until the router plugin has generated the route tree.

## Checklist


saeedkolivand:
@claude review

saeedkolivand:
Thanks. The diagnosis of the interior-drop case is right and the ?? over || is correct. Worth being honest that the review found the backend never populates result.text on this path, so the runtime effect is nil until the persistence side lands, that's now tracked as a post-batch follow-up along with making the test use the real job-result shape. Merged.

saeedkolivand:

PR #784, opened 2026-07-21
fix(extension): anchor the city and phone identity keywords

thejesh23:
## Summary

`city` and `phone`/`mobile` were unanchored substrings in `NAMED_KEY_PATTERNS`, so `"ethnicity"` resolved to `location` and `"smartphone"` to `phone` — writing the user's city into an EEO field and their phone number into a "Smartphone model" field. Fixes #783.

## Type of change

- [x] `fix` — Bug fix

## Changes

- Anchored `city` / `town` / `phone` / `mobile` on their **leading** side with `(?:^|[^a-z])`, matching the intent already expressed by the neighbouring `\bhandy\b` / `\bmobil\b` / `\bort\b` / `\bville\b` anchors and their comment (*"bare `mobil` ⊂ 'automobil…'"*).
- Legitimate compounds are spelled out so they keep matching: `cell|home|work|day|tele` + `phone`, and `hometown`.

**Why not `\b`:** `_` is a regex word character, so `\bphone\b` would stop matching `phone_number`, `candidate_phone`, `mobile_number` and `city_1` — all extremely common ATS field names. Leaving the **trailing** side open also keeps the camelCase-flattened signals working (`textSignal` joins `name`/`id`/`placeholder`/`aria-label`/label without splitting camelCase, so `name="phoneNumber"` arrives as `phonenumber`).

**Why `(?:^|[^a-z])` and not a lookbehind:** these patterns are only ever `.test()`ed (`matchNamedKey`), so consuming the boundary character is harmless, and this form has no engine-support caveats.

Incidental improvements from the same anchor: `"iphone"`, `"microphone"` and `"headphone"` no longer resolve to `phone` either.

## Testing

- [x] New `apps/extension/src/lib/field-signal.test.ts` — covers both directions: 17 phone signals + 15 location signals that must still match, and the words that must not.
- [x] `pnpm exec vitest run src/lib/field-signal.test.ts src/lib/autofill.test.ts src/lib/answers-capture.test.ts` — 128 passed, 0 failed. (`autofill` and `answers-capture` are the two consumers of these patterns.)
- [x] `pnpm --filter @ajh/extension typecheck` — clean.
- [x] `pnpm exec eslint` on the changed files — clean.
- [x] `pnpm exec prettier --check` — clean.
- [x] Added tests for new logic.

With the fix reverted, the new tests fail on exactly the reported inputs:

```
AssertionError: smartphone: expected 'phone' not to be 'phone'
[...]

PR #805, opened 2026-07-21
fix(extension): anchor the collision-prone denylist terms

### LINKED ISSUE #804: fix(extension): unanchored denylist terms skip Preferred/Research/Preferences fields entirely
## Describe the bug

Three entries in the `AMBIGUOUS` denylist (`apps/extension/src/lib/field-signal.ts:35,37,49`) are matched as **plain substrings** via `signal.includes(w)`, so they fire inside longer, perfectly ordinary words:

| denylist term | matches inside | fields it kills |
| --- | --- | --- |
| `referr` | "p**referr**ed" | "Preferred first name", "Preferred name", "Preferred pronouns" |
| `search` | "re**search**" | "Research experience", "Research interests" |
| `reference` | "p**reference**s" | "Work preferences", "Notification preferences" |

Verified against the current list:

```
"preferred first name"     => BLOCKED by [ 'referr' ]
"preferred pronouns"       => BLOCKED by [ 'referr' ]
"research experience"      => BLOCKED by [ 'search' ]
"work preferences"         => BLOCKED by [ 'reference' ]
"notification preferences" => BLOCKED by [ 'reference' ]
```

`isAmbiguousSignal` gates **both** consumers — autofill's `isCandidateField` and answers-capture's `isCapturable` — so an affected field is neither filled **nor captured**. That's worse than a plain under-fill: the user types the answer and it is never learned, so the next application asks again.

"Preferred first name" in particular is close to universal on ATS forms (Greenhouse, Lever, Workday all ship it).

## Steps to reproduce

1. Open an application form containing a "Preferred first name" or "Research experience" field.
2. Run autofill — the field is skipped.
3. Fill it by hand and save answers — it is not captured either.

## Expected behaviour

Only genuine referral/reference/search fields are skipped. "Preferred first name" fills as a first name; "Research experience" is capturable.

## Actual behaviour

All of them are treated as ambiguous and silently skipped by both paths.

## Environment

- **OS**: any (pure string matching)
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

Same class as #783 (the `city`/`phone` identity keywords), which you merged as #784 — this is the denylist half of it.

Note the anchor can't be `\b`: `_` is a regex word character, so `\breferr` / `\bsearch` would stop matching `job_search` and `reference_name`, and the trailing side must stay open for `referral` / `referrer` / `references` / `searchTerm`. A PR is ready using the same leading-anchor form as #784.

thejesh23:
## Summary

`referr`, `reference` and `search` were plain-substring denylist entries, so "Preferred first name", "Research experience" and "Work preferences" were treated as ambiguous — neither filled nor captured. Fixes #804.

## Type of change

- [x] `fix` — Bug fix

## Changes

- Moved the three collision-prone terms out of `AMBIGUOUS` into a new leading-anchored `AMBIGUOUS_PREFIXED = /(?:^|[^a-z])(?:referr|reference|search)/`, and folded it into `isAmbiguousSignal`.
- `referral` is dropped from the list as redundant — the `referr` alternative already covers it.

**Why leading-anchored only**, same two reasons as the identity patterns in #784:
- `_` is a regex word character, so `\breferr` / `\bsearch` would stop matching `job_search`, `reference_name`, `job_referral`;
- the trailing side must stay open for `referral`, `referrer`, `references`, `searchTerm`.

**Why this one bites harder than a normal under-fill:** `isAmbiguousSignal` gates *both* consumers — autofill's `isCandidateField` and answers-capture's `isCapturable` — so the field is skipped **and** the user's typed answer is never learned. "Preferred first name" is close to universal on ATS forms.

I also checked the rest of the list for the same class and deliberately left it alone: `company` ⊂ "accompany", `parent` ⊂ "apparent", `tax` ⊂ "syntax"/"taxonomy" are all real substring hits, but none is a plausible job-application field label, so anchoring them would be churn without benefit. Happy to include them if you'd rather have the list uniformly anchored.

## Testing

- [x] `pnpm exec vitest run src/lib/` — **272 passed**, 0 failed (10 files, incl. `autofill` and `answers-capture`, the two consumers).
- [x] Added tests for new logic — both directions: 16 signals that must still be skipped (including the `\b`-anchored `dni` / `contact d'urgence`, unchanged) and 7 that must not be.

With the fix reverted:

```
× does not skip a field that merely CONTAINS a denylist term
AssertionError: preferred first name: expected true to be false
```

## Checklist

- [x] New public APIs are documented (the new const is private, documented inline)

### DIFF
--- a/apps/extension/src/lib/field-signal.test.ts
+++ b/apps/extension/src/lib/field-signal.test.ts
@@ -12,7 +12,54 @@
 
 import { describe, expect, it } from 'vitest';
 
-import { matchNamedKey } from './field-signal';
+import { isAmbiguousSignal, matchNamedKey } from './field-signal';
+
+describe('isAmbiguousSignal', () => {
+  it('still skips the genuinely ambiguous / sensitive fields', () => {
+    for (const signal of [
+      'referral source',
+      'referrer',
+      'job_referral',
+      'professional references',
+      'reference_name',
+      'site search',
+      'job_search',
+      'emergency contact',
+      'confirm password',
+      'company name',
+      'recruiter',
+      'ssn',
+      'passport number',
+      // \b-anchored short terms, unchanged
+      'dni',
+      "contact d'urgence",
+    ]) {
+      expect(isAmbiguousSignal(signal), signal).toBe(true);
+    }
+  });
+
+  it('does not skip a field that merely CONTAINS a denylist term', () => {
+    // `referr` ⊂ "preferred", `search` ⊂ "research", `reference` ⊂ "preferences".
+    // These were skipped entirely — never filled AND never captured.
+    for (const signal of [
+      'preferred first name',
+      'preferred name',
+      'preferred pronouns',
+      'research experience',
+      'research interests',
+      'work preferences',
+      'notification preferences',
+    ]) {
+      expect(isAmbiguousSignal(signal), signal).toBe(false);
+    }
+  });
+
+  it('lets a freed-up field resolve to its real key', () => {
+    // "Preferred first name" is ubiquitous on ATS forms; once it is no longer
+    // treated as ambiguous it fills as a first name.
+    expect(matchNamedKey('preferred first name')).toBe('firstName');
+  });
+});
 
 describe('matchNamedKey — phone', () => {
   it('matches real phone fields, including the separator-less compounds', () => {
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Anchoring the collision-prone denylist terms without churning the safe ones is the right balance, and the revert-fails check gives it teeth. The one-word doc drift the review caught goes on the follow-up list. Merged.

saeedkolivand:

PR #807, opened 2026-07-21
fix(monitoring): gate the running pulse on a stable id, not a translated label

### LINKED ISSUE #806: fix(monitoring): Running card never pulses outside English (compares a translated label)
## Describe the bug

The "Running" metric card never pulses in any non-English locale.

`apps/desktop/src/renderer/features/monitoring/components/MonitoringPage/index.tsx:105` gates the animation on the **translated** label:

```tsx
{metrics.map(({ label, value, icon: Icon, color, bg }) => (
  <MetricCard
    …
    animate={label === 'Running' && counters.running > 0}
  />
))}
```

but `label` comes from `t('monitoring.metrics.running')` (`:50`), so the comparison only holds when the active translation happens to be the literal string `"Running"`:

| locale | `monitoring.metrics.running` | `label === 'Running'` |
| --- | --- | --- |
| en | `Running` | ✅ |
| de | `Läuft` | ❌ |

So on a German (or any other translated) UI the card is static even while jobs are running — the one bit of feedback that says work is in flight.

`key={label}` has the same smell: it keys a list on translated text rather than on a stable id.

## Steps to reproduce

1. Switch the app language to German.
2. Start any job so `counters.running > 0`.
3. The Running card does not animate. In English it does.

## Expected behaviour

The pulse depends on *which* metric it is, not on how that metric's name is spelled in the current language.

## Environment

- **OS**: any
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready giving each metric a stable `id` (`completed`/`running`/`failed`/`successRate`) and gating on `id === 'running'` (and keying the list on it).

thejesh23:
## Summary

The Running metric's pulse was gated on `label === 'Running'`, where `label` is the **translated** string — so it only ever animated in English. Fixes #806.

## Type of change

- [x] `fix` — Bug fix

## Changes

- Each metric now carries a stable `id` (`completed` / `running` / `failed` / `successRate`).
- `animate` gates on `id === 'running'`, and the list `key` moved from the translated `label` to `id` — same underlying problem, keying a list on translated text.

Concretely, `monitoring.metrics.running` is `"Running"` in `en` but `"Läuft"` in `de`, so the comparison was false for every non-English UI while jobs were actually running.

## Testing

- [x] `pnpm exec vitest run src/renderer/features/monitoring/` — 8 passed, 0 failed.
- [x] `pnpm exec eslint` / `prettier --check` on the changed file — clean.
- [ ] Added tests for new logic — see below.

**No new test, deliberately.** `CONTRIBUTING.md` § Testing lists UI components under *"Skip"*, and `vitest.config.ts` excludes `renderer/features/**` from coverage as *"presentational composition … covered by the Playwright E2E suite"*. Driving `MonitoringPage` in a unit test would mean standing up the router, query client, i18n and four service hooks to assert one boolean prop. Happy to add it if you'd rather have the coverage.

**Note on `pnpm typecheck`:** not runnable here — on a fresh clone `apps/desktop` fails typecheck because `routeTree.gen.ts` doesn't exist until the router plugin has run. Pre-existing on `main`, unrelated to this change.

## Checklist

### DIFF
--- a/apps/desktop/src/renderer/features/monitoring/components/MonitoringPage/index.tsx
+++ b/apps/desktop/src/renderer/features/monitoring/components/MonitoringPage/index.tsx
@@ -39,8 +39,11 @@ export function MonitoringPage() {
 
   const last24h = useMemo(() => binLast24Hours(allJobs, Date.now()), [allJobs]);
 
+  // `id` is the stable identity: `label` is translated, so branching on it (as
+  // the pulse below used to) only ever holds in English.
   const metrics = [
     {
+      id: 'completed',
       label: t('monitoring.metrics.completed'),
       value: counters.completed,
       icon: CheckCircle2,
@@ -48,6 +51,7 @@ export function MonitoringPage() {
       bg: 'bg-emerald-500/10',
     },
     {
+      id: 'running',
       label: t('monitoring.metrics.running'),
       value: counters.running,
       icon: Loader2,
@@ -55,6 +59,7 @@ export function MonitoringPage() {
       bg: 'bg-blue-500/10',
     },
     {
+      id: 'failed',
       label: t('monitoring.metrics.failed'),
       value: counters.failed,
       icon: XCircle,
@@ -62,6 +67,7 @@ export function MonitoringPage() {
       bg: 'bg-amber-500/10',
     },
     {
+      id: 'successRate',
       label: t('monitoring.metrics.successRate'),
       value: `${successRate}%`,
       icon: Zap,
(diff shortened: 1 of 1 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Keying the pulse on a translated label meant the animation quietly died in german, gating it on the stable id is the fix at the right layer. Merged.

saeedkolivand:

PR #809, opened 2026-07-21
fix(prompts): only flag a language mismatch when both languages are known

### LINKED ISSUE #808: fix(prompts): validateMetadata flags a language mismatch on unknown/blank languages
## Describe the bug

`validateMetadata` (`packages/prompts/src/generate/metadata/metadata.ts:59`) reports a language mismatch for inputs where the language simply isn't known:

```ts
mismatch: (parsed.resumeLanguage ?? 'en') !== (parsed.jobAdLanguage ?? 'en'),
```

Two problems:

1. **No `'unknown'` guard.** `{resumeLanguage:'en', jobAdLanguage:'unknown'}` → `'en' !== 'unknown'` → `mismatch: true`.
2. **`??` is nullish-only.** The model routinely answers with an empty string rather than omitting the key, and `'' ?? 'en'` is `''` — so `{resumeLanguage:''}` → `'' !== 'en'` → `mismatch: true`.

Both sibling implementations already get this right and require **both** sides to be known:

```ts
// packages/prompts/src/analyze/validate.ts:126
mismatch:
  langs.mismatch === true ||
  (resumeLang !== 'unknown' && jobAdLang !== 'unknown' && resumeLang !== jobAdLang),
```
```ts
// packages/shared/src/language-detection.ts:147
mismatch: resumeLang !== 'unknown' && jobAdLang !== 'unknown' && resumeLang !== jobAdLang,
```

`validateMetadata` was never brought along. A false `mismatch` puts a "the résumé and the ad are in different languages — rewrite entirely / do not translate" instruction into the generation prompt, so the model is steered by a premise that isn't true.

## Steps to reproduce

```ts
validateMetadata('{"resumeLanguage":"en","jobAdLanguage":"unknown"}').mismatch  // → true
validateMetadata('{"resumeLanguage":"","jobAdLanguage":"en"}').mismatch        // → true
```

## Expected behaviour

`false` in both cases — nothing is known to differ.

## Actual behaviour

`true`, and the downstream prompt carries a spurious language-mismatch instruction.

## Environment

- **OS**: any
- **Ollama version**: n/a (post-processing of the model's metadata answer)
- **Model in use**: any

## Additional context

A PR is ready that coerces the two languages first (so `''` falls back to `'en'` like a missing key already does) and then applies the same both-sides-known rule as `analyze/validate.ts`.

thejesh23:
## Summary

`validateMetadata` reported a language mismatch when a language was `'unknown'` or an empty string, putting a spurious "different languages — rewrite entirely" instruction into the generation prompt. Fixes #808.

## Type of change

- [x] `fix` — Bug fix

## Changes

- New `toLanguage()` coercion: a non-blank string wins, anything else falls back to `'en'` — so an empty string now behaves like a missing key already did (`'' ?? 'en'` is `''`, which is what let it through).
- `mismatch` now requires **both** sides known and different, matching `analyze/validate.ts:126` and `@ajh/shared`'s `detectLanguages:147`. `validateMetadata` was the only one of the three without the guard.
- `targetLanguage` now uses the coerced `jobAdLanguage`, so a blank one resolves to `'en'` instead of `''`. Same bug class, one line away; behaviour for present values is unchanged.

## Testing

- [x] `pnpm exec vitest run` in `packages/prompts` — **519 passed**, 0 failed (15 files).
- [x] `pnpm --filter @ajh/prompts typecheck` — clean.
- [x] `pnpm exec eslint` / `prettier` — clean.
- [x] Added tests for new logic — unknown side, both-unknown, empty-string, and a real `en`/`de` mismatch that must still be flagged.

With the fix reverted:

```
× does not flag a mismatch when a language is unknown   → expected true to be false
× does not flag a mismatch on an empty-string language  → expected '' to be 'en'
```

The pre-existing `parses well-formed JSON and applies defaults` test — which asserts `mismatch: true` for a `de` ad against the defaulted `en` résumé — still passes, so the real-mismatch path is untouched.

## Checklist

- [x] New public APIs are documented (the helper is private, documented inline)

### DIFF
--- a/packages/prompts/src/generate/generate.test.ts
+++ b/packages/prompts/src/generate/generate.test.ts
@@ -1377,4 +1377,30 @@ describe('validateMetadata', () => {
     expect(meta?.jobCountry).toBe('');
     expect(meta?.jobLocation).toBe('');
   });
+
+  it('does not flag a mismatch when a language is unknown', () => {
+    // `??` is nullish-only and there was no 'unknown' guard, so an undetected
+    // side raised a spurious "rewrite entirely / do not translate" instruction.
+    // Matches analyze/validate.ts and @ajh/shared's detectLanguages, which both
+    // require BOTH sides to be known.
+    const meta = validateMetadata('{"resumeLanguage":"en","jobAdLanguage":"unknown"}');
+    expect(meta?.mismatch).toBe(false);
+    expect(
+      validateMetadata('{"resumeLanguage":"unknown","jobAdLanguage":"unknown"}')?.mismatch
+    ).toBe(false);
+  });
+
+  it('does not flag a mismatch on an empty-string language', () => {
+    // The model answers with "" rather than omitting the key, and `'' ?? 'en'`
+    // is `''` — so `'' !== 'en'` used to read as a mismatch.
+    const meta = validateMetadata('{"resumeLanguage":"","jobAdLanguage":"en"}');
+    expect(meta?.resumeLanguage).toBe('en');
+    expect(meta?.mismatch).toBe(false);
+  });
+
+  it('still flags a real mismatch between two known languages', () => {
+    const meta = validateMetadata('{"resumeLanguage":"en","jobAdLanguage":"de"}');
+    expect(meta?.mismatch).toBe(true);
+    expect(meta?.targetLanguage).toBe('de');
+  });
 });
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Flagging a mismatch when one side was simply unknown was noise pretending to be signal. The asymmetric blank-vs-known case the review raised is queued with the post-batch follow-ups. Merged.

saeedkolivand:

PR #811, opened 2026-07-21
fix(applications): read the row inside set_status's transaction

### LINKED ISSUE #810: fix(applications): set_status records a stale from_status (read outside its transaction)
## Describe the bug

`ApplicationStore::set_status` (`apps/desktop/src-tauri/src/applications/mod.rs:742`) reads the row through a lock it then **releases**, and only afterwards opens the write transaction:

```rust
let existing = self.get(id)?;              // takes the mutex, then RELEASES it
let now = now_ms();
let applied_at = …;
let mut guard = self.conn.lock();          // re-takes it
let tx = guard.transaction()?;
tx.execute("UPDATE applications SET status = ?2, …")?;
tx.execute(
    "INSERT INTO status_events (application_id, from_status, to_status, at, note) …",
    params![id, existing.status.as_id(), to.as_id(), …],   // ← from the stale read
)?;
```

A transition that commits in that gap is invisible to this call, so the `status_events` row it appends records a **`from_status` the row no longer had**. The append-only history then contains a link that never happened, and the chain of events no longer joins up.

This is reachable from two independent IPC paths: the UI's `applications_set_status`, and the extension bridge's `status.update`.

The transaction below already exists for exactly this class of reason ("*the row UPDATE and its append-only status event must land together or not at all*") — the read was just left outside it. The sibling `transition_status_if` even documents the hazard: *"unlike a caller doing its own `.get()` status check and then calling `set_status` … that pattern can lose a race between the check and the write."*

## Steps to reproduce

Two threads calling `set_status` on the same Application (`applied` / `saved`), then read `status_events` ordered by `rowid` — the chain breaks:

```
assertion failed: event 9 records from_status "saved", but the previous event moved the row to "applied"
```

Fails on 3 of 3 runs against current `main`.

## Expected behaviour

Every `status_events` row's `from_status` is the status the row actually held when the transition committed.

## Actual behaviour

`from_status` can be a stale value, so the history is wrong.

## Environment

- **OS**: any
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready that moves the read inside the existing transaction (via the conn-scoped `row_by_id_conn`), plus the race test above.

To be explicit about scope: this makes the recorded history correct and the read+write atomic. It deliberately does **not** turn `set_status` into a compare-and-set — an unconditional setter is its contract, and `transition_status_if` already exists for callers that need CAS.

thejesh23:
## Summary

`set_status` read the row through a lock it released before opening its write transaction, so a concurrent transition landing in the gap made it append a `status_events` row with a `from_status` the row no longer had. Fixes #810.

## Type of change

- [x] `fix` — Bug fix

## Changes

- The read now joins the **existing** transaction, via the conn-scoped `row_by_id_conn` instead of the self-locking `get`. The transaction was already there for the two writes; the read was simply outside it.
- Comment updated to say why the read belongs inside.
- New race test `set_status_records_a_consistent_history_chain_under_contention`.

**Scope, explicitly:** this makes the recorded history correct and the read+write atomic. It does **not** turn `set_status` into a compare-and-set — an unconditional setter is its contract, and `transition_status_if` already covers callers that need CAS (its doc comment even names this hazard).

**On `applications/mod.rs` and the R8 cap:** the file is 1395 LOC on `main` against a 1400 hard cap, so I kept this deliberately tight — net **+2 LOC (1397)**. (I tripped R8 with an earlier PR and would rather not repeat it.)

## Testing

- [x] `cargo test --lib` — **2705 passed**, 0 failed

The test drives two threads between `applied` and `saved`, then walks `status_events` **ordered by `rowid`** — insert order under the shared connection, so commit order — and asserts each `from_status` equals the previous event's `to_status`. It reads `rowid` directly rather than going through `events()` because `at` is millisecond-resolution and ties freely, which would make the ordering (and therefore the assertion) ambiguous.

Against `main`'s `set_status` it fails **3 of 3** runs:

```
assertion failed: event 9 records from_status "saved", but the previous event moved the row to "applied"
```

With the fix it passed **5/5** consecutive runs.

## Checklist

### DIFF
--- a/apps/desktop/src-tauri/src/applications/mod.rs
+++ b/apps/desktop/src-tauri/src/applications/mod.rs
@@ -739,8 +739,16 @@ impl ApplicationStore {
     /// Transition an Application's status, appending one history event and bumping
     /// `updated_at`. Sets `applied_at` the first time it leaves `saved`.
     pub fn set_status(&self, id: &str, to: ApplicationStatus, note: &str) -> AppResult<()> {
-        let existing = self
-            .get(id)
+        // The READ joins the same transaction as the two writes. The row UPDATE
+        // and its append-only status event must land together or not at all —
+        // otherwise a crash between them leaves the status changed with no
+        // history row — and reading through a separately-released lock let a
+        // concurrent transition land in the gap, so `from_status` below recorded
+        // a status this row no longer had. `.transaction()` needs
+        // `&mut Connection`, so call on `&mut *guard`.
+        let mut guard = self.conn.lock();
+        let tx = guard.transaction()?;
+        let existing = Self::row_by_id_conn(&tx, id)?
             .ok_or_else(|| AppError::Validation(format!("application not found: {id}")))?;
         let now = now_ms();
         let applied_at = if existing.applied_at.is_none() && !to.is_pre_apply() {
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. A status read outside its own transaction was a textbook toctou, and the chain invariant in the test is a nice touch that documents the property, not just the fix. Merged.

saeedkolivand:

PR #813, opened 2026-07-21
fix(autopilot): key the ai-notes skip on the merge's canonical key

### LINKED ISSUE #812: fix(autopilot): AI-notes skip compares raw URLs while the merge uses the canonical key
## Describe the bug

The AI-notes "already seen" skip compares **raw URLs**, but the store merge collapses jobs by their **canonical key** — so a job that re-surfaces under different tracking parameters is "new" to the skip and "old" to the merge. The result is a paid provider call whose note the merge immediately discards, on every run.

`commands/autopilot.rs:381` builds the skip set from raw URLs:

```rust
let prior_urls: std::collections::HashSet<&str> = autopilot
    .found_jobs
    .iter()
    .map(|j| j.url.as_str())
    .collect();
```

and `autopilot_helpers/mod.rs:482` compares raw:

```rust
if prior_urls.contains(job.url.as_str()) {
    continue; // re-surfaced: keeps its prior note via merge, don't re-pay
}
```

But the merge keys on `merge_key` → `canonical_job_key` → `normalize_job_url` (`autopilot/mod.rs:919`), which strips tracking params and fragments. So for `…/job/1` vs `…/job/1?utm_source=indeed`:

- the skip set says **new** → a `Completer::complete` call is made and charged;
- `merge_found_jobs` says **same job** → the existing entry (and its earlier note) wins, and the fresh note is dropped.

The comment right above the skip states the intent this misses: *"a re-surfaced job keeps its earlier note for free via the store merge, so re-generating would just burn a call whose result the merge discards — in steady state (nothing new) this makes ZERO provider calls."* Aggregators append varying `utm_*` params, so in practice this fires every run rather than never.

## Steps to reproduce

1. Enable the autopilot assistant so notes are generated.
2. Let a run discover `https://board.example/job/1`.
3. Next run, have the board surface the same job as `https://board.example/job/1?utm_source=indeed`.
4. A provider call is made and charged; the note is then discarded by the merge.

Reproduced against the loop directly — with two already-seen jobs re-surfaced under tracking params, the raw-URL comparison makes **2** provider calls where the expectation is **0**:

```
assertion `left == right` failed
  left: 2
 right: 0
```

## Expected behaviour

Zero provider calls for a job the merge will treat as already known.

## Actual behaviour

One call per re-surfaced job, up to `ASSISTANT_NOTES_MAX` (3) per run, each discarded.

## Environment

- **OS**: any
- **Ollama version**: n/a (any configured assistant provider)
- **Model in use**: the configured assistant model

## Additional context

Bounded waste (≤3 calls/run) rather than a wrong result, but it's spend the code explicitly intends to avoid. A PR is ready keying the skip on `canonical_job_key` — the same identity the merge uses.

thejesh23:
## Summary

The AI-notes skip set was built from raw URLs while the store merge keys on `canonical_job_key`, so a job re-surfacing under new tracking params was paid for and then discarded — every run. Fixes #812.

## Type of change

- [x] `fix` — Bug fix

## Changes

- `prior_urls: &HashSet<&str>` → `prior_keys: &HashSet<String>`, built and compared with `canonical_job_key(url, title, company)` — the identity `merge_key` / `merge_found_jobs` already use.
- Threaded through `generate_assistant_notes` → `run_notes_loop`; doc comments updated to say "merge key" rather than "url".
- The existing `jobs_all_in_prior_urls_make_zero_calls` test is renamed and strengthened: the two already-seen jobs now re-surface **with tracking params** (`?utm_source=indeed`, `#apply`), which is the case that was broken.

This closes the gap the skip's own comment describes — *"in steady state (nothing new) this makes ZERO provider calls"* — which raw-URL comparison could not deliver, since aggregators append varying `utm_*` params.

**On `autopilot_helpers/mod.rs` and the R8 cap:** the file is 1392 LOC on `main` against the 1400 hard cap. My first draft landed at 1403 and R8 caught it locally, so the comments and the test fixture are written tight — final is **1399**.

## Testing

- [x] `cargo test --lib` — **2704 passed**, 0 failed
- [x] Added tests for new logic (strengthened the existing one)

Reverting just the comparison to the raw URL, keeping the test:

```
assertion `left == right` failed
  left: 2
 right: 0
```

— i.e. two provider calls charged for jobs the merge would have collapsed.

## Checklist

### DIFF
--- a/apps/desktop/src-tauri/src/autopilot_helpers/mod.rs
+++ b/apps/desktop/src-tauri/src/autopilot_helpers/mod.rs
@@ -453,8 +453,8 @@ impl NoteEnv for LiveNoteEnv<'_> {
     }
 }
 
-/// The pure(ish) notes loop: for each job whose `url` is genuinely new (∉
-/// `prior_urls`), request one note through `env`, honoring — every iteration — the
+/// The pure(ish) notes loop: for each job whose merge key is genuinely new (∉
+/// `prior_keys`), request one note through `env`, honoring — every iteration — the
 /// top-N call ceiling, the run's cancellation token, and the shared daily-ceiling
 /// charge. Split from [`generate_assistant_notes`] (mirroring
 /// `agent::controller::run_agent`'s split from its `AgentEnv`) so a fake `env`
@@ -468,7 +468,7 @@ async fn run_notes_loop(
     env: &dyn NoteEnv,
     resume_fence: &str,
     found_jobs: &mut [FoundJob],
-    prior_urls: &HashSet<&str>,
+    prior_keys: &HashSet<String>,
     cancel: &CancellationToken,
 ) -> usize {
     // `calls` bounds *provider calls* (the real cost) to ≤ ASSISTANT_NOTES_MAX,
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Paying for a fresh ai note on every url variant of a job we'd already annotated was pure waste, keying on the canonical key restores the zero-calls steady state. The intra-batch duplicate case the review spotted goes on the follow-up list. Merged.

saeedkolivand:

PR #814, opened 2026-07-21
ci: label fork prs via pull_request_target

saeedkolivand:
## Why

Fork PRs never get labels: the labeler used the `pull_request` trigger, where GitHub caps `GITHUB_TOKEN` to read-only on forks, so the job was gated off with a same-repo `if:` and contributors' PRs showed a skipped labeler.

## What

- Switch the trigger to `pull_request_target` and drop the same-repo gate, so fork PRs get area/type labels too.
- Updated the header comment to document the new safety argument.

## Security rationale

`pull_request_target` is the accepted-safe pattern for this specific workflow: actions/labeler reads the changed-file list via the API and never checks out or executes PR code, and in this mode both the workflow file and `.github/labeler.yml` are read from the base branch, so a PR cannot tamper with either. The hard rule (now in the header comment): never add an `actions/checkout` of the PR head to this workflow.

Follow-up from the PR-batch session; complements the batch advisory follow-ups.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -2,19 +2,18 @@ name: 🏷️ PR Labeler
 
 # ──────────────────────────────────────────────────────────────────────
 # Applies area/type labels to pull requests from their changed paths
-# (config: .github/labeler.yml). Uses the `pull_request` trigger (not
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #816, opened 2026-07-22
fix(ai-generations): key the per-job aggregate on the normalized url

### LINKED ISSUE #815: fix(ai-generations): per-job aggregate keys on the raw job_url, so one job splits into two rows
## Describe the bug

`AiGenerationStore` keeps **one aggregate row per job**, but it keys that row on the **raw** `job_url`, so the same job reached under different tracking parameters splits into two aggregates — and the résumé, cover letter and answers the user generated for one job end up on separate records.

`save_application` (`apps/desktop/src-tauri/src/ai_generations/mod.rs:311`) looks the row up with the raw string:

```rust
pub fn save_application(&self, incoming: AiGenerationRecord) -> AppResult<String> {
    if let Some(existing) = self.find_by_job_url(&incoming.job_url) {
```

and `find_by_job_url` (`:206`) matches it literally:

```sql
FROM ai_generations WHERE job_url = ?1 ORDER BY created_at DESC LIMIT 1
```

The sibling `ApplicationStore` normalizes before deduping (`normalize_job_url`, which strips tracking params and fragments), so the two stores disagree on what "the same job" means. The codebase already knows this — `commands/ai_generations.rs:84-86`:

> *"…instead of a raw-vs-normalized url string compare (**which never matches for query-id boards like Indeed**: the Application stores the normalized url, the generation the raw one)."*

## Steps to reproduce

1. Tailor a résumé for `https://acme.com/job/1?utm_source=indeed` → aggregate row A.
2. Later, from a different entry point, save answers for the same job as `https://acme.com/job/1#apply`.
3. Two rows exist where the design says one.

Reproduced as a unit test — with the raw-url match, the "one aggregate row per job" assertion sees:

```
assertion `left == right` failed: one aggregate row per job
  left: 2
 right: 1
```

## Expected behaviour

One aggregate per job, keyed the same way `ApplicationStore` keys it.

## Actual behaviour

A second aggregate whenever the URL differs only in tracking params/fragment — which is the norm on aggregator and query-id boards.

## Environment

- **OS**: any
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready that normalizes on the write path (so the stored key matches `ApplicationStore`'s) and keeps a raw-url fallback lookup so rows written before the change still merge and are migrated onto the normalized key rather than orphaned.

thejesh23:
## Summary

The per-job generation aggregate keyed on the **raw** `job_url`, so the same job reached under different tracking params split into two rows. Fixes #815.

## Type of change

- [x] `fix` — Bug fix

## Changes

- `save_application` normalizes `incoming.job_url` through `applications::normalize_job_url` before the lookup **and** before the write, so the stored key is the same identity `ApplicationStore` dedupes on.
- **Legacy rows are migrated, not orphaned:** if the normalized lookup misses, it falls back to the raw url. A row written before this change is then merged and rewritten onto the normalized key (`merge_application` takes the incoming `job_url`, which is now normalized). Guarded so an unusable url (`normalize_job_url` → `""`) can't accidentally merge into an unrelated legacy row.

Closes the raw-vs-normalized split the codebase already documents at `commands/ai_generations.rs:84-86` — *"which never matches for query-id boards like Indeed"*.

**Blast radius check:** `AiGenerationStore::applied_job_urls()` returns these stored urls, so normalizing them could in principle move a consumer's goalposts. It has no production consumer — the `applied` flag in `commands/autopilot.rs:71` comes from `ApplicationStore::applied_job_urls()`, which is already normalized. So nothing downstream changes shape.

## Testing

Two new tests: the same job under `?utm_source=indeed` / `#apply` merging into one row keyed on the normalized url, and a legacy raw-url row still merging and being migrated. Against `main`:

```
assertion `left == right` failed: one aggregate row per job
  left: 2
 right: 1
```

The pre-existing `save_application_upserts_by_job_url_into_one_aggregate` and `save_application_inserts_separate_rows_when_unlinked` both still pass.

## Checklist

### DIFF
--- a/apps/desktop/src-tauri/src/ai_generations/mod.rs
+++ b/apps/desktop/src-tauri/src/ai_generations/mod.rs
@@ -308,8 +308,20 @@ impl AiGenerationStore {
     /// résumé, cover letter, answers, and brief from separate user actions land on
     /// one record; otherwise insert a fresh row (manual generations with no link).
     /// Returns the id of the affected row.
-    pub fn save_application(&self, incoming: AiGenerationRecord) -> AppResult<String> {
-        if let Some(existing) = self.find_by_job_url(&incoming.job_url) {
+    pub fn save_application(&self, mut incoming: AiGenerationRecord) -> AppResult<String> {
+        // Key the aggregate on the NORMALIZED url — the identity `ApplicationStore`
+        // already dedupes on. Matching the raw url split the same job into two
+        // "one-per-job" rows as soon as it was reached under different tracking
+        // params, which is the norm on query-id boards like Indeed.
+        let normalized = crate::applications::normalize_job_url(&incoming.job_url);
+        let raw = std::mem::replace(&mut incoming.job_url, normalized);
+        let mut found = self.find_by_job_url(&incoming.job_url);
+        // A row written before this normalization still carries its raw url; match
+        // it too, and the write below migrates it onto the normalized key.
+        if found.is_none() && !incoming.job_url.is_empty() && raw != incoming.job_url {
+            found = self.find_by_job_url(&raw);
+        }
+        if let Some(existing) = found {
             let merged = merge_application(existing, incoming);
             let id = merged.id.clone();
             self.update(&merged)?;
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. One job splitting into per-tracking-param aggregates was quietly wrecking the numbers, keying on the normalized url plus the legacy-row migration path is thorough work. The pre-existing insert race the review noted joins the follow-up list. Merged.

saeedkolivand:

PR #803, opened 2026-07-21
feat(landing): re-render world clips at native 1024x576 with av1 delivery

saeedkolivand:
## Summary

Quality v2 for `/world` (follow-up to #762, whose merge raced the final push): all 22 clips re-rendered at native 1024×576 / 576×1024 WAN (previously 832×480), SeedVR2-upscaled to 1080p-class, with AV1 delivery for desktop.

- Two weak slump landings curated out via seed re-rolls; all 20 seams re-verified frame-locked (RMS 6–17)
- Desktop: AV1 primary (1440w, 23 MB set) + H.264 fallback (1104w, 20 MB), picked at runtime via `canPlayType` in `WorldClient` — vendored engine untouched; mobile stays H.264 (720w, 17 MB)
- ADR-0019 weight/pipeline numbers refreshed (~64 MB stored, one set streamed per visitor)
- Includes the AI-review-gate lint fix (core `no-unused-vars` off for the vendored engine)

## Test plan

- [x] Owner approved the rendered quality on the dev server
- [x] All 20 seam frame-pairs re-verified (RMS 6.2–16.4)
- [x] typecheck + 119 tests + scoped ESLint + repo-wide lint:strict green
- [x] Landing build + check:parity green via pre-push gate

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #818, opened 2026-07-22
fix: pr-batch advisory follow-ups (rust/shared)

saeedkolivand:
Small deferred advisory follow-ups from the 2026-07-21 PR batch merge (#764–#803). Each is low-risk drift-prevention / contract-tightening; no behaviour change to shipped happy paths.

## Items

- **#778** — `refactor(autopilot)`: the cancel branch in `autopilot_run` was duplicated across the scrape `Err` arm and the pre-record guard. Extracted one private `finish_cancelled` helper (site-specific span message passed as a parameter) so the two sites can't drift. Behaviour is byte-identical; the engine token unregister stays at each caller's own point.
- **#780** — `fix(spend)`: the free-call loopback gate matched hosts by literal string only, so the fully-expanded `[0:0:0:0:0:0:0:1]` and IPv4-mapped `[::ffff:127.0.0.1]` spellings billed a local `openai-compatible` server at the unknown-model rate, and a `user@[::1]` userinfo prefix defeated the bracket match. Added the extra literal arms (bracketed + bare) and strip an optional `userinfo@` before host parsing. Still literal-only — this is a cost/display gate, never routing or security, so no IP-parsing crate. Tests added.
- **#782** — `fix(shared)`: added `.min(0)` to `ApplicationUpdateSchema.nextActionAt` so the wire contract matches the server-side `parse_next_action_at` guard that already rejects a negative epoch-ms. `pnpm gen:ipc` produced no IPC `.rs` diff (`.min(0)` is a runtime refinement, not a type change). Also made that guard's rejection message echo the offending JSON type for easier renderer-side debugging.
- **#790** — `test(shared)`: `language-detection` maps both `nob` (Bokmål) and `nno` (Nynorsk) to `no`, but the suite only exercised the `nob` sample. Added a Nynorsk sample (verified franc detects it as `nno`) so the `nno` branch isn't silently regressable.

## Validation

- `cargo clippy --all-features` — clean (touched crate)
- `cargo test` — spend / autopilot / applications module tests pass; full pre-push suite incl. architecture boundary tests green
- `cargo fmt --check` — clean
- `pnpm typecheck` (whole graph, forced) — 13/13
- `pnpm test` — 4047 passed (incl. the new Nynorsk case and the shared namespace-enumeration test)

### DIFF
--- a/apps/desktop/src-tauri/src/commands/applications.rs
+++ b/apps/desktop/src-tauri/src/commands/applications.rs
@@ -60,9 +60,19 @@ fn parse_next_action_at(raw: Option<Value>) -> AppResult<Option<Option<u64>>> {
         None => Ok(None),
         Some(Value::Null) => Ok(Some(None)),
         Some(other) => other.as_u64().map(|ms| Some(Some(ms))).ok_or_else(|| {
-            AppError::Validation(
-                "next_action_at must be null or a non-negative integer timestamp in ms".into(),
-            )
+            // Echo the offending JSON type so a renderer sending the wrong shape
+            // (a string, a negative/fractional number, an object) sees why.
+            let kind = match &other {
+                Value::Null => "null",
+                Value::Bool(_) => "a boolean",
+                Value::Number(_) => "a non-integer or out-of-range number",
+                Value::String(_) => "a string",
+                Value::Array(_) => "an array",
+                Value::Object(_) => "an object",
+            };
+            AppError::Validation(format!(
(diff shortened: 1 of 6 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Four deferred advisories closed in one tidy pass: the cancel handler can't drift anymore, the rarer loopback spellings bill as local, the wire contract finally agrees with the server on negatives, and nynorsk has a test watching it. Merged.

saeedkolivand:

PR #819, opened 2026-07-22
fix(credentials): never treat an unreadable metadata index as empty

### LINKED ISSUE #817: fix(credentials): a corrupt metadata index is cached as empty, and the next write orphans every credential
## Describe the bug

Two problems in `CredentialStore`'s metadata index (`apps/desktop/src-tauri/src/credentials/mod.rs:161`/`:174`).

**1. A corrupt index is cached as empty, and the next write makes the damage permanent.**

```rust
fn load_meta(&self) -> HashMap<String, CredentialMeta> {
    …
    let loaded: HashMap<String, CredentialMeta> = std::fs::read_to_string(&self.meta_file)
        .ok()
        .and_then(|s| serde_json::from_str(&s).ok())
        .unwrap_or_default();          // ← read error AND parse error both → empty
    guard.0 = Some(loaded.clone());    // ← and it is cached as authoritative
    loaded
}
```

An unreadable or malformed `credential-meta.json` is indistinguishable from "no file yet". Because `save_meta` writes the **whole** map, the next `set`/`remove` then persists only its own entry and **orphans every previously saved credential** from the index. The keychain secrets themselves survive, but nothing can find them again.

**2. `set`/`remove` are a non-atomic read-modify-write.**

```rust
let mut meta = self.load_meta();   // locks the cache, then RELEASES it
meta.insert(…);
self.save_meta(meta)?;             // locks it again
```

The mutex is held *inside* `load_meta` and `save_meta`, never across the pair, so two concurrent writers can both start from the same snapshot and the second drops the first's entry.

## Steps to reproduce

**Corrupt index:** put anything non-JSON in `credential-meta.json`, then save or remove any credential. The call succeeds and the file is replaced with just that one entry — every other saved credential is now unreachable.

**Lost update:** two threads writing entries concurrently; some entries are absent from the file afterwards.

## Expected behaviour

- A corrupt/unreadable index is an error, not an empty index — and must never be silently overwritten.
- Concurrent writes both land.

## Actual behaviour

- The corrupt read is cached as empty and the next write replaces the file with a one-entry map.
- Concurrent writes can lose entries.

## Environment

- **OS**: any (metadata index is plain JSON on disk; the keychain is untouched)
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

A PR is ready that (a) distinguishes "file absent" from "unreadable/corrupt" and refuses to write over the latter, and (b) holds the cache lock across the whole read-modify-write via a single `mutate_meta` helper.

thejesh23:
## Summary

A corrupt `credential-meta.json` was cached as an **empty** index, so the next write replaced the file with a single entry and orphaned every other saved credential; and `set`/`remove` were a non-atomic read-modify-write. Fixes #817.

## Type of change

- [x] `fix` — Bug fix

## Changes

- **`read_meta_file()`** distinguishes the three outcomes: file absent → empty index (first run), present-but-unparseable → `Parse` error, unreadable → `Storage` error.
- **`load_meta()`** (the read-only path) still degrades to empty for its callers — it has nothing better to report — but no longer **caches** that, so a later write can't mistake an unreadable index for an empty one. The failure is logged.
- **`mutate_meta(f)`** holds the cache lock across the entire read-modify-write and is now the single write path for `set`/`remove`. On an unreadable index it returns the error instead of replacing the file, so the user's file stays recoverable by hand.
- `save_meta` is gone — `mutate_meta` subsumes it (and leaving it would trip `clippy -D warnings` as dead code).

## Testing

Two new tests: a corrupt index must not be cached as empty and must make a write refuse (asserting the file is left byte-for-byte intact), and 80 concurrent writes across two threads must all land.

**On demonstrating the "before":** I couldn't run these unchanged against `main`, because they exercise `mutate_meta`, which this PR introduces — I'd rather say that than imply a clean before/after. So I isolated defect (1) instead, reverting only the corrupt-read behaviour to `unwrap_or_default()` while keeping everything else:

```
thread 'credentials::test::a_corrupt_meta_file_is_not_cached_as_empty' panicked:
a write must not proceed on an unreadable index
```

i.e. with the old behaviour the write goes through and the index is replaced.

## Checklist

- [x] New public APIs are documented (both new helpers are private, documented inline)

### DIFF
--- a/apps/desktop/src-tauri/src/credentials/mod.rs
+++ b/apps/desktop/src-tauri/src/credentials/mod.rs
@@ -112,17 +112,16 @@ impl CredentialStore {
             .set_password(password)
             .map_err(|e| format!("keyring set error: {e}"))?;
 
-        let mut meta = self.load_meta();
-        meta.insert(
-            board_id.to_string(),
-            CredentialMeta {
-                board_id: board_id.to_string(),
-                username: username.to_string(),
-                saved_at: now_ms(),
-            },
-        );
-        self.save_meta(meta)?;
-        Ok(())
+        self.mutate_meta(|meta| {
+            meta.insert(
+                board_id.to_string(),
+                CredentialMeta {
+                    board_id: board_id.to_string(),
+                    username: username.to_string(),
+                    saved_at: now_ms(),
+                },
+            );
+        })
     }
 
     pub fn remove(&self, board_id: &str) -> AppResult<()> {
@@ -131,10 +130,9 @@ impl CredentialStore {
             // entry if the keychain was cleared externally.
             entry.delete_credential().ok();
         }
-        let mut meta = self.load_meta();
-        meta.remove(board_id);
-        self.save_meta(meta)?;
-        Ok(())
+        self.mutate_meta(|meta| {
+            meta.remove(board_id);
+        })
     }
 
     /// Remove every stored credential — board passwords and all AI/provider keys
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Treating a corrupt metadata index as empty and then caching that emptiness was a quiet path to overwriting real credentials metadata, separating absent from corrupt closes it properly. Merged.

saeedkolivand:

PR #821, opened 2026-07-22
fix(scraping): cap the body read on the user-supplied-url path

### LINKED ISSUE #820: fix(scraping): scrape_url buffers an attacker-influenced response body with no size cap
## Describe the bug

`resolve_uncached` (`apps/desktop/src-tauri/src/scraping/scrape_url/mod.rs:103`) buffers the whole response body with no size limit:

```rust
let html = match res.text().await {
    Ok(h) => h,
    Err(_) => return Ok(None),
};
```

This is the generic-HTML fallback for an **attacker-influenced** URL — that's precisely why the fetch goes through `net::http::get_guarded_following_redirects` rather than the shared client. The guard covers *where* we connect; nothing covers *how much* we then read. `Response::text()` accumulates the entire body in memory, so a hostile (or merely broken) host can stream an arbitrarily large one.

The 20s timeout bounds how **long** we read, not how **much** — plenty of time to deliver hundreds of MB on a fast link.

The board fetch path already gets this right: `scraping::http::fetch_text` applies an 8 MB `MAX_BYTES` cap, with a `Content-Length` pre-check plus a streamed accumulation that aborts mid-body ("*prevents OOM from a server that lies about or omits Content-Length*"). The `scrape_url` path just never received the same treatment.

Note `fetch_text` is not a drop-in here: it uses the shared client, so routing through it would **drop the SSRF guard** on exactly the URLs that need it.

## Steps to reproduce

Import a job from a URL whose host streams an unbounded body (e.g. an endless `text/html` response). Memory grows until the read finishes or the process dies; nothing intervenes.

## Expected behaviour

The read stops at a bounded size, same as the board path.

## Actual behaviour

Unbounded buffering.

## Environment

- **OS**: any
- **Ollama version**: n/a
- **Model in use**: n/a

## Additional context

Hardening rather than a wrong result — it needs a hostile/broken host to matter — but it's the one egress that takes a user-supplied URL, and the cap already exists a few modules over.

A PR is ready that extracts the existing capped-read logic out of `fetch_text` into a shared `read_text_capped(response, cap)` and uses it on this path too, so there is one implementation rather than two.

thejesh23:
## Summary

`resolve_uncached` buffered an attacker-influenced response body with `Response::text()` and no size limit. Now capped at the same 8 MB the board path uses. Fixes #820.

## Type of change

- [x] `fix` — Bug fix

## Changes

- Extracted the existing capped-read logic out of `fetch_text` into `scraping::http::read_text_capped(response, cap)` — `Content-Length` pre-check, streamed accumulation that aborts mid-body, and the same `Content-Type` charset handling (so umlauts / € still decode correctly).
- `fetch_text` now calls it, so there is **one** implementation rather than a copy.
- `resolve_uncached` reads its generic-HTML fallback through it with an 8 MB cap.

**Why not just call `fetch_text` there:** it uses the shared client, so routing this path through it would drop the SSRF guard — `scrape_url` deliberately fetches via `net::http::get_guarded_following_redirects` because the URL is user-supplied. The shared *reader* gives the bound without touching *which client* connects.

The 20s timeout bounds how long we read, not how much, so it isn't a substitute for this.

## Testing

- [x] `cargo test --lib scraping` — 880 passed (the `scrape_url` and `http` suites both exercise the touched paths)

Two new wiremock tests cover the helper directly: a body under the cap round-trips, and one over it fails with "Response too large" instead of being buffered. The 40 pre-existing `scraping::http` tests still pass, which is what covers the `fetch_text` refactor being behaviour-preserving.

## Checklist

- [x] New public APIs are documented (`read_text_capped` is `pub(crate)`, documented inline)

### DIFF
--- a/apps/desktop/src-tauri/src/scraping/http/mod.rs
+++ b/apps/desktop/src-tauri/src/scraping/http/mod.rs
@@ -99,6 +99,63 @@ pub struct FetchResult {
     pub text: String,
 }
 
+/// Read a response body as text, refusing to buffer more than `cap` bytes.
+///
+/// Two guards: a cheap `Content-Length` pre-check for honest servers, then a
+/// streamed accumulation that aborts the moment the running total exceeds `cap`
+/// — so a server that lies about or omits `Content-Length` still can't drive us
+/// into OOM. The charset comes from `Content-Type`, mirroring what
+/// `reqwest::Response::text()` does internally, so German umlauts / € decode
+/// correctly regardless of the cap.
+///
+/// Public to the crate because callers that hold a `Response` from their OWN
+/// client — e.g. the SSRF-guarded `net::http::get_guarded*` used for
+/// attacker-influenced URLs — cannot route through [`fetch_text`] (that would
+/// drop the guard) but need the same bound.
+pub(crate) async fn read_text_capped(response: reqwest::Response, cap: usize) -> AppResult<String> {
+    if let Some(content_length) = response.content_length() {
+        if content_length > cap as u64 {
+            return Err(AppError::Validation("Response too large".to_string()));
+        }
+    }
+
+    let encoding = response
+        .headers()
+        .get(reqwest::header::CONTENT_TYPE)
+        .and_then(|ct| ct.to_str().ok())
+        .and_then(|ct| {
+            // Extract charset=... from e.g. "text/html; Charset="ISO-8859-1""
+            // Key match is case-insensitive; strip surrounding quotes from value.
+            ct.split(';').find_map(|part| {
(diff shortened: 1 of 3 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Capping the body read on the user-supplied-url path shuts the obvious memory hole. The review's point stands that the personio feed resolver still does the same unbounded read one layer down, that sibling cap is on the follow-up list so the class gets closed, not just the instance. Merged.

saeedkolivand:

PR #823, opened 2026-07-22
fix(agent): clamp confirm-request args to the called tool's own cap

### LINKED ISSUE #822: fix(agent): confirm gate shows/edits more cover-letter text than save_cover_letter keeps
## Describe the bug

The agent confirm gate can show — and let the user **edit** — more cover-letter text than `save_cover_letter` will actually persist, so the approved text is silently clipped on save.

The display clamp is the **max** across both gated Write tools (`agent/gate.rs:137`):

```rust
const ARGS_DISPLAY_CAP: usize = if COVER_LETTER_CAP > SAVED_RESUME_CAP {
    COVER_LETTER_CAP
} else {
    SAVED_RESUME_CAP
};
```

with `COVER_LETTER_CAP = 20_000` and `SAVED_RESUME_CAP = 40_000` (`agent/tools.rs:501`/`:565`), so it resolves to **40 000** for *both* tools. But `save_cover_letter_handler` truncates at its own cap (`agent/tools.rs:474`):

```rust
.chars()
.take(COVER_LETTER_CAP)
.collect();
```

So for a 20 000–40 000-char cover letter the confirm step displays the whole thing, the user reviews and approves (possibly after editing) all of it, and only the first 20 000 chars are saved. The user has approved text that is not the text that gets persisted.

The module's own security invariant says this shouldn't happen:

> *"**Display/edit fidelity.** Confirm-request args are clamped to the LARGER of `COVER_LETTER_CAP` and `SAVED_RESUME_CAP` … the renderer shows/edits **exactly what will be persisted**, never a truncated preview."*

Taking the max guarantees the "never show *less*" half. The "never show *more*" half doesn't hold for whichever tool has the smaller cap.

## Steps to reproduce

1. Run the agent flow so it proposes `save_cover_letter` with a letter longer than 20 000 chars.
2. The confirm prompt shows the full text (up to 40 000) and lets you edit it.
3. Approve. The stored letter is the first 20 000 chars.

## Expected behaviour

The confirm step shows exactly what will be saved for the tool being called.

## Actual behaviour

Up to 40 000 chars shown/edited; 20 000 saved, silently.

## Environment

- **OS**: any
- **Ollama version**: n/a
- **Model in use**: any (needs a >20k-char generated letter, so a pathological output rather than a typical ~2k one)

## Additional context

Low likelihood — real cover letters are far under 20k — but the confirm gate is the safety core, so "what you approved is what was saved" seems worth keeping exact. A PR is ready clamping per tool (`display_cap_for`), with the max kept as the fallback for any future tool.

thejesh23:
## Summary

The confirm gate clamped args to the **max** cap across both Write tools (40k), so a 20k–40k cover letter was shown and edited in full and then silently clipped to 20k on save. Now clamped to the called tool's own cap. Fixes #822.

## Type of change

- [x] `fix` — Bug fix

## Changes

- New `display_cap_for(tool)`: `COVER_LETTER_CAP` for `save_cover_letter`, `SAVED_RESUME_CAP` for `save_resume`, and `ARGS_DISPLAY_CAP` (the max) for anything unknown.
- `resolve_write` clamps with `display_cap_for(&call.name)` instead of the shared max.
- Updated the module's *Display/edit fidelity* security invariant to describe the per-tool rule, since it currently documents the max-based one.

`ARGS_DISPLAY_CAP` is kept and still guards the important direction — never show **less** than a Write tool will persist — which is why it stays the fallback for a future tool. The per-tool clamp closes the other direction for the tool with the smaller cap.

## Testing

- [x] `cargo test --lib agent::gate` — 23 passed

Two new tests: the per-tool mapping (including the unknown-tool fallback), and a >20k letter clamping to exactly `COVER_LETTER_CAP + 1` (the cap plus the ellipsis) rather than the resume cap. The pre-existing `clamp_json_strings_truncates_long_leaves_but_keeps_structure` is untouched and still passes.

The "these two caps actually differ" check is a `const _: () = assert!(…)`, so it fails the **build** if they're ever equalised — a runtime `assert!` on two consts trips `clippy::assertions_on_constants`.

**Honest note on severity:** this needs a >20 000-char generated cover letter to bite, which is well outside normal output (~2k). I filed it because the confirm gate is the safety core and "what you approved is what was saved" reads like an invariant worth keeping exact, not because I've seen it happen.

## Checklist

- [x] New public APIs are documented (the helper is private, documented inline)

### DIFF
--- a/apps/desktop/src-tauri/src/agent/gate.rs
+++ b/apps/desktop/src-tauri/src/agent/gate.rs
@@ -26,10 +26,12 @@
 //! - **Prompt injection can REQUEST but never EXECUTE.** Hostile job/résumé text can
 //!   make the model *ask* for a write; it can never satisfy the gate on the user's
 //!   behalf — only a real `agent_confirm` IPC call (or a `resolve` in a test) can.
-//! - **Display/edit fidelity.** Confirm-request args are clamped to the LARGER of
-//!   [`COVER_LETTER_CAP`] and [`SAVED_RESUME_CAP`] (each gated Write tool's own
-//!   content cap), not an arbitrarily smaller number — the renderer shows/edits
-//!   exactly what will be persisted, never a truncated preview.
+//! - **Display/edit fidelity.** Confirm-request args are clamped to the CALLED
+//!   tool's own content cap ([`display_cap_for`]) — [`COVER_LETTER_CAP`] for
+//!   `save_cover_letter`, [`SAVED_RESUME_CAP`] for `save_resume` — so the renderer
+//!   shows/edits exactly what will be persisted: never a truncated preview, and
+//!   never more than the tool will keep. An unknown tool falls back to the larger
+//!   of the two, which preserves the never-show-less half of the guarantee.
 //!
 //! [`ToolKind::Write`]: super::tools::ToolKind
 //! [`ToolContext`]: super::tools::ToolContext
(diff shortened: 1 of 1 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. Clamping confirm-request args to the tool's own cap keeps one tool's generosity from becoming another's loophole. Merged.

saeedkolivand:

PR #800, opened 2026-07-21
fix(monitoring): give the last-24-hours sparkline an actual 24-hour window

thejesh23:
## Summary

The "Last 24 hours" sparkline had no time filter and bucketed by hour **of day**, so all-time completions were mixed together and a job from three days ago shared today's bar. Fixes #799.

## Type of change

- [x] `fix` — Bug fix

## Changes

- New pure helper `features/monitoring/lib/last-24-hours.ts` (`binLast24Hours(jobs, now)`), following the existing `features/<name>/lib/` convention (`jobs/lib/canonical-job-key.ts`, `applications/lib/extract-recipient.ts`):
  - filters to `now - 24h < ts <= now`;
  - buckets by **hours-ago**, oldest bucket first, so the bars read left → right chronologically instead of by wall-clock hour;
  - takes `now` as a parameter, so the caller owns the clock and the tests can pin it.
- `MonitoringPage` now calls it: `useMemo(() => binLast24Hours(allJobs, Date.now()), [allJobs])`.
- The `finishedAt ?? updatedAt` timestamp choice and the completed-only filter are unchanged.

Old behaviour, run with the exact previous expression on two completed jobs — one an hour old, one three days old:

```
old total counted: 2 (one of them is 3 days old)
old bins nonzero at hour-of-day: [ 3, 4 ]
```

Both counted, and the positions were hours-of-day rather than a timeline.

## Testing

- [x] `pnpm exec vitest run src/renderer/features/monitoring/lib/last-24-hours.test.ts` — 6 passed, 0 failed.
- [x] `pnpm exec eslint` / `prettier --check` on the changed files — clean.
- [x] Added tests for new logic: empty input, out-of-window exclusion, oldest-first ordering, completed-only, the `updatedAt` fallback, and a future timestamp.

**Note on `pnpm typecheck`:** not runnable here — on a fresh clone `apps/desktop` fails typecheck because `routeTree.gen.ts` doesn't exist until the router plugin has run. Pre-existing on `main` and unrelated to this change.

## Checklist

- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org)
- [x] No `console.log` left in source code
- [x] No `@ts-ignore` added without explanation
- [x] New public APIs are documented (the helper carries a doc comment)

PR #824, opened 2026-07-22
fix: pr-batch advisory follow-ups (renderer)

saeedkolivand:
Deferred renderer follow-ups from the 2026-07-21 PR batch (#764–#803).

## Items

- **Test-suite de-flake (highest priority)** — `JobLocationPreferences` scheduled two `setTimeout`s (the 150ms `onBlur` hide and a 0ms position-measure) that call `setState` but were never cancelled on unmount. When a vitest file finished inside the 150ms window the deferred `setState` fired after jsdom teardown → `ReferenceError: window is not defined` → the whole Tests job exited 1 with all tests passing (red Tests on 5 PRs last batch). Both timer ids now live in refs and are cleared in a `useEffect` cleanup; the 150ms blur delay behaviour is preserved.
- **#800 follow-up** — the last-24h sparkline data is hours-ago indexed (bin 23 = most recent hour), but `Sparkline` still drew hour-of-day labels (`12am`…`9pm`) and highlighted `new Date().getHours()`, mislabeling every bar. Now highlights the last bucket and uses relative axis notation (`-24h … now`, following the component's existing hardcoded-label precedent). Adds the two boundary tests to `last-24-hours.test.ts`: a completion at exactly `now` is included, one at `windowStart` is excluded.
- **#798 follow-up** — dropped the unreachable `cut > 0 ? id.slice(0, cut) : a.id` fallback in `useActivityFeed` (every live id is `${jobId}-${ts}`, so `lastIndexOf('-')` is always positive); slice directly.
- **#796 follow-up** — compressed the 6-line `handleTailor` rationale comment to two lines, and added a deferred-save ordering test proving `persistJob('applied')` fires only after `saveFromPosting` resolves.
- **#794 follow-up (LOW advisory)** — applied the reducer-purity fix: `Date.now()` is now read at dispatch time and closed into the functional patch, so StrictMode dev double-invocation can't produce two timestamps in the `setStepLogs` reducer body.

## Gates

- `TURBO_FORCE=1 pnpm typecheck` — 13/13 packages green
- `pnpm lint:strict` — clean (`--max-warnings 0`)
- `pnpm test` — 373 files / 4052 tests passing (full suite, no jsdom flake)

### DIFF
--- a/apps/desktop/src/renderer/features/autopilot/hooks/useAutopilotRun.ts
+++ b/apps/desktop/src/renderer/features/autopilot/hooks/useAutopilotRun.ts
@@ -60,10 +60,14 @@ export function useAutopilotRun() {
         ),
       }));
     }
+    // Read the clock at dispatch time, not inside the reducer body: under
+    // StrictMode dev double-invocation the reducer runs twice from the same base,
+    // and `Date.now()` there would produce two different timestamps (impure).
+    const ts = Date.now();
     setStepLogs((prev) => ({
       [event.autopilotId]: [
         ...(prev[event.autopilotId] ?? []).slice(-49),
-        { step: event.step, detail: event.detail, ts: Date.now() },
+        { step: event.step, detail: event.detail, ts },
       ],
     }));
   }, []);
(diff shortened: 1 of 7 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Merging with the review clean. The five renderer follow-ups land here, headlined by the timer-leak fix that was randomly reddening pr test runs. The reviewer's blur-refocus edge case lands separately as a one-liner right behind this.

saeedkolivand:

PR #825, opened 2026-07-22
fix: pr-batch advisory follow-ups (extension)

saeedkolivand:
## Summary

Deferred advisory follow-ups from the 2026-07-21 extension PR batch. Four items, all under `apps/extension/src/lib`, no protocol/permission changes.

## Items

**Item 1 — #786 MEDIUM: hidden résumé file input.** A custom upload widget hides the native `<input type=file>` (`display:none`) behind a styled button, which defeated the "decisive file input" short-circuit in `looksLikeApplicationForm` — a real application form with a hidden résumé input and fewer than three other visible fields went unrecognized. **Fix:** a résumé/CV-flavored file input (name/id/label/aria-label, or an `accept` listing document types) is now decisive **even when hidden**, checked across all inputs (not just the visible-filtered set). Narrowed to résumé-flavored inputs so an arbitrary hidden file input can't masquerade as an application form — the module keeps under-reporting over over-reporting. Visibility stays computed-style-only.

**Item 2 — #786 LOWs (both fixed; small + clearly correct under the under-report invariant):**
- *"Save draft" false positive:* a `type=submit` "Save draft"/"Add another" button inside the real application form still fires the form's `submit`, which the listener saw as sending the application. Now the `SubmitEvent`'s `submitter` is inspected and a non-final-submit control is skipped. A programmatic `form.submit()` (no submitter) still fires, as before.
- *checkbox/radio-only forms:* excluded checkbox/radio from the fillable-field count, so a filter / cookie-consent / survey widget no longer reads as an application form. Real application forms clear the bar on their text/email/résumé fields.

**Item 3 — #784 follow-up: camelCase compounds.** `textSignal` lowercases without splitting camelCase, so `name="workCity"` flattens to `workcity` and no longer matched the location pattern (`work` was not an enumerated city/town prefix). **Decision: extend, narrowly** — added `work` to the enumerated city/town prefix so `workCity` resolves to `location`, while keeping the leading-anchor enumeration so `ethnicity` stays a non-match (no new mis-fill surface). This favors the under-fill contract: it recovers a legitimate field without broadening the anchor. Phone needed no code change — `work` is already enumerated (`workPhone` matches), and `headphone`/`smartphone` correctly do not.

**Item 4 — #805 nit:** the localized-fields comment pointed only at `AMBIGUOUS_WORDS`, but the collision-prone `referr`/`reference`/`search` terms now live in `AMBIGUOUS_PREFIXED`. Comment now points at both lists.

## Tests

New fixtures pin every behavior change:
- `submit-watch.test.ts` — hidden résumé file input + <3 visible fields recognized; bare hidden non-résumé file input still ignored (the narrowing); checkbox/radio-only form ignored; "Save draft" `submitter` skipped; real submit `submitter` still fires.
- `field-signal.test.ts` — `headphone` (phone non-match), `workphone` (phone match), `workcity`/`homecity` (location match).

## Gates

- `pnpm lint:strict` — clean
- `TURBO_FORCE=1 pnpm typecheck` — 13/13 successful (cache bypassed)
- `pnpm test` (full) — 4050 passed / 373 files

Source PRs: #786, #784, #805.

### DIFF
--- a/apps/extension/src/lib/field-signal.test.ts
+++ b/apps/extension/src/lib/field-signal.test.ts
@@ -73,6 +73,9 @@ describe('matchNamedKey — phone', () => {
       'cell phone',
       'cellphone',
       'work phone',
+      // camelCase compound flattens to `workphone`; `work` is an enumerated
+      // prefix, so it still resolves to phone (see NAMED_KEY_PATTERNS note).
+      'workphone',
       'home phone',
       'mobile',
       'mobile_number',
@@ -88,8 +91,16 @@ describe('matchNamedKey — phone', () => {
 
   it('does not match a word that merely CONTAINS phone/mobile', () => {
     // Bare `phone`/`mobile` were unanchored substrings, so a "Smartphone model"
-    // field resolved to `phone` and received the user's phone number.
-    for (const signal of ['smartphone', 'smartphone model', 'iphone', 'microphone', 'automobile']) {
+    // field resolved to `phone` and received the user's phone number. `headphone`
+    // has a non-enumerated `head` prefix, so it stays a non-match too.
+    for (const signal of [
+      'smartphone',
+      'smartphone model',
+      'iphone',
+      'microphone',
+      'headphone',
+      'automobile',
+    ]) {
       expect(matchNamedKey(signal), signal).not.toBe('phone');
     }
   });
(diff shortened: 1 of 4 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Thanks. The hidden resume input was the last realistic way a genuine application form could dodge the tracker, and the save-draft submitter guard kills the loudest false positive. Both directions still respect under-report over over-report. Merged, any findings from the in-flight advisory review get handled as follow-ups.

saeedkolivand:

PR #826, opened 2026-07-22
fix(settings): clear a pending blur timer before re-arming it

saeedkolivand:
Follow-up to the LOW flagged in #824's review — the exact bug class that PR targeted. In `JobLocationPreferences`, `onBlur` overwrote `blurTimerRef.current` without clearing any prior pending id, so a blur → refocus-within-150ms → blur sequence orphaned the first timer; an unmount in that window cleared only the second, letting the first fire after jsdom teardown (`ReferenceError: window is not defined`) and re-open the test-flake it was meant to close. This clears the pending id before re-arming, mirroring the position-timer pattern already in the same component. One-line change; the component's tests and `pnpm typecheck` are green.


saeedkolivand:
Thanks. That closes the last edge of the timer-leak story, both timers now clear before re-arming and on unmount. Merged.

saeedkolivand:

PR #828, opened 2026-07-22
fix: persist completed stream text so the poll fallback works

saeedkolivand:
## Why

PR #802 ("prefer the persisted result over a truncated stream buffer") made the renderer's poll fallback (`awaitAiStream`) resolve `persisted.length > buffer.length ? persisted : buffer`, reading `result.text` from the completed job. Its review found the fix was a **runtime no-op**: the backend never populated any text on the completed-generation path — the job result was `{ done: true }` — so `result.text` was always `undefined`. A dropped interior stream frame (or a missed terminal `done` event) still resolved a truncated buffer.

## What

Populate the fallback for real, backend-side, at the **provider-agnostic** completion seam:

- **`stream::finish`** (shared by openai / anthropic / gemini / ollama via `stream_response`) now accumulates the completed answer from **non-thinking deltas only** and persists it as `result.text`. A new adapter inherits this for free.
- **`cli_agent::emit_done`** (the CLI-agent subprocess streaming path — claude-code / codex / gemini-cli — which feeds the *same* `awaitAiStream` contract) does the same, so the poll fallback is fixed for that provider family too. No per-provider duplication: both call one shared `strip_think_blocks`.

## Think-strip decision (the critical consistency point)

The renderer's streamed buffer is `<think>`-stripped renderer-side (via `think-split.ts`), and the poll prefers the **longer** of {persisted, buffer}. If the backend persisted **raw** `<think>…</think>` markup, that side would be spuriously longer **and** would resolve reasoning markup into the final document.

I chose to **persist think-stripped text backend-side** (not strip at the renderer comparison boundary), via `strip_think_blocks` mirroring `createThinkSplitter`'s final output exactly (blocks removed; an unterminated `<think>` discards to end). This keeps both sides of the length comparison in the **same shape**, guarantees no path can leak reasoning markup, and needs **no renderer-source change** — only the test was updated. (A `</think>` split across two stream frames — which the renderer's streaming splitter can mis-handle — is resolved correctly here because the whole answer is stripped in one pass, so the persisted text is strictly equal-or-more-correct.)

## Persistence / size

The job result is stored as SQLite `TEXT` (`jobs.result`) and rides the `job.completed` event `Value` — both untyped JSON, so the new `text` field is additive. A generation is at most a few hundred KB (already streamed delta-by-delta); well within SQLite's TEXT limit — **no truncation**.

## Tests

- Rust (`stream.rs`): pinned that the persisted answer excludes provider-flagged reasoning deltas, that inline `<think>` is stripped and **can never leak markup**, cross-frame close-tag robustness, and `strip_think_blocks` cases mirroring the renderer splitter. The tested control-flow core (`drive_stream` / `StreamSink::Complete`) now carries the accumulated answer so it stays an honest mirror of production.
- Renderer (`stream-promise.test.ts`): rewrote the false-confidence mock (was `result: { text }`, a shape the backend never produced) to the **real** contract `result: { done: true, text }`, and added a case pinning that a recovered persisted result never surfaces `<think>` markup.

Gates: `cargo clippy --all-targets --all-features --locked -D warnings`, `cargo fmt --check`, `cargo test` (ai_provider), `pnpm typecheck`, full `pnpm test`, `pnpm lint:strict` — all green (one unrelated `use-extension-bridge` timing flake, passes in isolation).

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Completed AI responses are now reliably saved and recovered when streamed updates are missed.
  * Reasoning content and `<think>` markup are excluded from final responses.
  * Recovered responses no longer expose internal reasoning tags.

* **Tests**
  * Added coverage for response recovery, persisted completion text, and reasoning-content filtering.

### DIFF
--- a/apps/desktop/src-tauri/src/commands/ai_provider/cli_agent/mod.rs
+++ b/apps/desktop/src-tauri/src/commands/ai_provider/cli_agent/mod.rs
@@ -525,6 +525,12 @@ async fn run_stream(
     // Whether a non-blank content delta has streamed yet — gates leading-blank
     // suppression so plain-text agents don't emit a stray newline before the answer.
     let mut seen_content = false;
+    // The full completed answer, accumulated from the exact non-thinking deltas
+    // we stream (post leading-blank suppression) — the same shape the renderer
+    // buffers. Persisted by `emit_done` so a renderer that missed frames or the
+    // terminal `done` event recovers it by polling, exactly like the cloud
+    // `stream::finish` path.
+    let mut answer = String::new();
 
     loop {
         // Race the next line read against a cancel poll so a cancellation mid-line
@@ -594,6 +600,7 @@ async fn run_stream(
                     }
                     seen_content = true;
                 }
+                answer.push_str(&text);
                 emit_event(
                     app,
                     AI_STREAM,
@@ -660,7 +667,7 @@ async fn run_stream(
     // `AiProvider::complete_with_usage`'s DEFAULT impl, which already reports
     // zero usage for any provider (like this one) that doesn't override it.
     super::record_usage(app, backend.id().as_str(), model, 0, 0, None);
-    emit_done(app, job_id);
+    emit_done(app, job_id, &answer);
     trace.end(status.and_then(|s| s.code()).map(|c| c as u16), true);
     Ok(())
 }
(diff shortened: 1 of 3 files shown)


saeedkolivand:
@coderabbitai review

saeedkolivand:
Thanks. This is the fix #802 thought it was: the completed text actually exists to fall back on now, think-stripped on both sides so the longer-wins comparison can't smuggle reasoning markup into a document. Both provider families inherit it from the shared seam. Merged.

saeedkolivand:

PR #829, opened 2026-07-22
fix: small advisory batch (rust)

saeedkolivand:
Final small-advisory batch from the 2026-07-21/22 PR-merge session — three Rust-side follow-ups. One commit per item.

## Item 1 — `perf(autopilot)`: skip duplicate note candidates within one run (#813 follow-up)

`run_notes_loop` de-duped a job's AI note only against **prior** runs. Within a single run, the same NEW job surfaced under two URL variants (same canonical key) bought a note for **each**; the later `merge_found_jobs` then collapsed them and discarded one — waste against the `ASSISTANT_NOTES_MAX` ceiling. Now a `HashSet` of canonical keys seen this run skips the second variant, charged **before** `charge_daily` so the duplicate also spares the shared per-provider ceiling.

- Test: `duplicate_url_variants_within_one_run_pay_for_only_one_note` (two variants → one provider call, one note).
- The oversized `autopilot_helpers` **inline test module was extracted to a sibling `tests.rs`** so `mod.rs` stays under the R8 hard LOC cap (the file was ~5 lines under the cap, so any addition tripped it; `test.rs`/`tests.rs` files are exempt from R8). Pure move, no test-behavior change.

## Item 2 — `fix(ai-generations)`: enforce one aggregate row per job url with upsert (#816 follow-up)

`save_application` released the store lock between `find_by_job_url` and the insert, so two concurrent saves for one job could both miss and both insert, **forking** the per-job aggregate.

- New migration `unique_job_url_aggregate`: a **partial** `UNIQUE(job_url) WHERE job_url != ''` index — so deliberately unlinked `''` rows (unusable raw urls) still coexist as separate rows. Forward-safe: collapses any pre-existing fork (keep the linked-or-newest row) **before** creating the index.
- `save_application` now **recovers** from the insert conflict by merging into the row the other writer created — one aggregate survives, the Rust merge stays the single source of merge truth (no merge logic duplicated into SQL).
- Tests: `unique_index_rejects_a_direct_duplicate_job_url`, `empty_job_url_rows_survive_the_unique_index` (the required empty-url multi-row case), `concurrent_saves_for_one_job_keep_a_single_aggregate` (6-thread race, invariant = one row, all Ok; stable ×5).
- Two `applications` delete/detach tests inserted duplicate-url child rows to exercise multi-row cleanup; they now use **distinct** urls per the new index (shipped in this commit so it stays green).

## Item 3 — `chore(applications)`: tighten the set-status test bound and trim a stale comment (#811 nits)

- `set_status`'s comment tail claiming "`.transaction()` needs `&mut Connection`, so call on `&mut *guard`" was stale (the code uses `guard.transaction()` via auto-ref) — trimmed.
- The contention test asserted `events.len() >= ITERS`; tightened to the **exact** count. Note: the exact count is `ITERS * 2 + 1` (2 threads × ITERS unconditional transitions + the single seed event appended at Application creation), not the review's `ITERS * 2` — the `>=` would otherwise still pass by luck; the `+1` seed is what the test's own chain-check already skips at index 0.

## Gates (all green)

- `cargo fmt --check`, `cargo clippy --all-targets --all-features --locked -- -D warnings` (exact CI command), `cargo check --all-targets --all-features --locked`, `cargo test --all-features` (2722 lib + 12 integration incl. architecture/R8), `pnpm typecheck` (13/13).
- Full `pnpm test`: one **unrelated, flaky** renderer test (`use-extension-bridge` `waitFor` timeout) failed once under full-suite load; it passes on the base and 6/6 in isolation with these changes. This diff is **Rust-only** (5 `.rs` files + the new `tests.rs`) and cannot affect a TS renderer test.

Source PRs: #813, #816, #811.


saeedkolivand:
@claude review

saeedkolivand:
Thanks. The partial unique index closes the aggregate fork for good while keeping the deliberately-unlinked empty-url rows intact, and the six-thread race test gives the recovery path real teeth. The review's note that the one-time migration keeps only the surviving fork rather than merging content is a fair point, worth a plainer sentence in the migration comment if anyone revisits it. Merged.

saeedkolivand:

PR #831, opened 2026-07-22
fix(prompts): never flag a mismatch against a blank language side

saeedkolivand:
## Summary
- Follow-up to #809's review: closes the asymmetric-blank guard gap where a known non-en language on one side (e.g. `"de"`) plus a blank/missing language on the other side (e.g. `""`) still produced a false-positive mismatch flag.
- Root cause: `toLanguage()` defaults a blank/missing value to `'en'` for downstream prompting, so the mismatch guard's `!== 'unknown'` check never saw the blank case — a blank `jobAdLanguage` normalized to `'en'`, and `"de" !== "en"` read as a genuine mismatch.
- Fix: gate `mismatch` on the raw (pre-normalization) blank check, so a blank side never counts as a mismatch regardless of what the known side is.
- Updated the one pre-existing test that encoded the old (buggy) expectation, and added a pinning test for `{resumeLanguage:"de", jobAdLanguage:""}` → no mismatch.

## Test plan
- [x] `pnpm -F @ajh/prompts test` — 522 tests green, including the new asymmetric-blank case
- [x] `TURBO_FORCE=1 pnpm typecheck` — whole graph green
- [x] `pnpm test` — whole graph, 373 files / 4057 tests green


saeedkolivand:
@claude review

saeedkolivand:
Thanks. A blank side masquerading as english and flagging phantom mismatches was subtle, checking the raw value before normalization is the right seam, and correcting the test that had enshrined the bug is the part that keeps it fixed. Merged.

saeedkolivand:

PR #833, opened 2026-07-22
fix(ui): stop rendering "0mo ago" for a 28-29 day old timestamp

### LINKED ISSUE #832: fix: relative time renders "0mo ago" for a 28-29 day old timestamp
## Describe the bug

`useFormatRelativeTime` renders **"0mo ago"** for any timestamp that is 28 or 29 days old.

The tier ladder is not contiguous: the week tier ends at `weeks < 4` (i.e. day 28), but the
month tier is floored on `days / 30`. Days 28 and 29 therefore fall past the week tier into
`monthsAgo` while `months` is still `0`.

https://github.com/saeedkolivand/ai-job-hunter-app/blob/main/apps/desktop/src/renderer/hooks/use-format-relative-time.ts#L58-L67

```ts
const days = Math.floor(hours / 24);
const weeks = Math.floor(days / 7);
const months = Math.floor(days / 30);
...
if (days < 7) return t(keys.daysAgo, { d: days });
if (weeks < 4) return t(keys.weeksAgo, { w: weeks });   // ends at day 28
return t(keys.monthsAgo, { m: months });                // months === 0 on days 28-29
```

The translation templates interpolate the count verbatim, so the zero is user-visible:
`"timeMonthsAgo": "{{m}}mo ago"` (en) and `"vor {{m}} Mo."` (de).

## Steps to reproduce

1. Have (or seed) a job posting whose `postedAt` is 28 days ago — job boards routinely keep
   listings ~30 days, so this is an ordinary row.
2. Open the **Jobs** page.

## Expected behaviour

`4w ago`.

## Actual behaviour

`0mo ago` (`vor 0 Mo.` in German).

Exact behaviour across the boundary:

| age | rendered |
|---|---|
| 27 days | `3w ago` |
| **28 days** | **`0mo ago`** |
| **29 days** | **`0mo ago`** |
| 30 days | `1mo ago` |

## Scope

Both namespaces are affected, so it shows up in more than one place:

- `jobs` — `JobsPage/index.tsx` (posting `postedAt`)
- `resumes.relativeTime` — `ApplicationDetailPage`, `GenerationCard`, `InteractionRow`
  (record `createdAt`)

## Environment

- **OS**: macOS (platform-independent — it is pure date arithmetic)

thejesh23:
Fixes #832

## What

`useFormatRelativeTime`'s tier ladder had a two-day gap: the week tier ended at `weeks < 4`
(day 28) while the month tier is floored on `days / 30`. Days 28 and 29 fell past the week
tier into `monthsAgo` with `months === 0`, so a month-old posting rendered **"0mo ago"**
(`"vor 0 Mo."` in de).

## How

Hand over to the month tier at the same 30-day boundary `months` is actually computed on:

```diff
 if (days < 7) return t(keys.daysAgo, { d: days });
-if (weeks < 4) return t(keys.weeksAgo, { w: weeks });
+if (days < 30) return t(keys.weeksAgo, { w: weeks });
 return t(keys.monthsAgo, { m: months });
```

| age | before | after |
|---|---|---|
| 27 days | `3w ago` | `3w ago` |
| 28 days | **`0mo ago`** | `4w ago` |
| 29 days | **`0mo ago`** | `4w ago` |
| 30 days | `1mo ago` | `1mo ago` |

Every other tier is byte-identical — `days < 7` still owns days 0-6, day 30 still starts
`1mo ago`. No translation strings change: `timeWeeksAgo` / `weeksAgo` already exist in every
locale.

Both namespaces were affected, so this shows up in more than one screen: `jobs` (posting
`postedAt` on the Jobs page) and `resumes.relativeTime` (`ApplicationDetailPage`,
`GenerationCard`, `InteractionRow`).

## Tests

Adds `use-format-relative-time.test.ts` — the hook had no test file:

- the week → month handover, pinning days 27/28/29/30/75;
- a sweep over days 1-400 asserting **no** tier ever interpolates a zero count. That is the
  property that was actually violated, so it guards the whole ladder rather than just this
  one boundary;
- the sub-week tiers and the falsy-timestamp case, unchanged;
- the `resumes.relativeTime` namespace.

I confirmed the suite fails on `main` with exactly the reported symptom
(`Expected "jobs.timeWeeksAgo:{"w":4}"` / `Received "jobs.timeMonthsAgo:{"m":0}"`) and passes
with the fix.

## Checklist

- [x] `pnpm typecheck` passes
- [x] `pnpm lint:strict` passes (no warnings)
- [x] `pnpm format:check` passes
- [x] Tests added

`TypeError: Cannot read properties of undefined (reading 'setItem')` (no `localStorage` in
this environment). I verified that by checking out clean `main` and re-running. The targeted
suite above and the lint/format/typecheck gates all pass; CI is authoritative.</sub>

### DIFF
--- a/apps/desktop/src/renderer/hooks/use-format-relative-time.test.ts
+++ b/apps/desktop/src/renderer/hooks/use-format-relative-time.test.ts
@@ -1,0 +1,57 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { useFormatRelativeTime } from './use-format-relative-time';
+
+/** Echoes the i18n key plus its interpolation values, so a test can assert both. */
+const t = ((key: string, vars?: Record<string, unknown>) =>
+  vars ? `${key}:${JSON.stringify(vars)}` : key) as never;
+
+const NOW = 1_750_000_000_000;
+const DAY = 24 * 60 * 60 * 1000;
+
+describe('useFormatRelativeTime', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(NOW);
+  });
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it('keeps the tiers contiguous across the week to month handover', () => {
+    // Days 28-29 used to fall through `weeks < 4` into the month tier, where
+    // `months = floor(days / 30)` is still 0 — rendering "0mo ago".
+    const format = useFormatRelativeTime(t);
+    const at = (days: number) => format(NOW - days * DAY);
+
+    expect(at(27)).toBe('jobs.timeWeeksAgo:{"w":3}');
+    expect(at(28)).toBe('jobs.timeWeeksAgo:{"w":4}');
+    expect(at(29)).toBe('jobs.timeWeeksAgo:{"w":4}');
+    expect(at(30)).toBe('jobs.timeMonthsAgo:{"m":1}');
+    expect(at(75)).toBe('jobs.timeMonthsAgo:{"m":2}');
+  });
+
+  it('never interpolates a zero count in any tier', () => {
+    const format = useFormatRelativeTime(t);
+
+    for (let days = 1; days <= 400; days += 1) {
+      expect(format(NOW - days * DAY)).not.toMatch(/:\{"[a-z]":0\}$/);
+    }
+  });
+
+  it('formats the sub-week tiers unchanged', () => {
+    const format = useFormatRelativeTime(t);
+
+    expect(format(NOW)).toBe('jobs.timeJustNow');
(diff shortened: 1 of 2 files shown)


saeedkolivand (approved):
Traced the boundary math myself rather than trusting the table: days 27-31 all land in the right tier now, and since the weeksAgo branch is only reachable at days >= 7, weeks can never be 0 there. Nice touch generalizing with the 1..400 property test, it guards the whole ladder instead of just the 28/29 seam. Also checked both en and de locale files for the week/month keys, all present with matching placeholders, and all four call sites go through this one hook.

Two minor things, neither blocking: the `t` mock cast `as never` in the test is a bit of a smell, and the flat 30-day month approximation (plus future timestamps collapsing to justNow) are pre-existing quirks this PR rightly leaves alone.

saeedkolivand:

PR #835, opened 2026-07-22
fix(scraper): delegate board location filtering to the central location_filter

thejesh23:
## Problem

Picking a city from the geocode suggestions returned **0 results** on the boards
that filter location client-side (The Muse, Comeet, GermanTechJobs). The location
box holds the picker's `"<city>, <country>"` label (`commands::geocoding` builds
it "so the UI only ever shows 'City, Country'"), but each board writes its own
shape — `"New York, NY"`, `"Munich, Bayern"`, `"Karl-Marx-Allee 1, Berlin"`. The
board-local filter required that **whole label** to appear as one contiguous
substring, which no board writes, so it dropped every row. Free-typed (unpicked)
input is one segment, which is why it survived manual testing.

## Fix (reworked per review)

The first revision added a board-local *segment-wise* matcher. Per review, this
version instead **removes board-local location filtering entirely** from the
three `supports_location() == false` boards and makes the engine's central
[`location_filter`](apps/desktop/src-tauri/src/scraping/engine/location_filter.rs)
the single authority on location — no second, divergent matcher.

Why this is correct and strictly better:

- These three boards are all non-location boards, so the engine **already** runs
  `location_filter` on their output, both as a live gate (`KeepItemByBoardFn`)
  and as a post-hoc safety net. The board-local location filter was redundant
  with it.
- The central filter is **diaeresis/exonym-aware** (München/Munich, Köln/Cologne
  via the curated table + DIN-5007-2 folding) and **conservative** (keeps
  remote / empty / unknown; a country-code-only request is inert). The raw
  board-local matcher had none of that and, running *before* the central filter,
  produced exactly the false drops the central filter exists to prevent (e.g. a
  `"Köln, Deutschland"` request against a `"Cologne"` feed).
- The original 0-results bug is still fixed: `location_spec()` feeds the picker's
  `"<city>, <country>"` label into the central filter, which **tokenises** it (it
  never required the whole label as one substring), so a geocode-picked city
  keeps its rows via the central filter alone.

## Changes

- `common.rs`: `matches_filters` is now **keyword-only** (drops the `location`
  param); `location_matches` is deleted.
- `germantechjobs`: `matches_gtj_filters` is keyword-only; the now-unused `loc`
  local is removed.
- The Muse / Comeet call sites pass query only.
- Tests across `common` / `themuse` / `germantechjobs`: the board-local location
  assertions are replaced with ones pinning that these boards **no longer drop on
  location** (delegated to the central filter, whose own exonym/diacritic
  behaviour is unit-tested in `engine::location_filter`).

Net: **−80 lines**, one authority instead of two.

## Known limitation (unchanged, now consistent across all non-location boards)

Because `location_spec()` stores the whole label in `city`, the central filter
also tokenises the country segment, so `"Munich, Germany"` keeps a
`"Frankfurt, Germany"` row. That is a property of the central filter shared by
**every** non-location board — not something this PR introduces (the old strict
board-local filter merely masked it for these three). Tightening it is a separate
change to `location_filter` / `location_spec`, out of scope here.

## Verification

- `cargo test -p ajh-tauri --lib scraping::` — **883 passed, 0 failed**.
- `cargo fmt --all -- --check` clean; `cargo clippy --lib` clean.

<sub>The full monorepo pre-push gate isn't runnable in my sandbox (the landing
build needs an install; Node 26 breaks jsdom `localStorage` in unrelated
frontend suites), so I verified the Rust crate directly as above. CI is
authoritative.</sub>

### DIFF
--- a/apps/desktop/src-tauri/src/scraping/boards/comeet/mod.rs
+++ b/apps/desktop/src-tauri/src/scraping/boards/comeet/mod.rs
@@ -265,11 +265,7 @@ impl Scraper for ComeetScraper {
         let mut out = Vec::new();
 
         for posting in parse_comeet_response(positions, company_uid.trim(), now) {
-            if !matches_filters(
-                &posting,
-                &input.query,
-                input.location.as_deref().unwrap_or(""),
-            ) {
+            if !matches_filters(&posting, &input.query) {
                 continue;
             }
             if let Some(ref on_item) = ctx.on_item {
--- a/apps/desktop/src-tauri/src/scraping/boards/common.rs
+++ b/apps/desktop/src-tauri/src/scraping/boards/common.rs
@@ -52,53 +52,27 @@ pub(crate) fn is_valid_dns_label_slug(slug: &str) -> bool {
         && !slug.ends_with('-')
 }
 
-/// Whether a board's own location text satisfies the requested location, for the
-/// boards that filter client-side.
+/// Client-side KEYWORD filter for boards with no server-side keyword search
+/// (The Muse, Comeet). Case-insensitive substring match on `title + company`
+/// for `query`; an empty query passes everything. Originally The Muse-local;
+/// extracted here once Comeet needed the identical filter instead of a second
+/// copy.
 ///
-/// `requested` is normally the geocode picker's `"<city>, <country>"` label
-/// (`commands::geocoding` builds it precisely "so the UI only ever shows
-/// 'City, Country'"), while a board writes its location in its own shape —
-/// `"New York, NY"`, `"Munich, Bayern"`, `"Karl-Marx-Allee 1, Berlin"`. Requiring
-/// the WHOLE label as one contiguous substring therefore matched nothing and
-/// zeroed the board for every picked city, so match SEGMENT-wise instead: any
-/// non-empty comma-separated part of the request appearing in `haystack` is
-/// enough.
-///
-/// Conservative by design — it errs toward keeping a posting, mirroring the
-/// engine's central [`crate::scraping::engine::location_filter`], which
-/// tokenizes the request and keeps on any token hit for the same reason. Both
-/// arguments must already be lowercased by the caller.
-pub(crate) fn location_matches(requested: &str, haystack: &str) -> bool {
-    requested
-        .split(',')
-        .map(str::trim)
-        .filter(|part| !part.is_empty())
-        .any(|part| haystack.contains(part))
-}
-
-/// Client-side query/location filter for boards with no server-side keyword
-/// search (The Muse, Comeet). Case-insensitive substring match on
-/// `title + company` for `query`; segment-wise case-insensitive match on
-/// `location` for `location` (see [`location_matches`]); an empty filter passes
-/// everything; both clauses AND-combine. Originally The Muse-local; extracted
-/// here once Comeet needed the identical filter instead of a second copy.
-pub(crate) fn matches_filters(
-    posting: &crate::scraping::types::JobPosting,
-    query: &str,
-    location: &str,
-) -> bool {
+/// Location is deliberately NOT filtered here. These boards do not consume the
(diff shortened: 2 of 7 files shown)


saeedkolivand (changes requested):
The root-cause analysis is right, and segment-wise matching does fix the literal "City, Country" vs board-label mismatch on both boards. But I don't think this can ship as is.

The main problem: location_matches is a raw case-insensitive contains with no diacritic or exonym handling, and the board-local filter drops rows before the engine's central filter ever sees them, so the München/Munich and Köln/Cologne table in scraping/engine/location_filter.rs can't rescue anything. The Nominatim call in commands/geocoding.rs isn't locale-pinned, so a request of "Köln, Deutschland" against a feed that says "Cologne" drops the row silently. That's the exact false-drop class location_filter.rs was built to prevent (its own tests name those cities), reintroduced on the DACH board this PR is fixing. None of the new fixtures use a spelling variant, so the tests don't catch it.

Second, matching on any segment means the country segment matches too: "Munich, Germany" now keeps a "Frankfurt, Germany" posting. The laxity technically pre-exists (location_spec() stuffs the whole label into city, so the central filter tokenizes the country as a needle as well), but the old strict filter was masking it, and this PR is what makes it user-visible.

The simplest fix, which the issue itself points at: drop board-local location filtering for these supports_location()==false boards entirely and let the central filter be the single authority, instead of maintaining a second divergent matcher. If you'd rather keep the local filter, route the segments through the same normalization location_filter.rs uses, require the city segment specifically rather than any segment, and add München/Munich plus a same-country-different-city case to the tests.

Minor extras: comeet shares matches_filters so its behavior changes with no test touched, and the helper has no minimum-segment-length guard like the central MIN_TOKEN_LEN.

thejesh23:
Pushed f570e73 taking the first option you laid out: dropped board-local location filtering for the three `supports_location() == false` boards and let the central `location_filter` be the single authority, rather than maintaining a second divergent matcher.

I traced the engine before changing direction to confirm it's safe: The Muse, Comeet and GermanTechJobs are all non-location boards, and `location_filter` already runs on their output both as the live gate (`KeepItemByBoardFn` at `engine/mod.rs`) and as the post-hoc net — so the board-local location clause was pure redundancy sitting *in front of* the exonym/diacritic-aware filter, which is exactly why the München/Munich and Köln/Cologne rescues couldn't fire. Deleting `location_matches` and making `matches_filters`/`matches_gtj_filters` keyword-only removes that.

The original 0-results bug stays fixed without a board-local matcher: `location_spec()` hands the picker's "City, Country" label to the central filter, which tokenises it rather than requiring the whole label as one substring, so a picked city keeps its rows.

On your two specific points:
- **Exonym false-drops** — now handled: location for these boards goes through the central filter's curated table + folding. No board-local `contains` to bypass it.
- **Country-segment over-match** ("Munich, Germany" keeping "Frankfurt, Germany") — you're right that it pre-exists in the central filter (`location_spec` stuffs the whole label into `city`, so "germany" becomes a needle). This PR no longer makes it *uniquely* visible on these boards — their behaviour is now identical to every other non-location board. I've called it out as a known limitation in the description; tightening it is a separate change to `location_filter`/`location_spec` and I didn't want to widen this PR's scope into it. Happy to open a follow-up if you'd like it fixed.

Also addresses the two minor points: comeet's shared filter now has explicit coverage via the keyword-only `matches_filters` tests in `common/test.rs`, and there's no longer a segment matcher needing a `MIN_TOKEN_LEN` guard — the central filter already enforces its own.

`cargo test scraping::` is 883/0, fmt + clippy clean.

saeedkolivand (approved):
Re-reviewed after f570e737: exactly the rework I asked for — board-local location matching deleted, the engine's central location_filter is the single authority, keyword-only filters remain, tests pin the no-drop-on-location behavior. CI green, 883 scraping tests pass. Approving over my earlier CHANGES_REQUESTED.

saeedkolivand:

PR #837, opened 2026-07-22
fix(ai): stop corrupting streamed text at a chunk boundary

### LINKED ISSUE #836: fix: streamed AI text is corrupted when a multi-byte character spans a chunk boundary
## Describe the bug

Streamed AI text is silently corrupted whenever a multi-byte character lands on a transport
read boundary: the character is replaced by replacement characters (`���`) in the live view
**and** in the saved document.

### Root cause

`commands/ai_provider/stream.rs` decodes every read independently:

```rust
match response.chunk().await {
    Ok(Some(bytes)) => {
        buf.push_str(&String::from_utf8_lossy(&bytes));
```

`reqwest::Response::chunk()` yields whatever the body decoder has, split at arbitrary byte
offsets — a chunked-transfer body surfaces as the socket reads land, and h2 DATA frames are
cut wherever the server's write buffer flushed. So a multi-byte character straddles two reads
on any reasonably long generation. `String::from_utf8_lossy` on a slice that ends (or begins)
mid-sequence replaces the partial bytes with `U+FFFD` and consumes them — nothing carries the
incomplete tail to the next read.

Trace, with an em dash (`—` = `E2 80 94`) cut across two reads:

| read | bytes | decoded |
|---|---|---|
| n | `…"content":"we deliver \xE2` | `…"content":"we deliver \u{FFFD}` |
| n+1 | `\x80\x94 fast"…` | `\u{FFFD}\u{FFFD} fast"…` |

The parser then json-parses that perfectly happily — `U+FFFD` is a legal character inside a
JSON string — and emits the delta `we deliver ��� fast`.

### Why it is not just cosmetic

The corrupted text reaches **both** sinks:

- `emit_delta` → the user's live cover letter;
- `answer.push_str(&piece.delta)` → `finish()` → `job_complete(… "text": …)`, so the mojibake
  is persisted as the job result and survives the renderer's poll fallback.

No error, no log, no indication anything went wrong.

### Scope

This is the single shared streaming loop behind **every** cloud adapter — `openai.rs`
(native OpenAI, every OpenAI-compatible gateway, and Ollama Cloud), `anthropic.rs`,
`gemini.rs`, `ollama.rs` — reached from `Completer::stream` → `generate_pipeline` and from
`Completer::stream_complete` (the extension bridge's `answer.assist`).

The same per-read decode is repeated at `ollama.rs` in the model-pull progress stream.

## Steps to reproduce

Generate a cover letter long enough to span several transport reads with any cloud provider.
Any of `—`, `’`, `“`, `”`, `é`, `ü`, `€` that happens to fall on a read boundary comes out as
`�`. It is intermittent by nature — it depends on where the server flushed.

Deterministically, at the loop level: feed the byte sequence `[b'a', 0xE2]` then
`[0x80, 0x94, b'b']`. Expected `a—b`; actual `a���b`.

## Expected behaviour

`a—b` — an incomplete trailing sequence is held back until the next read completes it.

## Actual behaviour

`a���b`, forwarded to the renderer and persisted as the finished document.

## Environment

thejesh23:
Fixes #836

## What

The shared streaming loop decoded every transport read on its own:

```rust
buf.push_str(&String::from_utf8_lossy(&bytes));
```

`reqwest::Response::chunk()` splits the body at arbitrary byte offsets, so a multi-byte
character routinely straddles two reads. `from_utf8_lossy` replaced **both** halves with
`U+FFFD` and consumed the partial bytes — nothing carried the incomplete tail forward.

The failure was silent: `U+FFFD` is legal inside a JSON string, so the frame still parsed and
the mojibake flowed to the live view **and** into `answer` → `finish()` → `result.text`,
meaning the renderer's poll fallback recovered the same corrupted document. Cover letters
routinely contain `—`, `’`, `“ ”`, `é`, `ü`, `€` — all 2-3 byte sequences.

This is the one loop behind **every** cloud adapter (openai / anthropic / gemini / ollama and
every OpenAI-compatible gateway), so all of them were affected.

## How

Add `push_utf8`, which holds an incomplete trailing sequence in a `carry` buffer until the
next read completes it:

```rust
fn push_utf8(buf: &mut String, carry: &mut Vec<u8>, bytes: &[u8]) {
    carry.extend_from_slice(bytes);
    loop {
        match std::str::from_utf8(carry) {
            Ok(text) => { buf.push_str(text); carry.clear(); return; }
            Err(e) => {
                let valid = e.valid_up_to();
                buf.push_str(std::str::from_utf8(&carry[..valid]).unwrap_or_default());
                match e.error_len() {
                    Some(n) => { buf.push(char::REPLACEMENT_CHARACTER); carry.drain(..valid + n); }
                    None => { carry.drain(..valid); return; }
                }
            }
        }
    }
}
```

Genuinely invalid bytes — a corrupt transfer, never a provider — still collapse to a single
`U+FFFD` and are skipped, so a malformed stream can never stall the loop. The `Err(_)`
`error_len()` distinction is what separates "incomplete, wait for more" from "invalid, move
on".

Applied at **all three** sites that decoded a read in isolation, so no path is left behind:

1. the production `stream_response` loop,
2. the test-only `drive_stream` core (kept in lockstep — it is the harness the tests use),
3. `ollama::pull`'s progress stream.

## Tests

Four cases through the real `drive_stream` core:

- an em dash cut **1|2** across two reads → `a—b`;
- the same char cut **2|1**, plus a 4-byte emoji dribbled **one byte per read** — the carry
  must survive an arbitrary number of reads that yield no complete character, not just one;
- genuinely invalid bytes (`0xFF`) still produce exactly one `U+FFFD` and decoding continues,
  pinning the no-stall guarantee;
- a body that **ends** mid-character still completes exactly once (no hang, held-back bytes
  dropped).

Verified the two split-char tests **fail** on `main` and pass with the fix. The other two pass
either way by design — they guard behaviour the fix must not change.

The pre-existing loop tests are untouched and still pass, including
`completes_once_on_end_of_body_without_sentinel` (Gemini's no-sentinel completion) and the
three cancellation tests.

## Checklist

- [x] `cargo test --all-features` — 2725 passed, 0 failed
- [x] `cargo test --test architecture` — 11 passed (incl. the R8 LOC cap)
- [x] Tests added

### DIFF
--- a/apps/desktop/src-tauri/src/commands/ai_provider/ollama.rs
+++ b/apps/desktop/src-tauri/src/commands/ai_provider/ollama.rs
@@ -668,8 +668,10 @@ pub async fn pull(app: &AppHandle, job_id: &str, model: &str) -> AppResult<()> {
     }
 
     let mut line_buf = String::new();
+    // Bytes from a read that ended mid-UTF-8-sequence — see `stream::push_utf8`.
+    let mut carry: Vec<u8> = Vec::new();
     while let Some(bytes) = response.chunk().await.map_err(|e| e.to_string())? {
-        line_buf.push_str(&String::from_utf8_lossy(&bytes));
+        super::stream::push_utf8(&mut line_buf, &mut carry, &bytes);
         // Walk by a `consumed` offset and drain the parsed prefix once after the
         // inner loop, instead of reallocating the whole tail per line (O(n²)).
         let mut consumed = 0;
--- a/apps/desktop/src-tauri/src/commands/ai_provider/stream.rs
+++ b/apps/desktop/src-tauri/src/commands/ai_provider/stream.rs
@@ -210,6 +210,52 @@ pub(super) fn strip_think_blocks(text: &str) -> String {
     out
 }
 
+/// Append one transport read to `buf` as UTF-8, holding back an incomplete
+/// trailing sequence in `carry` for the next read.
+///
+/// `reqwest::Response::chunk` splits the body at arbitrary byte offsets — a
+/// chunked-transfer body surfaces as the socket reads land, and h2 DATA frames
+/// are cut wherever the server flushed. So one multi-byte character (an em dash,
+/// a curly quote, an accented letter, an emoji) routinely straddles two reads.
+/// Decoding each read on its own with `String::from_utf8_lossy` replaced BOTH
+/// halves with `U+FFFD`, and since a replacement char is legal inside a JSON
+/// string the frame still parsed — the mojibake was forwarded to the renderer
+/// and persisted as the finished document, with no error anywhere.
+///
+/// Genuinely invalid bytes (a corrupt transfer, never a provider) still collapse
+/// to a single `U+FFFD` and are skipped, so a malformed stream can never stall
+/// the loop.
+pub(super) fn push_utf8(buf: &mut String, carry: &mut Vec<u8>, bytes: &[u8]) {
+    carry.extend_from_slice(bytes);
+    loop {
+        match std::str::from_utf8(carry) {
+            Ok(text) => {
+                buf.push_str(text);
+                carry.clear();
+                return;
+            }
+            Err(e) => {
+                let valid = e.valid_up_to();
+                // `valid_up_to()` is by definition a valid UTF-8 prefix.
+                buf.push_str(std::str::from_utf8(&carry[..valid]).unwrap_or_default());
+                match e.error_len() {
+                    // A truly invalid sequence: emit one replacement char, skip
+                    // it, and keep decoding the rest of this read.
+                    Some(n) => {
+                        buf.push(char::REPLACEMENT_CHARACTER);
+                        carry.drain(..valid + n);
+                    }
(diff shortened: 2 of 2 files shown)


saeedkolivand (approved):
Walked through push_utf8 by hand with an em dash split 1|2 and 2|1, and the 4-byte emoji fed one byte per read. The carry is bounded at 3 bytes and every loop iteration either drains or returns, so a malformed stream can't stall. All three chunk-decode sites are covered (stream_response, drive_stream, the ollama pull progress loop), and the cli_agent paths decode after reassembly so they were never affected. Held-back bytes are always >= 0x80, so an SSE delimiter can never get stuck in the carry and frame timing is unchanged.

Two observations, not blockers: at end of body a truncated final char is now dropped where main emitted U+FFFD. Dropping mojibake seems like the better call, but a doc comment on that choice wouldn't hurt. And the unwrap_or_default after valid_up_to is unreachable-but-defensive, fine as is.

saeedkolivand:

PR #839, opened 2026-07-22
fix(export): keep body paragraphs separate in a letter with no letterhead

### LINKED ISSUE #838: fix: cover letter with no letterhead exports as one run-on paragraph
## Describe the bug

A cover letter that does **not** begin with a name/contact letterhead renders as one run-on
wall of text — every paragraph break is lost — in both the exported PDF and the live preview.

### Root cause

`export/typst_engine/letter.rs`, in `parse_cover_letter`'s line loop:

```rust
// Skip lines that were part of the header (name + blank + contact echo).
if skip_lines < 3
    && (clean.is_empty()
        || clean.trim().trim_end_matches([',', '.']).to_lowercase() == name_lower
        || (!contact_md.is_empty() && contact_md.contains(clean.trim())))
{
    skip_lines += 1;
    continue;
}
```

`skip_lines` only advances when the arm *fires*, so the arm stays armed for the whole
document until three skippable lines have accumulated — it is not anchored to the leading
header block, despite the comment saying "lines that were part of the header".

Because `clean.is_empty()` **alone** satisfies the condition, any of the first three blank
lines in the document is consumed by the `continue` and never reaches the paragraph-flush
branch a few lines below:

```rust
// Empty line → paragraph break.
if clean.is_empty() {
    if !current_para_text.is_empty() {
        paragraphs.push(…);
        current_para_text.clear();
    }
    continue;
}
```

so `current_para_text` is never flushed and the next body line is space-joined onto it.

### Why it triggers in practice

The agent path produces exactly this shape: `agent/tools.rs`'s `COVER_LETTER_SYSTEM` asks for
the finished letter only — *"Output ONLY the finished letter — no preamble or commentary"* —
so there is no letterhead echo to burn the three skips before the body starts. That draft is
persisted verbatim by `save_cover_letter_handler` and exported from there.

It is also the natural hand-edited shape, since the template renders the letterhead from
`ContactProfile` and the parser deliberately drops any contact echo from the body.

The repo's own fixture `EN_LETTER` happens to spend all three skips on its letterhead
(name → contact → blank) *before* the body, which is why the existing tests pass.

## Steps to reproduce

Export (or preview) a cover letter whose text is:

```
Dear Hiring Manager,

First paragraph.

Second paragraph.

Third paragraph.

Sincerely,

Jane Smith
```

## Expected behaviour

Three separate body paragraphs.

## Actual behaviour

One paragraph: `First paragraph. Second paragraph. Third paragraph.`

Measured with `parse_cover_letter` directly:

| input | `model.body.len()` |
|---|---|
| repo fixture `EN_LETTER` (full letterhead) | 3 ✅ |
| opens at the salutation | **1** ❌ |
| date + salutation, no addressee block | **1** (expected 2) ❌ |

## Scope

Both output paths — `render_letter_pdf` (download) and `render_letter_svg_pages` (preview) —
go through `parse_cover_letter`, and the `.typ` templates iterate `data.body` as paragraphs,
so a lost split is lost in the final output.

## Environment

thejesh23:
Fixes #838

## What

`parse_cover_letter`'s header-skip arm was not anchored to the header:

```rust
if skip_lines < 3
    && (clean.is_empty() || …name… || …contact echo…)
{ skip_lines += 1; continue; }
```

`skip_lines` only advances when the arm fires, so it stays armed for the whole document until
three skippable lines accumulate. `clean.is_empty()` alone satisfies it, so **body** blank
lines got consumed by the `continue` and never reached the paragraph-flush branch — the next
body line was then space-joined onto the un-flushed paragraph.

A letter that opens straight at the salutation therefore rendered as one run-on block. That is
exactly what the agent produces (`COVER_LETTER_SYSTEM` asks for the letter only, "no preamble"),
and the natural hand-edited shape too, since the template renders the letterhead from
`ContactProfile` and the parser deliberately drops any contact echo from the body.

The fixture `EN_LETTER` happens to spend all three skips on its letterhead before the body,
which is why this was invisible to the existing tests.

## How

```diff
-if skip_lines < 3
+if !body_started
+    && skip_lines < 3
     && (clean.is_empty()
```

Safe by construction: `current_para_text` is only ever appended to in the `body_started`
branch, so no paragraph can be pending while this arm is reachable — the change can only
stop the arm from eating body blank lines, never alter header handling.

## Tests

Two cases in `letter.rs`'s test module:

- a letter opening at the salutation → **3** body paragraphs (and the salutation is still
  detected, so the change didn't just shift the problem);
- a letter with only a date + salutation ahead of the body — two pre-body lines, still not
  enough to burn the skip budget → **2** paragraphs.

Verified both **fail** on `main` with exactly the reported symptom (`left: 1, right: 3` and
`left: 1, right: 2`) and pass with the fix.

`parses_english_letter_all_fields`, `body_rich_text_preserves_bold` and
`all_caps_name_adopts_profile_casing` are unaffected — the full-letterhead fixture takes the
identical path as before.

## Checklist

- [x] `cargo test --all-features` — 2723 passed, 0 failed
- [x] `cargo test --test architecture` — 11 passed (incl. the R8 LOC cap)
- [x] Tests added

### DIFF
--- a/apps/desktop/src-tauri/src/export/typst_engine/letter.rs
+++ b/apps/desktop/src-tauri/src/export/typst_engine/letter.rs
@@ -324,7 +324,16 @@ pub(super) fn parse_cover_letter(
         let clean = strip_md_links(trimmed);
 
         // Skip lines that were part of the header (name + blank + contact echo).
-        if skip_lines < 3
+        //
+        // Bounded to the PRE-BODY zone by `!body_started`: `skip_lines` only
+        // advances when this arm actually fires, so on a letter with fewer than
+        // three leading header lines the arm stayed armed into the body and ate
+        // its blank lines — swallowing the paragraph breaks (the `clean.is_empty()`
+        // flush below never ran) and space-joining the whole letter into one
+        // run-on block. Safe: `current_para_text` is only ever appended to in the
+        // `body_started` branch, so nothing can be pending while this is reachable.
+        if !body_started
+            && skip_lines < 3
             && (clean.is_empty()
                 || clean.trim().trim_end_matches([',', '.']).to_lowercase() == name_lower
                 || (!contact_md.is_empty() && contact_md.contains(clean.trim())))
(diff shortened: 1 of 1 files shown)


saeedkolivand (approved):
The !body_started gate only ever narrows the old condition, so a letter that exhausts all 3 skips before the salutation behaves identically. That's why the letterhead fixtures stay green, and it makes the regression argument easy. I also grepped every current_para_text mutation site: the only append is inside the body_started branch, so the skip arm can never eat a pending paragraph. All three .typ templates plus the SVG preview go through this one parser.

One thing that deserves a follow-up issue rather than a change here: the DOCX renderers in apps/desktop/src-tauri/src/export/docx/mod.rs are a separate line-scanner with the same first-line-is-the-name assumption. A no-letterhead letter still misclassifies the salutation there, and when candidate_name is set the salutation text gets silently replaced. Out of scope per the issue's own scope note, but it's the same bug class.

saeedkolivand:

PR #841, opened 2026-07-22
fix(ui): flush a pending RichTextEditor edit on unmount instead of dropping it

### LINKED ISSUE #840: fix: RichTextEditor drops the last edit when the view is switched within the debounce window
## Describe the bug

Anything typed in the rich-text editor during the last ~200 ms before switching away from
**Edit** is silently discarded.

### Root cause

`packages/ui/src/components/RichTextEditor/RichTextEditor.tsx`:

```tsx
// Flush any pending debounce on unmount so the last edit is never lost.
useEffect(() => {
  return () => {
    if (debounceRef.current) clearTimeout(debounceRef.current);
  };
}, []);
```

The comment promises a **flush**; the body only `clearTimeout`s. `emit` is never called from
the cleanup — its only two call sites are the debounce timeout itself and `replaceSelection`.
So when the component unmounts with a timer pending, the pending markdown is dropped and
`onChange` never fires.

```tsx
onUpdate: ({ editor: ed }) => {
  setIsEmpty(ed.isEmpty);
  if (debounceRef.current) clearTimeout(debounceRef.current);
  debounceRef.current = setTimeout(() => emit(ed), ONCHANGE_DEBOUNCE_MS);  // 200ms
},
```

### Why the editor unmounts

`EditableOutput` renders the three views conditionally:

```tsx
{view === 'edit' ? ( … <RichTextEditor … /> … ) : view === 'source' ? ( … <TextArea … /> … )
```

so a Preview / Edit / Source `SegmentedControl` switch unmounts `RichTextEditor` outright.

## Steps to reproduce

1. Open a generated résumé or cover letter → **Edit**.
2. Type a few words — e.g. finish a sentence with `…reducing p99 latency by 60%`.
3. Within ~200 ms of the final keystroke, click **Preview** (or **Source**, or close the panel).
4. Click **Edit** again.

## Expected behaviour

The typed text is still there.

## Actual behaviour

The last edit is gone. The parent-owned canonical `value` still holds the pre-keystroke
markdown, so the preview/PDF renders stale text and re-entering Edit re-parses that stale
value. There is no error and no indication anything was lost.

The same loss happens on any unmount of the surrounding panel/modal inside the debounce
window — i.e. the very common "type, then immediately click away" pattern.

## Why nothing else catches it

No blur handler emits — `EditableOutput`'s `onBlur` only calls the caller's `onBlur?.()`, it
never reads the editor. `editor.destroy()` does not dispatch a transaction, and even if it
did, `onUpdate` only *re-schedules* the same debounce.

## Environment

thejesh23:
Fixes #840

## What

The unmount effect is commented *"Flush any pending debounce on unmount so the last edit is
never lost"*, but its body only cleared the timer — `emit` was never called from the cleanup,
so the pending markdown was discarded.

`RichTextEditor` is rendered conditionally (Preview / Edit / Source in `EditableOutput`), so
switching view unmounts it. Type, then click Preview within `ONCHANGE_DEBOUNCE_MS` (200 ms),
and the edit is gone with no error: the parent-owned `value` still holds the pre-keystroke
markdown, the preview/PDF renders stale text, and re-entering Edit re-parses that stale value.

## How

Capture the pending document synchronously in `onUpdate` and serialize **that** in the
cleanup:

```diff
+const pendingDocRef = useRef<PMNode | null>(null);

 onUpdate: ({ editor: ed }) => {
   setIsEmpty(ed.isEmpty);
+  pendingDocRef.current = ed.state.doc;
   if (debounceRef.current) clearTimeout(debounceRef.current);
-  debounceRef.current = setTimeout(() => emit(ed), ONCHANGE_DEBOUNCE_MS);
+  debounceRef.current = setTimeout(() => {
+    debounceRef.current = undefined;
+    const doc = pendingDocRef.current;
+    pendingDocRef.current = null;
+    if (doc) emitDoc(doc);
+  }, ONCHANGE_DEBOUNCE_MS);
 },
```

Two details worth flagging for review:

- **Why a doc snapshot rather than reading the editor in the cleanup.** A ProseMirror node is
  immutable, so keeping it is a pointer copy — and it means the flush never has to touch an
  editor Tiptap may already have destroyed by the time this component's cleanup runs.
- **No double-emit.** The timer path clears `pendingDocRef` *before* emitting, so an unmount
  immediately afterwards finds nothing left to send.

`emitDoc` is `useCallback([])`-stable, so `emit` stays stable and the existing
`useImperativeHandle(ref, …, [editor, emit])` dependency list is unaffected.

## Tests

New `RichTextEditor.unmountFlush.test.tsx`, driving **real** transactions through the real
editor via a toolbar command (typing into the ProseMirror surface needs layout APIs jsdom
does not implement; a toolbar command dispatches a genuine transaction, which is all
`onUpdate` needs):

- an edit made inside the debounce window is emitted when the component unmounts —
  **fails on `main`** (`expected "vi.fn()" to be called 1 times, but got 0 times`);
- the flush is value-**preserving** for an untouched document, so a plain view switch cannot
  mark it dirty with different content;
- a normally-fired debounce still emits exactly once, pinning the no-double-emit property.

Worth noting for the second test: the editor already schedules one emit on mount today —
loading the initial content is itself a transaction — so a mounted editor emits the unchanged
value after the debounce on `main` too. I verified that on unmodified `main` before writing
the assertion, so the test pins "never emits *different* content" rather than "never emits".

### CI determinism (updated)

The first revision of this suite was red on a slower CI runner even though it passed locally.
Two independent test-only causes, both now fixed (the production change is unchanged):

- **Mount-emit race.** Because loading the initial content schedules a debounce on mount, on a
  slow runner that mount timer could fire *before* the "nothing emitted yet" assertion under
  real timers. The suite now drains and clears that mount emit before making the edit under
  test, so the assertion only ever sees the edit's own timer.
- **Unhandled ProseMirror error.** jsdom implements no layout, so `Range.getClientRects` is
  `undefined` and ProseMirror's `coordsAtPos → singleRect` throws
  `target.getClientRects is not a function` from `scrollToSelection` on every transaction — one
  instance landing in the no-double-emit test. `Range.getClientRects`/`getBoundingClientRect`
  are now stubbed in the shared `vitest.setup.ts` (next to the existing `scrollIntoView` stub),
  so editor-driven tests no longer emit that unhandled error.

The three `unmountFlush` tests pass, with no unhandled errors, across repeated runs.

## Behavior note

Because the initial-content transaction populates `pendingDocRef`, a **fast mount → unmount**
inside the debounce window now fires one `onChange` even when the user typed nothing (previously
the cleanup only cleared the timer and emitted nothing). The emitted markdown round-trips to the
input value, so it is value-preserving — but any autosave or dirty-flag consumer downstream will
see that extra, same-value call and should treat a no-op change accordingly.

## Checklist

- [x] `pnpm typecheck` passes (`@ajh/ui`)
- [x] `eslint --max-warnings 0` clean on the touched files
- [x] `prettier` clean
- [x] Tests added

`TypeError: Cannot read properties of undefined (reading 'setItem')` (no `localStorage` in this
environment). Verified by checking out clean `main` and re-running. CI is authoritative.</sub>


saeedkolivand (changes requested):
The production change looks right: the pending doc is captured in a ref so the unmount flush can't close over a stale value, and both paths clear the timer before emitting so it can't double-fire. But the PR isn't mergeable in this state.

The new test file is red on CI at this exact commit: "flushes a debounced edit when the editor unmounts" fails with expected vi.fn() not to be called, but it was called with ["Hello"]. The suite drives real timers against the 200ms debounce, and Tiptap's initial-content load schedules a debounce on mount by itself, so on the slower runner that timer fires before the test's nothing-emitted-yet assertion. The "all 94 tests pass" note in the description only holds locally. Switch the suite to vi.useFakeTimers() and advance explicitly, or drain the mount-triggered debounce before starting the scenario under test.

Vitest also reports two unhandled "target.getClientRects is not a function" errors from ProseMirror's scrollToSelection during this file, and one of them lands in the "still emits exactly once" test, which is the test that's supposed to prove no double-emit. jsdom needs the scroll stubs before that assertion means anything.

Also worth a line in the description: because the initial-content transaction populates the pending ref, a fast mount and unmount now fires onChange even when the user typed nothing. Content should round-trip identically so it's probably harmless, but autosave or dirty-flag consumers will see extra calls.

(Not for this PR: replaceSelection already had a double-emit race on main, unchanged here.)

thejesh23:
Pushed 1d7a331 addressing the review. The production change is untouched — all three were in the test setup.

- **Mount-emit race.** Switched the suite to drain the mount-scheduled debounce first: after render it waits for that initial emit, `mockClear()`s it, and only then makes the edit under test. So the "nothing emitted yet" assertion sees only the edit's own 200ms timer, not the mount timer that was firing early on the slower runner. (I kept real timers — `userEvent` + `vi.useFakeTimers()` deadlocks the `user.click` await; draining is the option you offered and it removes the wall-clock dependence that mattered here.)
- **`getClientRects` errors.** Root cause is `Range.getClientRects` being `undefined` in jsdom, so `coordsAtPos → singleRect` throws on every `scrollToSelection`. Stubbed `Range.getClientRects`/`getBoundingClientRect` in the shared `vitest.setup.ts`, next to the existing `scrollIntoView` stub. The no-double-emit test now runs with no unhandled errors; verified across repeated runs.
- **Description.** Added a "Behavior note": the initial-content transaction populates `pendingDocRef`, so a fast mount→unmount now fires one same-value `onChange` where before it emitted nothing — value-preserving, but autosave/dirty-flag consumers will see the extra call.

On the "all 94 tests pass" line — you're right that it only held locally; corrected in the description. I can't run the full `pnpm test` gate in my sandbox (Node 26 breaks jsdom `localStorage` — the same 29 `theme.test.ts` failures show up on unmodified `main`), so I verified the `ui` package directly: `unmountFlush` + `Toolbar.linkSuggestions` green ×3, `tsc --noEmit` clean, eslint clean on the touched files. CI is authoritative.

Noted the `replaceSelection` double-emit race on `main` as out of scope for this PR.

saeedkolivand (dismissed):
Re-reviewed after 1d7a3317: test suite is now deterministic (mount-emit drained, Range stubs guarded in shared setup) and CI is fully green. Production change unchanged and correct — the ref handoff can't double-emit. Approving over my earlier CHANGES_REQUESTED.

saeedkolivand:

PR #843, opened 2026-07-22
fix(ui): ignore a superseded geocode response

### LINKED ISSUE #842: fix: a slow geocode response can overwrite the suggestion list for a newer query
## Describe the bug

The location autocomplete can show suggestions that do not match what is in the search box,
and can silently reset the keyboard highlight mid-selection.

### Root cause

`packages/ui/src/hooks/useGeocoding.ts`:

```ts
useEffect(() => {
  const trimmed = query.trim();
  if (trimmed.length < 2) { setSuggestions([]); return; }
  if (timerRef.current) clearTimeout(timerRef.current);
  timerRef.current = setTimeout(() => {
    fetchRef.current(trimmed)
      .then((s) => { setSuggestions(s); setActiveIndex(-1); })   // ← unconditional
      .catch(() => {});
  }, 300);
  return () => { if (timerRef.current) clearTimeout(timerRef.current); };
}, [query]);
```

The cleanup cancels only the **pending timer**. Once a fetch has actually started, nothing
cancels it and nothing marks its reply stale — so `setSuggestions(s)` runs for whichever
request resolves last, not for the current query.

The 300 ms debounce does not serialize the requests: it only limits how often one is
*started*. Two are in flight whenever the user resumes typing after a pause longer than the
debounce, which is the normal rhythm for typing a city name.

### Timeline

With `api.geocode.suggest` (IPC → Rust → Nominatim; latency ranges from ~40 ms cached to
seconds, since Nominatim is rate-limited to ~1 req/s):

| t | event |
|---|---|
| 0 ms | user types `berl` → timer scheduled |
| 300 ms | timer fires → **fetch A(`berl`)** starts (uncached, ~2 s) |
| 400 ms | user types `berlin` → effect re-runs; cleanup clears an already-fired timer (no-op) |
| 700 ms | timer fires → **fetch B(`berlin`)** starts (cached, ~40 ms) |
| 740 ms | B resolves → correct list shown ✅ |
| 2300 ms | A resolves → **overwrites with the `berl` list** ❌ and resets `activeIndex` |

## Steps to reproduce

1. Open any location autocomplete (Jobs → Location, autopilot Step: Target, contact profile,
   education/experience steps).
2. Type a partial city, pause ~½ s, then finish typing it.
3. Occasionally the dropdown ends up showing results for the partial query. Arrowing through
   the list while this happens also loses the highlight.

## Expected behaviour

The list always corresponds to the current query text; a superseded request is discarded.

## Actual behaviour

A slower earlier request overwrites the newer results, and `setActiveIndex(-1)` drops the
user's keyboard selection.

## Scope

`LocationInput` renders `suggestions` straight through to `LocationDropdown` with no guard of
its own, and every production call site injects a real network-backed fetcher:
`ScrapeForm/ScrapeFilters.tsx`, autopilot `wizard-steps/StepTarget/index.tsx`,
`ContactProfileForm/index.tsx`, `StepEducation/index.tsx`, `StepExperience/index.tsx`.

## Environment

thejesh23:
Fixes #842

## What

`useGeocoding`'s cleanup cancels only the **pending timer**. Once a fetch has started nothing
cancels it and nothing marks its reply stale, so `setSuggestions(s)` ran for whichever request
resolved last rather than for the current query.

The 300 ms debounce does not serialize requests — it only bounds when one *starts*. Two are in
flight whenever the user resumes typing after a pause longer than the debounce, the normal
rhythm for typing a city name. When the slower older one wins, the dropdown shows results for
a query the user has already typed past, and `setActiveIndex(-1)` wipes their keyboard
highlight mid-selection.

## How

```diff
 useEffect(() => {
+  let cancelled = false;
   const trimmed = query.trim();
   …
     .then((s) => {
+      if (cancelled) return;
       setSuggestions(s);
       setActiveIndex(-1);
     })
   return () => {
+    cancelled = true;
     if (timerRef.current) clearTimeout(timerRef.current);
   };
 }, [query]);
```

React always runs the previous effect's cleanup before the next run, so this also covers the
`< 2 chars` early-return path — which otherwise let an in-flight request repopulate a list the
hook had just cleared. It also removes a setState-after-unmount.

## Tests

Two cases added to `useGeocoding.test.ts`:

- a slow `Be` request resolving **after** a fast `Berlin` one must not overwrite the list;
- an in-flight request must not repopulate the list after the query drops below 2 characters
  (the early-return path).

Verified both **fail** on `main` with exactly the reported symptom
(`expected [ { display: 'Belgrade, Serbia' } ] to deeply equal [ { display: 'Berlin, Germany' } ]`
and `expected [ { display: 'Berlin, Germany' } ] to deeply equal []`) and pass with the fix.
All 10 tests in the file pass.

## Checklist

- [x] `pnpm typecheck` passes (`@ajh/ui`)
- [x] `eslint --max-warnings 0` clean on the touched files
- [x] `prettier` clean
- [x] Tests added

`TypeError: Cannot read properties of undefined (reading 'setItem')` (no `localStorage` in this
environment). Verified by checking out clean `main` and re-running. CI is authoritative.</sub>

### DIFF
--- a/packages/ui/src/hooks/useGeocoding.test.ts
+++ b/packages/ui/src/hooks/useGeocoding.test.ts
@@ -31,6 +31,65 @@ describe('useGeocoding', () => {
     expect(result.current.activeIndex).toBe(-1);
   });
 
+  it('ignores a slow superseded response that resolves after a newer one', async () => {
+    // The 300ms debounce only bounds when a request STARTS. Resume typing after
+    // a pause and two are in flight; if the older (slower) one wins the race it
+    // used to overwrite the list with suggestions for a stale query.
+    const fetcher = vi.fn((q: string) =>
+      q === 'Be'
+        ? new Promise((resolve) =>
+            setTimeout(() => resolve([{ display: 'Belgrade, Serbia' }]), 2000)
+          )
+        : Promise.resolve([{ display: 'Berlin, Germany' }])
+    ) as unknown as (q: string) => Promise<{ display: string }[]>;
+
+    const { result, rerender } = renderHook(({ q }) => useGeocoding(q, fetcher), {
+      initialProps: { q: '' },
+    });
+
+    rerender({ q: 'Be' });
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(350); // fires the slow 'Be' fetch
+    });
+
+    rerender({ q: 'Berlin' });
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(350); // fires + resolves the fast 'Berlin' fetch
+    });
+    expect(result.current.suggestions).toEqual([{ display: 'Berlin, Germany' }]);
+
+    // Now let the stale 'Be' request land.
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(2000);
+    });
+    expect(result.current.suggestions).toEqual([{ display: 'Berlin, Germany' }]);
+  });
+
+  it('ignores an in-flight response after the query drops below 2 characters', async () => {
+    // The short-query branch returns early, but React still runs the previous
+    // cleanup first — so the superseded request must be flagged there too,
+    // otherwise it repopulates a list the hook just cleared.
+    const fetcher = vi.fn(
+      () =>
+        new Promise((resolve) => setTimeout(() => resolve([{ display: 'Berlin, Germany' }]), 2000))
+    ) as unknown as (q: string) => Promise<{ display: string }[]>;
+
+    const { result, rerender } = renderHook(({ q }) => useGeocoding(q, fetcher), {
+      initialProps: { q: '' },
+    });
+
+    rerender({ q: 'Berlin' });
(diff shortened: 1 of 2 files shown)


saeedkolivand (approved):
The per-effect cancelled flag is the right shape here. Better than comparing query strings (breaks when the user retypes the same text) and better than AbortController (the injected fetcher can be Tauri IPC, which has no abort semantics). React runs the previous cleanup synchronously before the next effect, so there's no microtask window for a stale resolve to sneak through. The catch path never touches state, so a stale rejection can't clear a fresh list either. And since this hook lives in packages/ui, the injectable-fetcher design is the right seam, React Query was never an option there.

One suggestion: add a test for the backward transition, e.g. type "berlin", backspace to "berl", let the berlin fetch resolve late. The guard is keyed to effect identity so it should hold, but right now only forward typing is pinned. Fine to do here or in a follow-up.

saeedkolivand:

PR #845, opened 2026-07-22
fix(scraper): keep distinct Hacker News job ids apart in cross-source dedup

### LINKED ISSUE #844: fix: distinct Hacker News job postings collapse into one row in cross-source dedup
## Describe the bug

Text-only Hacker News job postings all collapse into a **single** row after a scrape — only
one of them survives cross-source dedup.

### Root cause

`normalize_job_url` rebuilds the query through `retain_identifying_params`, which returns `""`
for every host without an allowlist entry — so the whole query is discarded. The allowlist has
only Indeed:

```rust
fn identifying_query_params(host: &str) -> &'static [&'static str] {
    if host == "indeed.com" || host.ends_with(".indeed.com") {
        &["jk"]
    } else {
        &[]
    }
}
```

But `scraping/boards/ycombinator/mod.rs` builds a URL that puts the id **in the query**
whenever an HN job story has no external link (`HnItem.url == None` — the fallback exists
precisely because that is common for text-only "X is hiring" posts):

```rust
let url = item
    .url
    .clone()
    .unwrap_or_else(|| format!("https://news.ycombinator.com/item?id={}", item.id));
```

So every such posting normalizes to the same key:

```
https://news.ycombinator.com/item?id=44100001  ->  https://news.ycombinator.com/item
https://news.ycombinator.com/item?id=44100002  ->  https://news.ycombinator.com/item
```

and `dedup_cross_source` (which runs unconditionally on the scrape result) collapses them.

### The allowlist's own comment already states the rule being violated

> Keep in sync with the canonical URL builders … that place a job id in the query string:
> currently only Indeed.

`boards::ycombinator` is a second such builder. The aggregator's own `canonical_url` documents
the same hazard from the other direction — *"some boards encode the job id in query params, so
stripping would merge distinct jobs."*

## Steps to reproduce

1. Scrape the **Hacker News / Y Combinator** board when the current `jobstories` feed contains
   two or more text-only posts (no external link).
2. Only one of them appears in the results.

Or directly:

```rust
dedup_cross_source(vec![
    posting("https://news.ycombinator.com/item?id=44100001", "Backend Engineer", "Acme"),
    posting("https://news.ycombinator.com/item?id=44100002", "Frontend Engineer", "Beta"),
])
```

## Expected behaviour

2 postings — different jobs, different ids.

## Actual behaviour

1 posting.

## Note

The streamed `on_item` copies are **not** deduped, so the UI cache and the returned/persisted
set disagree: the job flashes into the list during the scrape and then disappears.

## Environment

thejesh23:
Fixes #844

## What

`identifying_query_params` allowlisted only Indeed, so `normalize_job_url` discards the entire
query for every other host. `boards::ycombinator` puts the job id **in the query** whenever an
HN job story has no external link:

```
https://news.ycombinator.com/item?id=44100001  ->  https://news.ycombinator.com/item
https://news.ycombinator.com/item?id=44100002  ->  https://news.ycombinator.com/item
```

so `dedup_cross_source` collapsed every text-only HN posting into one row.

The allowlist's own doc comment already stated the rule it was violating — *"Keep in sync with
the canonical URL builders … that place a job id in the query string"* — and the aggregator's
`canonical_url` documents the same hazard from the other direction.

## How

Add the `news.ycombinator.com` → `id` entry on **both** sides of the lockstep pair:

- `applications/mod.rs` — `identifying_query_params`
- `renderer/features/jobs/lib/canonical-job-key.ts` — `identifyingQueryParams`

Tracking params on the same id still collapse (only `id` is retained, in the allowlist's fixed
order), so the dedup this allowlist exists to enable is unchanged.

I also tightened the doc comment on the Rust side: the fix pushed `applications/mod.rs` to
1401 LOC, over the R8 hard cap of 1400. It now sits at **1399** and
`cargo test --test architecture` passes.

## Tests

- `scraping/engine/test.rs` — two distinct HN ids must not collapse **and** the same id with a
  tracking param must still collapse, so the test pins both directions rather than just
  loosening the key.
- `canonical-job-key.test.ts` — the mirrored assertions, including param-order independence.

Verified both **fail** on `main` with exactly the reported symptom (Rust `left: 1, right: 2`;
TS `expected 'https://news.ycombinator.com/item' not to be 'https://news.ycombinator.com/item'`)
and pass with the fix.

## Known residual (not fixed here)

JSearch rows that fall back to `job_google_link` put their identity in the URL **fragment**,
which `normalize_job_url` drops unconditionally — an allowlist entry cannot address that. Left
alone deliberately rather than widened into this PR.

## Checklist

- [x] `cargo test --all-features` — 2722 passed, 0 failed
- [x] `cargo test --test architecture` — 11 passed (incl. the R8 LOC cap; file now 1399/1400)
- [x] `eslint --max-warnings 0` + `prettier` clean on the touched TS
- [x] Tests added on both sides of the lockstep pair

### DIFF
--- a/apps/desktop/src-tauri/src/applications/mod.rs
+++ b/apps/desktop/src-tauri/src/applications/mod.rs
@@ -1280,13 +1280,16 @@ pub fn normalize_job_url(url: &str) -> String {
 
 /// Per-host allowlist of *identifying* query params that must survive normalization
 /// (every other query param — utm_*, ref, tracking — is dropped, and hosts absent
-/// here drop the entire query). Keep in sync with the canonical URL builders in
-/// `scraping::scrape_url` that place a job id in the query string: currently only
-/// Indeed (`/viewjob?jk=<id>`). LinkedIn et al. put the id in the PATH, so they need
-/// no entry here and normalize exactly as before.
+/// here drop the entire query, so a host whose id lives in the QUERY collapses to
+/// one key for every job). Keep in sync with every canonical URL builder that does
+/// that: Indeed (`/viewjob?jk=<id>`) and Hacker News (`/item?id=<id>`, built by
+/// `boards::ycombinator` when a job story has no external url). Path-based ids
+/// (LinkedIn et al.) need no entry.
 fn identifying_query_params(host: &str) -> &'static [&'static str] {
     if host == "indeed.com" || host.ends_with(".indeed.com") {
         &["jk"]
+    } else if host == "news.ycombinator.com" {
+        &["id"]
     } else {
         &[]
     }
--- a/apps/desktop/src-tauri/src/scraping/engine/test.rs
+++ b/apps/desktop/src-tauri/src/scraping/engine/test.rs
@@ -3242,6 +3242,52 @@ fn dedup_posting_with_extra(
 }
 
 #[test]
+fn dedup_cross_source_keeps_distinct_hacker_news_job_ids_apart() {
+    // `boards::ycombinator` falls back to `news.ycombinator.com/item?id=<id>`
+    // whenever an HN job story carries no external url — the id lives in the
+    // QUERY. `normalize_job_url` drops the whole query for any host without an
+    // allowlist entry, so every such posting used to normalize to the bare
+    // `/item` path and collapse into one row.
+    let out = dedup_cross_source(vec![
+        dedup_posting(
+            "ycombinator",
+            "https://news.ycombinator.com/item?id=44100001",
+            "Backend Engineer",
+            "Acme",
+            None,
+        ),
+        dedup_posting(
+            "ycombinator",
+            "https://news.ycombinator.com/item?id=44100002",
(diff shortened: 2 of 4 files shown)


saeedkolivand (approved):
Root cause checks out: text-only HN postings all normalize to a bare /item once the query is stripped, and the allowlist had no HN entry. The fix is a purely additive else-if on both sides of the Rust/TS pair, and I hand-traced retain_identifying_params with an id plus a tracking param, true duplicates still collapse.

Two notes for the description rather than the code. Pre-fix HN application rows keep the old bare job_url and will never re-match a re-scraped posting. That's the same trade-off #568 shipped for Indeed's jk with no backfill, so it's established practice here, but it'd be good to state it as a "Known residual" the way you did for the JSearch fragment case. And applications/mod.rs is now one line under the file-size cap, so whoever touches it next should bring a split plan.

saeedkolivand:

PR #857, opened 2026-07-22
ci: bump codeql-action init to 4.37.3 in lockstep with analyze

saeedkolivand:
## Why

#850 (merged) bumped only the `analyze` pin to 4.37.3 while `init` stayed at 4.37.0. The init step writes a config stamped with its own version and the newer analyze refuses to load it, so every CodeQL run on main now fails with:

> Loaded a configuration file for version '4.37.0', but running version '4.37.3'

This was the exact failure already visible on #850's own PR checks before merge.

## What

- `.github/workflows/security.yml`: bump `codeql-action/init` to the same 4.37.3 SHA as `analyze` (comments now carry the exact version).
- `.github/dependabot.yml`: group `github/codeql-action*` so future bumps arrive as a single PR instead of drifting apart again.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -73,4 +73,5 @@ updates:
     groups:
       codeql-action:
         patterns:
-          - github/codeql-action*
+          - github/codeql-action
+          - github/codeql-action/*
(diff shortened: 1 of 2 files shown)


saeedkolivand:

PR #865, opened 2026-07-22
fix(export): bump typst family to 0.15.1 in lockstep

saeedkolivand:
## Why
PR #851 bumped only `typst-layout` to `=0.15.1`, leaving the rest of the typst family (`typst`, `typst-pdf`, `typst-svg`, `typst-render`) pinned at `=0.15.0`. That split the family across two point releases — against the in-file pin comment ("exact pin kept in lockstep with the rest of the family") and the repo's hard rule that the typst family must bump together.

## What
- `apps/desktop/src-tauri/Cargo.toml` — bump every remaining typst-family exact pin `=0.15.0 → =0.15.1`:
  - `typst`
  - `typst-pdf`
  - `typst-svg`
  - `typst-render` (dev-dependency, showcase-banner rasteriser)
  - (`typst-layout` was already `=0.15.1` from #851)
- `apps/desktop/src-tauri/Cargo.lock` — regenerated via `cargo update --precise 0.15.1`; transitive `typst-eval`, `typst-html`, `typst-realize` also advance 0.15.0 → 0.15.1. Every `typst-*` crate in the lockfile now resolves to 0.15.1.
- Comments left intact and accurate (they already assert lockstep / matching pins).

All four family members have a published 0.15.1 release, so `cargo update` resolved cleanly (no forcing).

## Test evidence
`cargo test` in `apps/desktop/src-tauri` (delayload fix present on main):
- Full suite: **2741 passed, 26 ignored, 0 failed** (5 suites, exit 0).
- Export subset (`cargo test export::`): **264 passed, 3 ignored, 0 failed** (exit 0) — covers the typst PDF/SVG render path (valid-PDF, per-template render, multipage sidebar, ATS reading-order, contact-link annotations, accent fallback) and the docx path.
- **Golden status: no drift.** The export tests are structural/property assertions (not committed byte-diff snapshot fixtures); typst 0.15.1 shifted no layout invariant — all passed and the working tree shows **only** `Cargo.toml` + `Cargo.lock` changed (no snapshot files regenerated).
- Re-ran the export subset against the post-merge tree (main's docx-rs/tokio bumps that the pre-push hook merged in): still 264 passed / exit 0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #866, opened 2026-07-22
ci: group typst family bumps in dependabot

saeedkolivand:
## Why

The typst crates are exact-pinned and must move in lockstep. Dependabot treats each crate as a separate dependency, so it proposed them one at a time — #851 merged a typst-layout-only bump and split the pins, which #865 had to repair. #858/#860 were the next piecemeal bumps in the queue.

## What

Add a `typst` group (patterns `typst`, `typst-*`) to the cargo ecosystem in `.github/dependabot.yml`, mirroring the codeql-action group added in #857. Future family releases arrive as one PR.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -57,6 +57,15 @@ updates:
     labels:
       - dependencies
 
+    # The typst family (typst, typst-pdf, typst-svg, typst-render,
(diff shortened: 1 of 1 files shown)


saeedkolivand:

PR #874, opened 2026-07-23
fix(ai-generations): make the mismatch merge follow a language-bearing save

### LINKED ISSUE #873: fix(ai-generations): mismatch flag latches true and a correction can never clear it
## Describe the bug

`merge_application` (`apps/desktop/src-tauri/src/ai_generations/mod.rs:535`) OR-accumulates the `mismatch` flag, so once any generation sets it `true` it can never be cleared:

```rust
mismatch: incoming.mismatch || existing.mismatch,
```

Every other scalar in the merge is `pick` (incoming wins only when it carries content) or latest-wins. `mismatch` is the lone `||`. `mismatch` is derived from the résumé/job-ad **language pair**, so:

- A **corrected regeneration** — e.g. the user re-runs with an English résumé against an English ad, so the new save carries `mismatch = false` — cannot clear the stale warning: `false || true = true`.
- Conversely `mismatch` would already be wrong under plain latest-wins, because a **content-less save** (answers-only / interview-questions-only) carries no language pair and its `mismatch` defaults to `false`, which would then wipe a real prior verdict.

## Steps to reproduce

1. Generate a cover letter with a German résumé against an English ad → the aggregate records `mismatch = true` (the "languages differ / do not translate" warning).
2. Re-generate with an English résumé (`mismatch = false`).
3. The aggregate still reads `mismatch = true`; the stale warning persists.

## Expected behaviour

A later language-bearing generation's verdict wins (so a correction can clear it), while a save that carries no language pair leaves the flag alone.

## Actual behaviour

The flag latches `true` forever after the first mismatch.

## Environment

- **OS**: any
- **App version**: `main` @ `395a895`
- **Ollama version**: n/a
- **Model in use**: any

## Additional context

This was noted in pass 1 (`aijh-011`) and held back only because it shared `ai_generations/mod.rs` with `aijh-008` (#815/#816, since merged). A PR is ready that follows the incoming `mismatch` only when the save actually computed the language pair — mirroring how the language fields themselves merge via `pick`.

thejesh23:
## Summary

`merge_application` OR-accumulated the language-`mismatch` flag, so a corrected regeneration could never clear a stale "languages differ" warning. Fixes #873.

## Type of change

- [x] `fix` — Bug fix

## Changes

- `mismatch` now follows `incoming.mismatch` **only when the incoming save carries the language pair** (both language fields non-empty), else keeps `existing.mismatch`.
- This mirrors how `resume_language` / `job_ad_language` / `target_language` themselves merge (`pick` — incoming wins only when non-empty), so `mismatch` stops being the lone `||` outlier.
- A corrected regeneration (`mismatch=false`, both languages present) now clears the warning; an answers-only / interview-only save (blank languages, default `mismatch=false`) no longer clobbers a real prior verdict.

## Testing

- [x] `cargo test --lib ai_generations` — 28 passed, 0 failed
- [x] `cargo test --test architecture` — 11 passed (R8 cap included)

Regression test covers both directions. Against `main`:

```
a language-bearing save with mismatch=false must clear the stale warning
test result: FAILED
```

## Checklist

### DIFF
--- a/apps/desktop/src-tauri/src/ai_generations/mod.rs
+++ b/apps/desktop/src-tauri/src/ai_generations/mod.rs
@@ -523,6 +523,12 @@ fn merge_application(
     incoming: AiGenerationRecord,
 ) -> AiGenerationRecord {
     let pick = |inc: String, ex: String| if inc.trim().is_empty() { ex } else { inc };
+    // `mismatch` is derived from the resume/jobAd language pair, so it is only
+    // meaningful when the incoming save actually carries that pair. An
+    // answers-only / interview-only save leaves the language fields blank and its
+    // `mismatch` defaults to `false` — not a real verdict.
+    let incoming_has_languages =
+        !incoming.resume_language.trim().is_empty() && !incoming.job_ad_language.trim().is_empty();
     AiGenerationRecord {
         id: existing.id,
         created_at: existing.created_at,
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand (approved):
Vetted by domain review: merge logic fix confirmed correct (the || latch on mismatch is real and the language-pair gate is the right repair), new test pins the bug, no out-of-scope changes. Red CI diagnosed as the known use-extension-bridge renderer flake (sighting #3), unrelated to this Rust-only diff — re-run triggered.

saeedkolivand:

PR #876, opened 2026-07-23
fix(export): do not consume a letter's salutation as the letterhead name

thejesh23:
## Summary

Both DOCX letter renderers consumed the first non-blank line as the letterhead name whatever it was, so a letter opening at the salutation lost its "Dear …" line and rendered the body in the muted addressee style. Fixes #875.

## Type of change

- [x] `fix` — Bug fix

## Changes

- Guarded the name block with `!is_salutation && !is_signoff` (both already computed just above it), so a salutation-first, letterhead-less letter falls through to the salutation arm — which is the only place `header_done` / `in_body` get set.
- Applied to **both** copies of the block: `generate_cover_letter_docx_classic` (`:96`, Classic) and `generate_cover_letter_docx_layout` (`:321`, Refined/Banded).

## Testing

- [x] `cargo test --lib export::docx` — 24 passed, 0 failed
- [x] `cargo test --test architecture` — 11 passed
- [x] `cargo clippy --all-targets --all-features --locked -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] Added tests for new logic

New test opens a letterhead-less letter (`meta.candidate_name` set so the drop is observable) and asserts `"Dear Hiring Manager"` survives, across all three layouts. Against `main`:

```
Classic: the salutation was consumed as the letterhead name
test result: FAILED
```

## Checklist

- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org)
- [x] No `console.log` left in source code
- [x] No `@ts-ignore` added without explanation
- [x] New public APIs are documented (no new public API)

PR #878, opened 2026-07-23
fix(export): keep a subject-line-first letterhead-less letter's subject line

### LINKED ISSUE #877: fix(export): docx letter with a subject-line-first, letterhead-less opening still loses its first line
Follow-up found while vetting #876: the first-line-as-letterhead-name guard in apps/desktop/src-tauri/src/export/docx/mod.rs excludes is_salutation/is_signoff but not is_subject_line, so a letter that opens with a subject/reference line (e.g. a German Betreff: line) and has no letterhead name still gets that line consumed as the candidate name. Pre-existing behavior, not a regression from #876 — needs a second guard clause (!is_subject_line) in both renderer arms plus a pinning test.

saeedkolivand:
Extends #876's name-block guard in both DOCX letter renderer arms with `!is_subject_line`, so a letter opening with a subject/reference line (e.g. a German `Betreff:` line) and no letterhead name no longer has that line consumed as the candidate name. Later subject-line call sites reuse the computed binding instead of recomputing.

Pinning test covers all three letter layouts (Classic/Refined/Banded) with a Betreff-first German letter — verified to fail on the unguarded renderer with the exact defect ('the subject line was consumed as the letterhead name') and pass with the fix. Full docx module 25/25, clippy `-D warnings` and fmt clean. Audited by the export domain critic with an independent revert-and-rerun check: zero findings.

Fixes #877

### DIFF
--- a/apps/desktop/src-tauri/src/export/docx/mod.rs
+++ b/apps/desktop/src-tauri/src/export/docx/mod.rs
@@ -91,13 +91,20 @@ fn generate_cover_letter_docx_classic(
         // market (was English/German only).
         let is_salutation = crate::locale::letter::is_salutation(&clean);
         let is_signoff = crate::locale::letter::is_signoff(&clean);
+        let is_subject_line = crate::locale::letter::is_subject_line(&clean);
 
         // First line is name — but ONLY a real letterhead name, not a letter that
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #880, opened 2026-07-23
test: de-flake use-extension-bridge cache-seeding assertion

saeedkolivand:
Root cause: the test mounts only the mutation hook, so the cache entry seeded by `onSuccess` has zero observers; with the shared test client's `gcTime: 0`, TanStack Query schedules an immediate eviction macrotask that races `waitFor`'s polling — the entry can vanish before the first poll. Reproduced 1–2/40 in isolation; 4 documented red CI sightings (#874, #879, plus pre-push reds).

Fix is test-only: a `persistentClient()` (`gcTime`/`staleTime` Infinity) for that one test, mirroring `use-autopilot.test.ts`'s documented pattern for observer-less cache seeding. Proof: 40/40 consecutive isolated runs green, full renderer suite 2462/2462.


saeedkolivand:
@claude review

saeedkolivand:

PR #886, opened 2026-07-27
fix: pin document viewer header while output scrolls

saeedkolivand:
The tailor-flow document viewer header (Resume | Cover letter | Job ad + Copy/Export) scrolled away because the parent panel owned the scroll. The scroll boundary now lives inside the viewer: the header pins, the option strips scroll with the document, and the document region keeps a 20rem floor.

Chromium-measured at 10 window/tab configurations: header moves 0px on scroll in 10/10; usable document area 340px in every config (the first iteration pinned the strips too and measured 0px at the app default window — caught by review and reverted to the measured-good shape).

Review: frontend-author → frontend-reviewer (1 HIGH, empirically measured, fixed + re-measured). Validation: 4103 tests; TailorFlow suites 272; typecheck 13/13.

### DIFF
--- a/apps/desktop/src/renderer/features/documents/components/TailorFlow/GenerationOutput.test.tsx
+++ b/apps/desktop/src/renderer/features/documents/components/TailorFlow/GenerationOutput.test.tsx
@@ -928,9 +928,11 @@ describe('GenerationOutput', () => {
   // Regression guard for the "header scrolls away" bug: the viewer used to grow
   // past its host (intrinsic `min-h-[32rem]` panel + no overflow of its own), so
   // the PARENT scrolled the whole component — header included. The scroll boundary
-  // now lives on the tabpanel and every control above it is a pinned `shrink-0`
-  // sibling, i.e. OUTSIDE the scrollport. jsdom has no layout, so these assert the
-  // structural invariants that produce the behaviour, not pixels.
+  // now lives on the tabpanel: the tab/action bar is a pinned `shrink-0` sibling
+  // OUTSIDE it, while the option strips live INSIDE and scroll with the document.
+  // jsdom has no layout, so these assert the structural invariants that produce the
+  // behaviour, not pixels — the pixel measurements come from the Chromium run
+  // recorded in the handoff.
 
   describe('scroll boundary', () => {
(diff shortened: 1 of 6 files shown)


saeedkolivand:
@claude review

saeedkolivand:
@claude review

saeedkolivand:
Thanks — all three actionable items are fixed and pushed (`5327ca22`).

**MEDIUM 2 (height chain)** — fair catch, and it was the real gap: every layout assertion I had stops at its own render container, so the two links at the top of the chain (the `min-h-0 flex-1` stage body and the `h-full` stage element) were unguarded and the regression could return with a green suite. `TailorFlow.test.tsx` now has a `height chain` test asserting both links on both stages; I mutation-verified it — dropping `min-h-0` fails with `expected 'flex-1' to contain 'min-h-0'`, and blanking the stage className fails with `expected '' to contain 'h-full'`.

**LOW 1 (test-id anchor)** — the document region now carries a real `data-testid` from `@ajh/test-ids` (`documents.documentRegion`) and the assertion targets that instead of `editable-output`, which only ever existed in that file's own mock.

**LOW 2 (numbers in comments)** — stripped from both layout comments; only the invariants remain (root is height-bounded, only the tab bar pins, the floor is what makes the scrollport engage). The measurements live in the task handoff.

**MEDIUM 1 (scroll chaining / strip consolidation)** — acknowledged, and deliberately not implemented here. You're right that a wheel gesture starting over the preview scrolls the preview first and only chains to the outer scrollport once it bottoms out. What I did verify in Chromium against the compiled CSS, at 1280×{569,600,769,800,1040} for both tabs: the document region is a full 340px in every configuration (including the 900×600 minimum), the header moves 0px on scroll in 10/10, and every strip is reachable — so nothing is unusable, it's an ergonomics nit rather than a defect. Collapsing the three strips into one compact row is a genuine improvement but it's a strip redesign that changes controls this PR doesn't otherwise touch, and it belongs with the design-system owner (`ui-ux-expert`) rather than a layout bug-fix. It's recorded in the handoff as a follow-up for the steward.

Also for the record: the coverage report's "No changed files found" note is expected here — the diff is class-only plus tests, so nothing moves the coverage table.

saeedkolivand:

PR #887, opened 2026-07-27
fix: localize apply-by-email greeting and drop contact line

saeedkolivand:
- the apply-by-email prompt hardcoded "Dear Hiring Manager," — it now derives greeting AND sign-off from the same market-conventions machinery as the cover letter (DACH → "Sehr geehrte Damen und Herren," / named variants, 16 markets), via resolveMarket(jobCountry, targetLanguage)
- generated emails no longer append a contact line / profile links — sign-off is the candidate name only
- outputTone preference now applies to emails (parity with cover letters)
- injection guard preserved and hardened (quote/brace strip, code-point-safe 80 cap); conditional prompt clauses now derive from the convention template, never from user-substituted strings (review caught es/fr misgendering edge cases)

Review: ai-provider-author → ai-provider-expert (APPROVE; advisories applied — 21 hostile names × 18 markets probed, zero placeholder leaks). Validation: prompts 544; whole graph 4115; typecheck 13/13.

### DIFF
--- a/apps/desktop/src/renderer/lib/generate/generation/generation.test.ts
+++ b/apps/desktop/src/renderer/lib/generate/generation/generation.test.ts
@@ -1041,24 +1041,45 @@ describe('output tone wiring (Settings → Output Tone)', () => {
     expect(systemOf(client)).toMatch(/TONE: a more narrative, distinctive voice/);
   });
 
-  it('threads the store outputTone AND the resolved market into the application-email prompt', async () => {
-    usePreferencesStore.setState({ outputTone: 'formal' });
-    const client = register();
-    // English email for a German job: only the market resolved from `jobCountry`
-    // can put a German salutation in the prompt — drop the `market` argument and
-    // this falls back to the international "Dear Hiring Manager,".
+  // The application-email path resolves its own market + tone (mirroring
+  // generateCoverLetter); each is asserted alone so a regression in one is never
+  // reported as a failure of the other.
+  type EmailMeta = Parameters<typeof generateApplicationEmail>[0]['meta'];
+  const runEmail = async (client: ReturnType<typeof register>, meta: EmailMeta) => {
     const p = generateApplicationEmail({
       resume: 'My resume',
       jobAd: 'Backend role in Berlin',
-      meta: { ...META, jobCountry: 'DE' },
+      meta,
       model: 'llama3',
     });
     await flushUntilStreaming();
     emit('Subject: Application\n\nGreeting.');
     done();
     await p;
-    const system = systemOf(client);
-    expect(system).toMatch(/TONE: formal and precise/);
+    return systemOf(client);
+  };
+
+  it('threads the store outputTone into the application-email system prompt', async () => {
+    usePreferencesStore.setState({ outputTone: 'formal' });
+    const client = register();
+    expect(await runEmail(client, META)).toMatch(/TONE: formal and precise/);
+  });
+
+  it('threads the market resolved from meta.jobCountry into the application-email prompt', async () => {
+    const client = register();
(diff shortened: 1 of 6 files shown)


saeedkolivand:
@claude review

saeedkolivand:
Review round addressed in `refactor: share the market-language fit between letter and email builders`.

**MEDIUM 1 — parenthesized gender variants.** The pick-one gate was slash-only, so `pt` (`Exmo.(a) Senhor(a) {lastName},`) and `ru` (`Уважаемый(ая) {firstName} {patronymic}!`) handed a named recipient an unresolved mail-merge artifact with no instruction to resolve it. `hasGenderVariants()` now covers both shapes (`includes('/') || /\(\p{L}{1,3}\)/u`), with explicit pt/ru named-recipient tests.

**MEDIUM 2 — hand-listed markets.** The sweep is now derived from the exported `LETTER_MARKET_IDS`, so a market added to the conventions table is covered the day it lands (br/ru/cn were the gap, and br/ru were exactly the MEDIUM 1 markets). It also asserts the invariant rather than a list: for every market, any rendered greeting still carrying alternatives must carry the clause that resolves them.

**MEDIUM 3 — shared helper.** Agreed, and the argument is the same one this PR is about. `marketLanguageFit(market, language)` now lives next to `letterConventions()` and owns the `sameLanguage` derivation, the `the formal <lang> equivalent of "…"` phrasing and the default sign-off; both builders read it, each keeping its own surrounding wording (the cover letter's `use "X" (named recipient) or "Y" (unknown recipient)` vs. the email's concrete greeting). Cover-letter output is unchanged **byte for byte** — verified empirically, not just by the suite: the `<market_conventions>` block was dumped for all 17 markets × 5 languages (incl. the empty-language corner case) before and after the refactor and diffed identical.

**LOW — brackets + test split.** `[` and `]` joined the sanitized delimiter set (they would close the `[Greeting: …]` skeleton slot) and the hostile fixture now asserts on them. The renderer wiring test is split into three: output tone, market from `meta.jobCountry`, and the `jobCountry`-absent → target-language-market fallback; each was verified to fail on its own when the corresponding argument is dropped from the builder call.

**LOW (no code) — two items recorded rather than changed here:**

1. *fr named slot.* Giving `fr` a name slot in `salutations.named` is a conventions-table change, and that table is shared with the cover letter and mirrored by the Rust renderer (`fixtures/letter-conventions.json` parity test), so it is out of scope for a prompt fix. Recorded as a follow-up for the resume-export/locale owner. Current behavior is the safe one: `fr` falls back to the generic `Madame, Monsieur,` rather than printing gendered alternatives with the recipient's name absent from the prompt.
2. *Brief-depth bracketed-greeting echo.* At `brief` depth the FORMAT skeleton is the only greeting instruction, so a weak local model could in principle echo the `[Greeting: …]` wrapper literally. Two things argue against changing it now: the skeleton already uses bracketed directives throughout (`[Body: …]`, `[Sign-off: …]`) at every depth, so this introduces no new shape, and the same-language case renders a concrete quoted salutation the model can copy verbatim. Flagged for the owner's manual small-model spot-check rather than pre-emptively re-worded.

**Known residual (unchanged, not a regression):** `ApplyByEmailTab` never sets `meta.jobCountry`, so the email market resolves from `targetLanguage` alone and can differ from the cover letter's market for the same job. Fixing it means carrying `jobCountry` on the saved generation / Application record — data plumbing, deliberately outside this diff.

Validation after the round (exit codes verified): `pnpm typecheck` (forced) 13/13 · `pnpm test` 382 files / 4143 tests · `pnpm lint` · `pnpm format:check` all green.

saeedkolivand:

PR #893, opened 2026-07-28
feat: offline geonames geocoding with photon fallback

saeedkolivand:
Location autocomplete no longer depends on the public Nominatim API (whose usage policy forbids autocomplete and whose throttling produced the silent zero-suggestion failures). Suggestions now come from a bundled offline GeoNames index (~34k cities + countries with ISO codes, 1.6 MB gz, parsed once at first use) with Photon as a zero-hit online fallback; Nominatim is fully retired — including the packages/ui Storybook default fetch that still hit it from the public site.

Every suggestion always carries an ISO country code — offline, in milliseconds — which the aggregator market routing needs. Review-hardened: country CODES and curated endonyms (deutschland, schweiz, …) match exact-only so junk prefix hits can't veto the fallback or persist a wrong country (pre-fix: "schweiz" → South Africa); "City, Country" round-trips re-resolve offline; a quality-gated + length-floored fallback keeps Photon fair-use; wiremock covers the real request path (URL shape, UA, timeout, 503 degrade).

CC BY 4.0 attribution for GeoNames (+ Photon/OSM ODbL) added in Settings → About (en+de) and docs; the network-egress ADR updated; regeneration script (network only at regen, never at build).

Review: scraping-applier-author → scraping-applier-expert (1 HIGH + 3 MEDIUM found and fixed). Validation: cargo 2764 (geocoding 40) + clippy/fmt/deny CI-exact clean; 4087 + 485 ui tests; typecheck 13/13.


saeedkolivand:
@claude review

saeedkolivand:
@claude review

saeedkolivand:

PR #895, opened 2026-07-28
feat: revoke extension pairing over the bridge on token rotation

saeedkolivand:
After a factory reset (or Settings→Regenerate), the extension kept saying "authorized" while nothing worked — the token rotated silently, live sockets survived, and the extension retried the dead token forever. This adds a `token.revoked` protocol message: sent only over already-authenticated sessions right before rotation, then sockets close and the connected count zeroes; the extension clears its stored token and shows the pairing view. Reset confirm copy now says the browser must be paired again (en+de).

Security-hardened through three review rounds: the revoke handler is gated on an explicit post-serverProof `authenticated` flag (an unauthenticated port-squatter probing the between-handshake-steps window was empirically driven pre-fix and is pinned by a permanent probe test, 7-window matrix); no token material rides the frame; unauthenticated peers still get the silent no-oracle close; the rotation lock-ordering was adversarially verified (no interleaving authenticates on the old token and misses the revoke). Old published extensions ignore the new verb safely (unknown-type no-op, pinned).

**Release note / known limitation.** The revoke only reaches browsers that are LIVE at the moment of rotation. A browser that was closed (or whose MV3 service worker had been evicted) never receives the frame, so on its next launch it still retries the dead token and shows "app not running" rather than the pairing view — the user has to hit Unpair manually. This is inherent to the no-oracle design: the desktop cannot tell a stale-token reconnect from a crashed app without becoming a token oracle, which is exactly what ADR-0010 forbids. Follow-up (not in this PR): after N consecutive failed handshakes, have the extension surface a non-authoritative "still not connecting — re-pair?" hint.

Note: shipping the fix to users requires a Chrome Web Store + Firefox AMO re-release.

Review: extension-author → extension-reviewer (APPROVE) + tauri-security-reviewer (1 HIGH found, fixed, CONFIRMED-RESOLVED). Validation: 535 extension tests; cargo bridge suite 315; clippy CI-exact clean; both dists import-free.


saeedkolivand:
@claude review

saeedkolivand:
@claude review

saeedkolivand:

PR #889, opened 2026-07-27
fix: match attribute-only and hyphenated name fields in autofill

saeedkolivand:
The extension autofill name matcher required a literal space ("first name"), so attribute-only fields (first_name, firstName, fname, job_application[first_name]) never matched — while hyphenated first-name/last-name fell to the catch-all and wrote the FULL name into both boxes. Patterns now match attribute compounds; aria-labelledby labels are read (Workday/Ashby-class forms); disabled/readonly fields are skipped in fill AND capture.

Three review rounds hardened the ambiguity denylist against the widened reach: third-party name boxes (spouse/reference/beneficiary families incl. camelCase), edu/org full-name fields, group-label-vs-attribute precedence, label dedupe, kana/furigana, alsoKnownAs — all differential-pinned (558 extension tests; every new pin verified red against the old matcher). allFrames injection evaluated and rejected on least-privilege/store-policy grounds; documented as a known top-frame-only limit.

Review: extension-author → extension-reviewer (1 HIGH + 4 MEDIUM found and fixed; re-verify CONFIRMED-RESOLVED). Both store dists rebuilt; classic bundles remain import-free.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #883, opened 2026-07-27
ci: bump claude review model to opus 5

saeedkolivand:
- Bumps the @claude-review workflow model from `claude-opus-4-8` to `claude-opus-5` (the schema-1 ai-review-gate job stays on `claude-sonnet-5`).
- Scopes the pre-push hook's build gates: only `ci-pipeline.yml` and `.github/actions/**` can change what the frontend/rust gates do, so workflow-only pushes (like this one) no longer run the full frontend+Rust suites locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #898, opened 2026-07-28
docs: close out the 2026-07-27/28 bug batch (#883-#895)

saeedkolivand:
Documentation-only close-out of the merged batch. Every claim below was checked against the code on `main` before it was written; nothing from an unmerged branch is described as shipped.

## Knowledge base

**`docs/knowledge/extension-domain.md`**

- Primary-paths table gains `apps/desktop/src-tauri/src/extension_bridge/msg.rs` (the Rust mirror of the TS wire-type table) and `extension_bridge/revoke.rs` (`revoke_frames(authenticated)` — ADR-0010's no-oracle rule as a pure, socket-free decision).
- Protocol-lockstep rule re-aimed at `msg.rs`; `mod.rs` no longer holds the constants.
- Verb table gains `token.revoked`: sent **only** over an already-authenticated session on token rotation / factory reset, no payload and no token material; an unauthenticated peer still gets the silent reply-less close, so the token oracle stays denied.
- New **Revocation reach** section records the inherent limitation: revoke reaches only browsers live at the moment of rotation. A browser closed at rotation reconnects with the dead token, gets the silent close, and retries — the pre-#895 stranding, narrowed rather than eliminated. The bounded-retry "re-pair?" hint is filed as a follow-up, not claimed as shipped.
- New **Store re-release pending** section: both the wire protocol (#895) and the autofill field matcher (#889) changed, so nothing reaches users until a Chrome Web Store + Firefox AMO release ships.

**`docs/knowledge/scraping-domain.md`**

- `companies-failed:<n>` documented (`ats_failed_fetches_note`, Ashby/Lever — the validator-less ATS boards): per-company failures now surface as a chip instead of vanishing into the log. Added to the note-precedence table with the disjoint-emitter note, and the i18n key recorded.
- Trust-program fast-follow 1 (`fetches-failed:<n>`) marked **CLOSED** — it shipped under the `companies-failed` name.
- New **Geocoding — offline first** section: bundled GeoNames index (no build-time fetch, digest-asserted), Photon as a quality-gated fallback, Nominatim retired and why, mandatory CC BY 4.0 + ODbL attribution, and the ADR-0005 exemption clause. Counts, memory figures and digests are left in `apps/desktop/src-tauri/geodata/README.md` rather than copied.

## Stale-reference cleanup (comment text and docs only — no logic touched)

- `apps/desktop/src-tauri/src/net/http.rs` — the UA-override example now names Photon; `geocoding.rs` really does set `USER_AGENT` for the Photon request.
- `apps/desktop/src-tauri/src/commands/scrape.rs` — three test/doc comments no longer describe a "Nominatim round trip".
- `docs/CONTEXT.md` — **user-initiated lookup** split out of **enrichment egress**. The glossary previously required location autocomplete to be opt-in/default-OFF, which now contradicts ADR-0005's exemption clause.

Deliberately left alone: the historical mentions in `geocoding.rs` and `useGeocoding.ts` ("the Nominatim era", "which violated that endpoint's policy") — those describe the past correctly.

## Dev setup and status

- `docs/DEVELOPMENT.md` — new **Migrations and stale dev DBs** subsection: per-store migration lists, `PRAGMA user_version` as the guard, one transaction per migration, and the two cases that actually need a manual delete.
- `docs/ARCHITECTURE_STATUS.md` — refreshed for the batch: jobs command bar + `Drawer` primitive + split view, `companies-failed` chips, offline geocoding, autopilot `pages` field, apply-by-email draft persistence, localized salutation, bridge revocation, autofill matcher.

## Follow-ups ledger

`docs/NEXT_ISSUES.md` items 9–19, each verified against source:

1. **[blocking the next edit]** `extension_bridge/mod.rs` is at **1398 of the 1400-line R8 cap** — split it before touching it again.
2. Design-system light-scheme contrast sweep (incl. the company-avatar tint bump).
3. `ModalShell` focus-return hardening (`Drawer` has it, `ModalShell` does not).
4. `Tag.closeLabel` defaults to hardcoded English.
5. Missing applications-cache invalidation + error handling in the two remaining direct `api.aiGenerations.save` callers.
6. Email draft-delete path — extend `update_texts`' `Option` pattern to the email columns.
7. Key the email-generation save on `applicationId` so URL-less applications can persist a draft.
8. RAII guard for the scrape cancellation-token registry (needs a synchronous lock first).
9. Extension bounded-retry "re-pair?" hint.
10. `PipelineStrip` narrow-width row density.
11. `GenerationCard` empty badges on fresh inserts.

## Lessons

Nine appended to the local `.claude/memory/lessons.jsonl` (gitignored, so not in this diff): the unlayered `utilities.css` cascade trap, `autoFocus` vs. focus-trap return-focus, skeleton class-constant sharing + reserved row height, sentinel fields must never gain cost semantics, mutation-testing the original line rather than a consumer, the endpoint-parameter seam with self-asserting preconditions, discriminated-union IPC results as a compiler-driven call-site audit, plus two harvested from the applications-upgrade handoff (dead equal-specificity override; the eslint/`tsc` cast conflict). The store held at its 200 cap; the 7 oldest auto-archived.

## Deferred — intentionally not in this PR

- **ADR-0001 "grouped/sectioned list" supersession** — the applications-page redesign (`feat/applications-upgrade`) has no PR yet. The supersession gets written when it merges.
- **Aggregator page-loop docs** — PR #896 is still in review. The related standing rule is captured as a lesson, but no doc claims the enforcing refactor has landed.

## Correction worth flagging

The brief said the email-persistence migration took `user_version` to **8**. It is **7**: `AiGenerationStore::MIGRATIONS` has seven entries, `add_email_draft` is the last, and `run_migrations` sets `user_version` to the 1-based index. `docs/DEVELOPMENT.md` records 7.

## Verification

- `pnpm format:check` clean; `cargo fmt --check` clean.
- `graphify update .` run (17178 nodes / 33200 edges); `codegraph sync` reported already-up-to-date. Both outputs are gitignored, so neither appears in the diff.
- The pre-push hook merged `origin/main` once and aborted; re-pushed as designed.


saeedkolivand:
@claude review

saeedkolivand:

PR #899, opened 2026-07-28
docs: record the applications redesign and aggregator spend budget (#896-#897)

saeedkolivand:
Docs addendum for the 2026-07-27/28 batch, closing out PRs #896 (aggregator page loop) and #897 (applications page redesign). Docs only — no source changes. Every claim below was verified against the code on `origin/main`, not against the PR descriptions.

## 1. ADR-0001 — dated supersession addendum

`docs/adr/0001-application-aggregate-split.md` keeps its body and its decision. A new `## Addendum (2026-07-28, PR #897)` section supersedes **only** the last Consequence bullet's "grouped/sectioned collapsible list by stage" as the route's single list shape:

- unfiltered → stage sections (as originally decided);
- **any** active filter → one **flat** list, because a section header under a filter either repeats the strip card selected above it or reprints each row's own stage tag.

Points at the owning source (`features/applications/lib/pipeline.ts`, the `sections` memo in `ApplicationsPage/index.tsx`) and notes that `PIPELINE_GROUPS` is hand-derived on purpose and pinned against `APPLICATION_STAGES` by `pipeline.test.ts`. No card count copied into the doc.

## 2. `docs/knowledge/scraping-domain.md` — aggregator page loop & spend budget (#896)

New section covering the `SearchBudget` contract, verified in `apps/desktop/src-tauri/src/scraping/boards/aggregator/mod.rs`:

- `SearchBudget::from_input(&BoardSearchInput)` is the only production constructor — `new`/`items_only`/`manual` are `#[cfg(test)]`, deliberately a separate hop so a hand-built test budget cannot mask a rewrite of the real mapping.
- `provider_amount: Option<u32>` **fails closed**: passed through unchanged, and `None` (the default, and every scheduled run) means no provider spend. Only `commands/scrape.rs` sets it; `autopilot_helpers` leaves it `None`.
- The paid Apify tier both **gates and sizes** off `provider_amount` via `apify_cap()` — `None ⇒ 0 ⇒ no request at all`; `Some(spend)` asks only for the unmet remainder, clamped to the item ceiling.
- Metered Adzuna/JSearch fetches use `retries: 0` — a 429 from a metered API *is* the over-quota signal, so retrying re-spends the quota that just ran out. Follows the `apify_fetch_options` precedent; all three invariants are pinned by tests asserting through the same helpers production calls.
- `pages` is inert for the aggregator, so the autopilot wizard disables the "Pages to scrape" field (`pagesInert`) when the aggregator is the only selected board.

**Behavior change recorded, not buried:** Apify-enabled users' scheduled/autopilot runs no longer buy LinkedIn results (manual scrapes unchanged), and **there is no UI signal for it** — flagged as follow-up 22 rather than left implicit.

## 3. `docs/ARCHITECTURE_STATUS.md`

- **Applications & Tracker** — four new rows for #897: pipeline strip, unified primary contact (`contact_*` canonical, `recipient_*` deprecated aliases folded by both the migration and `applications/contact.rs`), follow-up reminders through the notification center (`reminder_scheduler.rs` → `push_and_notify`, data half in `applications/reminders.rs`), and status notes into the append-only `status_events` history.
- **Scraping** — new aggregator paging/spend-budget row for #896.
- `Last updated` line refreshed to name #896–#897.

## 4. Follow-ups ledger (`docs/NEXT_ISSUES.md`, items 20–22)

20. **Metered scraping has no daily backstop** — AI providers carry `PROVIDER_DAILY_MAX`; scraping has only `SCRAPE_RATE_MAX` per the rolling `RATE_WINDOW` plus a concurrency cap. The window ceiling was raised twice during review as paging multiplied per-search calls, which is the signal it bounds burst rate rather than the day's spend.
21. **Byte-aware client guard for the status-note cap** — server `MAX_STATUS_NOTE_BYTES` counts bytes (consistent with every other cap in that module); the `StatusNoteModal` `maxLength` counts chars, so an over-byte multi-byte note gets a save-time Validation error instead of being stopped while typing.
22. **UI signal for the Apify scheduled-run change** — from item 2 above.

## 5. Lessons (`.claude/memory/lessons.jsonl`)

Five persisted after dedupe against the existing log (untracked file, not part of this diff):

- **rust/testing** — rusqlite `prepare` compiles against the connection's *cached* schema, so after a `DROP TABLE` the first sweep fails at STEP (per-row arm) and only the *next* one fails at prepare. The original single-sweep test passed even with the guard replaced by `.expect()`.
- **rust/design** — exhaustive struct literals as a wire-contract guard; adding `next_action_notified_at` to `Application` caught all three stale fixtures on the first `cargo check`.
- **process/review** — falsify structural fixes by reverting them; this probe is what exposed the wrong test above.
- **frontend/a11y** — a chip nested in a keyboard-activatable row must stop **keydown** propagation, not just click (Enter/Space fires both).
- **process/rust** — near an R8 LOC cap, extract a cohesive submodule and re-export from the parent so no caller changes (`extension_bridge/revoke.rs`, `applications/reminders.rs`).

## Verification

`graphify update .` (17405 nodes / 33728 edges) and `codegraph sync` (already up to date) both run. Claims checked against source: `SearchBudget`/`apify_cap`/`apify_fetch_options`, `limits/mod.rs` constants, `MAX_STATUS_NOTE_BYTES` + `validate_status_note`, `pipeline.ts`, the `ApplicationsPage` `sections` memo, `ApplicationRow`'s note-chip `onKeyDown` guard, `reminder_scheduler.rs`, `applications/reminders.rs`, `applications/contact.rs`, and `StepTarget`'s `pagesInert`.


saeedkolivand:
@claude review

saeedkolivand:

PR #915, opened 2026-07-30
refactor: split architecture-map interactions into focused modules

saeedkolivand:
## What

`ArchitectureMap.tsx` was **972 lines**, dominated by a single **454-line `useEffect`** that wired pan, zoom, drag, keyboard shortcuts, a focus trap and a11y announcements by reaching into prerendered DOM through selectors. PR 2 of 4 splitting the oversized `apps/landing` components.

**The effect decomposes into `lib/architecture-map/`:**

| Module | Lines | Owns |
| --- | ---: | --- |
| `geometry.ts` | 111 | `CANVAS`, `edgePath` bezier, and `fit`/`screenToVB`/`stageCenter`/`zoomAt`/`ensureVisible` — now pure, taking rects as parameters instead of reading the DOM |
| `filter.ts` | 35 | `buildAdjacency`, `hasBadge`, `passNodeFilter`, `passEdgeFilter` — take `filter` as a parameter instead of closing over a mutable |
| `viewport.ts` | 102 | the mutable transform state, wheel zoom, drag pan, resize |
| `keyboard.ts` | 132 | the shortcut switch and the help-dialog focus trap |
| `interactions.ts` | 217 | orchestrator — `attachMap(root, { onPanel })`, node/chip/stage listeners, viewBox handoff |

Zoom factors, scale clamps and arrow steps are named constants now (`ZOOM_FACTOR`, `ZOOM_MIN`/`ZOOM_MAX`, `WHEEL_ZOOM_FACTOR`, `ARROW_STEP`/`ARROW_STEP_FAST`) instead of bare literals scattered through the switch.

**Markup splits into** `Sidebar.tsx` (163, with one `Field` component collapsing a 9× repeated shape), `Layers.tsx` (142, the three memo components unchanged), `MapChrome.tsx` (102 — five `<marker>`s that differed only in `id`+`fill` now map over a const). Legend rows and keyboard-shortcut rows move to `data/architecture-map.ts`, already the declared single source of truth for this page.

`ArchitectureMap.tsx` is now **116 lines**: state, ref, one `useEffect`, and the page skeleton.

## Behavior

**Strictly behavior-preserving.** Every id, class and `data-*` the interaction layer queries survives byte-identical, as does the viewBox save/remove/restore handoff that passes the SSR no-JS baseline off to pixel-space transforms.

## Verification

- `pnpm --filter @ajh/landing typecheck` — clean
- `pnpm --filter @ajh/landing test` — 167 tests, 21 files, all pass, including all 6 `ArchitectureMap.test.tsx` regression tests (pin/unpin, Enter+Escape with `#a11y-status` announcements, chip filters toggling `.dim`/`.hide`, help dialog)
- `pnpm lint:strict` — clean at `--max-warnings 0`
- `prettier --check` — clean

Audited by `frontend-reviewer`, which focused on the highest-risk surface: mutable state (`tx/ty/scale`, `dragging/moved/lx/ly`, `filter/pinned/hover`) that used to be plain closure variables read live by every handler and is now split across four modules. It verified no value is captured where the original read live, that `consumeDragMoved` reproduces the click-suppression-after-drag semantics exactly, that every listener registered through the relocated `on()` collector is still torn down by `attachMap`'s returned cleanup, and that the boot ordering (`renderSidebar → applyView → fit`) is unchanged. No HIGH/CRITICAL findings.


saeedkolivand:

PR #916, opened 2026-07-30
refactor: split agent-fleet into sections and pure scroll modules

saeedkolivand:
## What

`AgentFleet.tsx` was **830 lines**: six page sections, three scroll-driven effects, and a component defined inside the render body. PR 3 of 4 splitting the oversized `apps/landing` components.

**Pure math → `lib/agent-system/`:** `links.ts` (bezier geometry), `belt.ts` (scrub progress, `activeX` interpolation, nearest-station argmin), `reveal.ts` (count-up cubic easing, scene clamp), `clipboard.ts` (copy helper + the Enter/Space `onKeyDown` that was duplicated on both `.copy-cmd` buttons), `constants.ts`.

The hardcoded `/ 9` at three sites is now `STATION_COUNT`, derived from `STATIONS.length`. The `(max-width:780px)` breakpoint, the 160ms resize debounce, the 900ms count duration, the 1300ms copied timeout and the IO thresholds are named constants instead of bare literals.

**Effects → `hooks.ts`:** `useMapLinks`, `useBeltScrub`, `useReveal`, each keeping its own rAF ticking flag and listener teardown.

**Markup → components:** six `sections/*Section.tsx`, plus `Glyph`, `Machine`, `Divider` (6 render sites), `Station` (one component rendering both the `.belt` and `.belt-vert` DOM shapes) and `NodeButton`. Hero coords, the six divider paths, legend rows, the three team lists and the save-job prompt move to `data/agent-fleet.ts` as **additive** exports — `AgentTuple`/`AUTHORS`/`PAIRS`/`CROSS_NODES`/`ROUTES`/`STATIONS`/`BY_NAME` are untouched, and `scripts/check-agent-system.mjs` still passes.

`AgentFleet.tsx` is now **29 lines**.

## The one behavior change

**`NodeButton` is hoisted from the render body to module scope.** It previously got a fresh component identity on every render, so React unmounted and remounted it — focus and CSS transitions restarted. It now takes `{ name, roleClass, selected, onSelect, onHover }`, with the three call sites passing `selected === name`, `setSelected` and `setLitName` explicitly. Everything else in this PR is behavior-preserving.

Beyond the plan, state also follows its markup: `route` into `IntakeSection`, `selected`/`litName` into `FleetMapSection`, with `useMapLinks`/`useBeltScrub` called from the sections that own their DOM. Only `useReveal` stays at the root, since `.reveal`/`.draw` spans every section. That is why the root landed at 29 lines instead of the planned ~120, with no prop drilling.

## Verification

- `pnpm --filter @ajh/landing typecheck` — clean
- `pnpm --filter @ajh/landing test` — 167 tests, 21 files, all pass
- `pnpm lint:strict` — clean at `--max-warnings 0`
- `pnpm exec prettier --check` — clean
- `node scripts/check-agent-system.mjs` — in sync

**This component has no unit test**, so the audit was the gate. `frontend-reviewer` verified the two risky changes empirically rather than by reading: it mounted the tree and clicked a node to confirm `aria-expanded` flips and the detail panel updates through the new component boundary, and it mounted with a synchronous fake `IntersectionObserver` to confirm the root-level `useReveal` still activates 100% of `.reveal` nodes across all six child sections. It also hand-derived Babel's JSX whitespace collapse for every `WORK_TOGETHER_TEAMS` and `FLEET_LEGEND` segment against the original — the lost-`{' '}` class of regression that has bitten this repo before. No HIGH/CRITICAL findings.

## Follow-up

The newly-pure modules (`nearestStationIndex`, `activeStationX`, `beltProgress`, `linkPathD`, `sceneProgress`, `easeOutCubic`, `countUp`) were extracted partly to be testable but ship without tests — tests were scoped out of this batch. Cheap follow-up for `test-author`.


saeedkolivand:

PR #917, opened 2026-07-30
refactor: split home body and beats into per-section components

saeedkolivand:
## What

`HomeBody.tsx` (**753 lines**) and `HomeBeats.tsx` (**769**) were pure static JSX — no hooks, no handlers, no logic, no imports beyond `@/lib/site-links`. Final PR of 4 splitting the oversized `apps/landing` components.

They split **verbatim** into `components/home/sections/{PageChrome,Hero,Features,Testimonials,Finale}.tsx` and `components/home/beats/Beat{1,2,3,4}.tsx`, leaving shells of **40** and **23** lines.

**No JSX was rewritten.** The repeated markup that looks like DRY bait — the ~40 decorative `<svg className="deco">` blocks, 11 `.feat` cards, 9 `.quote` blocks, 4 identical `.npc` figures — stays literal. That was a deliberate call: see below.

## Why verbatim, and how it's verified

While planning this batch I found that **ADR-0018's description of the parity gate is stale**. It claims `pnpm check:parity` "byte-matches the legacy static-site layout per page via `scripts/diff-dom.mjs` element-by-element." It does not — `check-parity.mjs` never imports `diff-dom`, and `diff-dom.mjs` has **no callers at all**; the `src/content/*/body.html` baselines it compared against were deleted in the TSX conversion. The live gate runs three *content* checks against `out/`: signature phrases, a frozen href list, installer URLs.

So there is no structural regression gate on these pages — only text and links. That is exactly why data-ifying the markup was the wrong move here, and why the split was proven by hand instead. Against a pre-split build:

| Check | Before | After |
| --- | ---: | ---: |
| Tag sequence (name + id + class, document order) | 531 | **531, byte-identical** |
| Visible text (whitespace-collapsed, scripts stripped) | 38,260 | **38,260, byte-identical** |
| Direct `<section>` children of `<main>` | 8 | **8** |

That last one matters: `public/scripts/home-0.js:231` does `querySelectorAll('main > section')` and routes the journey path by index.

Re-verified after the pre-commit hook's `prettier --write` pass, since JSX reflow is precisely how a lost `{' '}` would sneak in.

## Config

- **`eslint.config.mjs`** — the `no-restricted-syntax` inline-hex exemption now lists the two shells plus `sections/**` and `beats/**`. Deliberately **not** a `home/**` glob: that would silently extend the exemption to `CookieGag.tsx`, which is hand-written (it exists to convert an inline `onclick` to `onClick`) and should stay subject to the brand-token rule. `frontend-reviewer` caught this by running the actual minimatch matcher rather than eyeballing the glob.
- **`scripts/check-landing-drift.mjs`** — all 9 new files added to `SECRET_SCAN_FILES`. That guard hard-fails only on a *listed* file that has vanished, so an unlisted file fails **open** and is silently never secret-scanned — the exact regression its own comment at `:44-53` documents.

## Docs (via `project-steward`)

- **ADR-0018** — parity-gate paragraph corrected to describe the three content checks, and states plainly that `diff-dom.mjs` is orphaned and no structural DOM gate currently exists. Steward's recommendation: delete `diff-dom.mjs` + its test as dead code. Left for a separate decision rather than bundled here.
- **ADR-0019** — the `HomeBody.tsx` lines 693-694 citation for the `/world` link becomes a stable anchor reference into `sections/Finale.tsx`.
- `graphify update .` run: 18,187 nodes, 35,860 edges.

## Verification

`typecheck` · `test` (167 landing, 4293 repo-wide) · `lint:strict` · `prettier --check` · `check-landing-drift` · `build` · `check:parity` (27 phrases, 13 links, 7 installer URLs) — all clean.

Unrelated note: the pre-push hook initially failed on a missing Playwright `chromium_headless_shell-1234` for the `packages/ui` Storybook browser tests — a local browser-cache gap from a Playwright bump already on `main`, fixed with `playwright install chromium-headless-shell`. Nothing to do with this branch.


saeedkolivand:

PR #914, opened 2026-07-30
refactor: split mission-control into sections and a pure model module

saeedkolivand:
## What

`MissionControl.tsx` was **701 lines** mixing GitHub data loading, view-model assembly, formatters, and five page sections in one file. This is PR 1 of 4 splitting the oversized `apps/landing` components.

**Pure logic → `lib/mission-control/model.ts`**, joining the eight already-tested modules in that directory:

- `RepoData` / `BenchmarkGlobal`
- `fmtInt` / `round1` / `pctOrDash`
- `safe`, `loadAll`
- `buildModel` + `type Model` — 66 lines of pure `RepoData → Model` with zero React

`actionById` moved beside `WRITE_ACTIONS` in `write-actions.ts` where it belongs.

**Markup → components**, collapsing the repeated shapes to one definition each:

| Component | Call sites collapsed |
| --- | ---: |
| `Stat.tsx` | 15 |
| `Section.tsx` (`mc-section` + eyebrow + title) | 5 |
| `ExternalLink.tsx` | 5 |
| `sections/{Delivery,Work,Quality,Community}Section.tsx`, `sections/MaintainerActions.tsx` | — |
| `MissionControlSkeleton.tsx` | — |

`MissionControl.tsx` now holds state, the four effects, `confirm`/`runAction`, the three `useMemo`s and composition — **235 lines**.

## Behavior

**Strictly behavior-preserving.** Rendered DOM, text content, class names, attribute values and interaction flow are unchanged. `seriesColors` was being recreated every render and is now a module constant — the only perf-visible difference.

## Verification

- `pnpm --filter @ajh/landing typecheck` — clean
- `pnpm --filter @ajh/landing test` — 167 tests, 21 files, all pass. The six existing `MissionControl.test.tsx` end-to-end tests are **untouched** and still green, including the two security-relevant ones (untrusted issue title stays escaped; write is gated behind the confirm dialog and the token never reaches a URL).
- `pnpm lint:strict` — clean at `--max-warnings 0`

Audited by `frontend-reviewer`: no HIGH/CRITICAL findings. It verified all 5 `Section` and all 15 `Stat` call sites reproduce their original markup, that `buildModel` kept every slice limit and `MC_CONFIG` read, and that no `dangerouslySetInnerHTML` or token-in-URL path was introduced.

## Note

[...]

PR #872, opened 2026-07-23
refactor: convert landing privacy and download pages to tsx (PR1/3)

saeedkolivand:
PR1 of 3 converting the landing's raw-HTML pages to pure TSX (plan: extract shared components/CSS/fonts, delete the `body.html` pipeline). This PR: the shared foundation + the two highest-duplication pages, `/privacy` and `/download`.

## What changed

- **Parity gate** (`scripts/check-parity.mjs`): the expected-href check no longer derives from `src/content/*/body.html` (those files are being deleted as pages convert) — the 13 hrefs are frozen as a curated literal like `PHRASES`, and built output is entity-normalized before matching (React's serializer escapes `'`/`"` in text nodes). Phrase, installer-URL, and deploy-shape checks untouched. New `scripts/diff-dom.mjs`: jsdom DOM-equivalence checker (skips style/script/link + whitespace-only text) used to prove each converted page renders an identical body vs a main-build baseline.
- **Fonts**: `<Fonts />` hoisted from all 9 pages into the root layout (built output unchanged — React dedupes the stylesheet link).
- **Shared foundation**: `marketing-tokens.css` (the byte-identical `:root` font/paper/ink prefix), `marketing-base.css` (only rules byte-identical between the two paper pages; `.top-back:focus-visible` deliberately stays download-only), `site-links.ts` URL constants, `SiteFooter` (current page rendered as plain text) + tests, `BackLink`.
- **/privacy** → `PrivacyBody.tsx` + `styles/privacy.css`; `src/content/privacy/` deleted.
- **/download** → `DownloadBody.tsx` + `DownloadCards.tsx` (server component replacing the `buildDownloadsHtml` string seam; the seven `.dl-btn` anchors keep the positional order `DownloadFreshness` swaps by index) + `styles/download.css`; `src/content/download/` and `src/lib/downloads.ts` deleted. `DownloadFreshness.test.tsx` now renders the real `DownloadCards` so the fixture cannot drift from production markup.
- **Structural render tests** for both new bodies (anchor ids, `.dl-btn` order, `copy-cmd` `data-copy` values, footer wiring) — added to enforce the DOM contract going forward.

## Verification (ADR-0018 protected marketing skin)

- `diff-dom.mjs`: **all 9 routes render an identical DOM** vs the pre-refactor baseline (privacy 166 elements, download 79, plus the 7 untouched routes).
[...]

PR #918, opened 2026-07-30
chore: remove orphaned diff-dom parity tooling and dead label metric

saeedkolivand:
## What

Two of the three follow-ups deferred from the landing component split (#914–#917). The third (tests for the extracted pure modules) is a separate PR.

**1. `scripts/diff-dom.mjs` + `diff-dom.test.mjs` deleted.** The script compared built output element-by-element against `src/content/*/body.html` baselines. Those baselines were removed in the TSX conversion (#872/#879/#881), so it has had no input and no caller since — `check-parity.mjs` never invoked it.

ADR-0018 now records what actually replaces it: a manual before/after diff of the built `out/*.html` comparing the tag sequence (element + `id` + `class`, document order) and whitespace-collapsed visible text, **with `<script>` blocks stripped from both first**. That caveat matters — Next's per-build RSC hydration payload and random `buildId` produce spurious drift otherwise. This is the check that verified #917.

**2. `work.labels` removed.** `buildModel` computed `issuesByLabel(data.issues).slice(0, 8)` and nothing ever rendered it — true before the split too. With the field gone, `issuesByLabel` had no production caller left (only its own test), so the helper and its assertions in `metrics.test.ts` go too.

**3. Two stale provenance comments** in `HomeBody.tsx` and `CookieGag.tsx` pointed at `diff-dom.mjs` to explain why the `onclick` → `onClick` conversion was safe. They now state the deviation directly.

Net **−316 / +6**.

## Verification

`pnpm --filter @ajh/landing typecheck` · `test` (167 pass) · `pnpm lint:strict` — all clean.


saeedkolivand:

PR #919, opened 2026-07-30
test: cover the pure modules extracted from the landing component split

saeedkolivand:
## What

#915 and #916 pulled geometry, scroll and easing math out of `ArchitectureMap` and `AgentFleet` into pure modules **specifically so it could be tested**, then deferred the tests. This pays that back — the last of the three follow-ups from that batch.

**90 new cases** across six co-located test files:

| Module | Covers |
| --- | --- |
| `lib/agent-system/belt.ts` | `STATION_COUNT` invariant, `beltProgress` clamp table, `activeStationX` interpolation, `nearestStationIndex` |
| `lib/agent-system/links.ts` | `linkPathD` anchors, cubic control-point structure, degenerate rects |
| `lib/agent-system/reveal.ts` | `easeOutCubic`, `sceneProgress`, `countUp` (both the reduced-motion sync path and the animated one) |
| `lib/agent-system/clipboard.ts` | `copyChip` precedence + timer, `copyChipOnKeyDown` key filtering |
| `lib/architecture-map/geometry.ts` | `edgePath`, `fit`, `screenToVB`, `stageCenter`, `zoomAt`, `ensureVisible` |
| `lib/architecture-map/filter.ts` | `buildAdjacency`, `hasBadge`, `passNodeFilter`, `passEdgeFilter` |

The cases target **boundaries, not happy paths**: zoom clamped at, below and above `ZOOM_MIN`/`ZOOM_MAX`; `easeOutCubic` checked for *deceleration* by comparing deltas across the domain rather than just pinning endpoints; `screenToVB` round-tripped against its forward transform; `nearestStationIndex` swept for a valid index on every input including ±Infinity; `copyChip` driven through the `.copied` timer with fake timers; `edgePath` asserted structurally rather than by brittle full-string match on the `d`.

Landing suite: **167 → 257 tests**.

## Two source quirks documented, not fixed

Both asserted with explanatory comments rather than silently patched, since this is a test-only change:

- `sceneProgress(0, 0, 0)` returns `NaN` — `0/0` escapes both clamp comparisons, because `NaN < 0` and `NaN > 1` are each `false`. Needs a simultaneously zero-height scene *and* viewport, which a real `getBoundingClientRect` never produces.
- `nearestStationIndex([], x)` returns `0`, an invalid index for an empty array. `centers` derives from `STATIONS`, which is never empty.

Neither is reachable through the real callers. A one-line guard on each (`Number.isFinite` clamp, `-1` sentinel) is cheap hardening if you want it — say the word and it's a fast-follow.

## Verification

`test` (257 pass) · `typecheck` · `lint:strict` · `prettier --check` — all clean.

`testing-reviewer` audited and found nothing blocking. It specifically attacked the risk that geometry tests just restate the source formula: it hand-mutated a sign flip into `zoomAt`'s pan term and traced the assertions through, confirming they fail against the broken implementation. It also verified the RAF stub in `reveal.test.ts` drives three explicit frames deterministically rather than firing once, and confirmed `git diff` touches **no production code**.

### DIFF
--- a/apps/landing/src/lib/agent-system/belt.test.ts
+++ b/apps/landing/src/lib/agent-system/belt.test.ts
@@ -1,0 +1,71 @@
+import { describe, expect, it } from 'vitest';
+
+import { STATIONS } from '@/data/agent-fleet';
+
+import { activeStationX, beltProgress, nearestStationIndex, STATION_COUNT } from './belt';
+
+describe('STATION_COUNT', () => {
+  it('is derived from STATIONS, never a hardcoded literal', () => {
+    expect(STATION_COUNT).toBe(STATIONS.length);
+  });
+});
+
+describe('beltProgress', () => {
+  it.each([
+    // [sectionTop, sectionHeight, viewportHeight, expected, why]
+    [0, 500, 500, 0, 'zero runway (height === viewport) never divides by zero'],
+    [0, 400, 500, 0, 'negative runway (section shorter than viewport) also stays 0'],
+    [50, 600, 500, 0, 'not yet scrolled into view (positive top) clamps to 0'],
+    [0, 600, 500, 0, 'exactly at the top of the runway is 0'],
+    [-50, 600, 500, 0.5, 'linear interior point of a 100px runway'],
+    [-100, 600, 500, 1, 'exactly at the end of the runway is 1'],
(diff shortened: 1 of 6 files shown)


saeedkolivand:

PR #920, opened 2026-07-30
refactor: split the remaining ported landing bodies into sections

saeedkolivand:
## What

`HowItWorksBody.tsx` (**340**), `PrivacyBody.tsx` (**327**) and `CreatureBody.tsx` (**262**) were the last oversized files in `apps/landing` — all three pure static JSX ported from hand-authored HTML, with no hooks, handlers or logic.

Split **verbatim** into `sections/` directories beside each body, cutting at the markup's own structural units:

| Page | Sections | Shell |
| --- | --- | ---: |
| how-it-works | `Sidebar`, `Overview`, `Boot`, `Flows`, `IpcReference`, `Subsystems`, `CheatSheet` | **38** |
| privacy | `IntroShort`, `Extension`, `Desktop`, `Footer` | **25** |
| creature | `Doodles`, `Stage`, `Overlays` | **24** |

No JSX was rewritten — same approach as #917 for the home files.

## Verification

Against pre-split builds, tag sequence (element + `id` + `class`, document order) and whitespace-collapsed visible text, `<script>` stripped from both:

| Page | Tags | Text chars |
| --- | --- | --- |
| how-it-works | 189 → **189, identical** | 23,474 → **23,474, identical** |
| privacy | 191 → **191, identical** | 15,134 → **15,134, identical** |
| creature | 94 → **94, identical** | 10,560 → **10,560, identical** |

Re-checked after the pre-commit hook's `prettier --write` pass — still identical.

`typecheck` · `test` (167 pass, incl. `HowItWorksBody.test.tsx` and `PrivacyBody.test.tsx` **unmodified**) · `lint:strict` · `prettier --check` · `check-landing-drift` · `build` · `check:parity` — all clean.

## Config

- **`eslint.config.mjs`** — the inline-hex exemption gains `creature/sections/**` (the `#sitefoot` hex moved into `Overlays.tsx`). Scoped, **not** a `creature/**` glob, so the directory stays covered by the brand-token rule if a hand-written file ever lands there — the trap #917 hit with `home/**` and `CookieGag.tsx`.
- **`scripts/check-landing-drift.mjs`** — all 14 new files added to `SECRET_SCAN_FILES`, which fails open for anything unlisted.

## Comment corrections

The privacy section headers claimed `privacy-0.js` binds to the page. It doesn't — it's a nine-line console easter egg with no DOM calls; `check-parity.mjs` is what actually pins that page's copy and anchor hrefs. Corrected on the sections and on `PrivacyBody.tsx`, where the claim originated (it pre-dates this split; the split had quadrupled its footprint).

`frontend-reviewer` audited and found nothing blocking. Notably it verified `creature-0.js`'s `#stage feTurbulence` dependency by diffing the **subtree nesting**, not just tag order — a re-parent would preserve the tag sequence and slip past the parity check above. It also confirmed the three "console-only" scripts really are, and tested the eslint glob with the actual matcher.


saeedkolivand:

PR #921, opened 2026-07-30
test: cover the agent-fleet components and scroll hooks

saeedkolivand:
## What

`AgentFleet.tsx` had **no test at all**. #916 split its 830 lines into six sections, three hooks and five small components with code review as the only gate — it was the last untested surface from the landing refactor batch. This closes it.

**49 cases** across `AgentFleet`, `NodeButton`, `Station`, `IntakeSection`, `FleetMapSection` and `TogetherSection`. Landing suite: **257 → 306**.

## The headline pair

#916 hoisted `NodeButton` out of the render body because a fresh component identity on every render made React unmount and remount every node, dropping focus and restarting transitions. Two tests pin that fix.

Rather than assume they work, I proved it: I temporarily re-inlined the real `NodeButton` into `FleetMapSection`'s render body — the actual pre-#916 shape — and ran the file. Both tests failed, with `document.activeElement` falling back to `<body>`:

```
AssertionError: expected <body><div>…(6)</div></body>
              to be <button type="button" …>…</button>
  FleetMapSection.test.tsx:153
  FleetMapSection.test.tsx:164
```

Then reverted; `git diff HEAD` on production files is empty. The two cases cover the two distinct state paths that trigger the re-render — hovering (`litName`) and clicking (`selected`).

## The rest

Targeted at wiring that review can't keep honest:

- **Section render order** — an ordered `toEqual` on the heading array, so a reorder fails, not just a drop.
- **`useReveal` across children** — the hook lives at the root while `.reveal`/`.draw` markup is in six sections. Driven by a synchronous fake `IntersectionObserver`, asserting elements from multiple sections, so moving the hook into one section fails.
- **`prefers-reduced-motion`** skips the `requestAnimationFrame` path.
- **Both copy chips** write the *copy* string, not the displayed one — those genuinely differ.
- **Every rendered list derives from `data/agent-fleet.ts`** (`STATIONS.length`, `AGENT_COUNT`, `it.each(ROUTES)`, `WORK_TOGETHER_TEAMS`), so adding data needs no test edit but dropping one from the render fails.
- **`Station`** renders both the horizontal `.station` and vertical `.belt-vert` shapes from one component.

`matchMedia`, `IntersectionObserver`, `requestAnimationFrame` and `navigator.clipboard` are stubbed deterministically; no assertion depends on layout geometry, since jsdom has none.

## Verification

`test` (306 pass) · `typecheck` · `lint:strict` · `prettier --check` — all clean. **No production code changed.**

`testing-reviewer` audited and ran the suite shuffled (`--sequence.shuffle`) and in isolation to rule out stub leakage and ordering dependence — both green. Its one MEDIUM finding is fixed here: the `WORK_TOGETHER_TEAMS` test asserted card and row *counts* but never segment content, so a dropped `<code>` wrapper or reordered segment would have passed. It now pins each row's flattened text and which segments are `<code>`.

### DIFF
--- a/apps/landing/src/components/agent-system/AgentFleet.test.tsx
+++ b/apps/landing/src/components/agent-system/AgentFleet.test.tsx
@@ -1,0 +1,171 @@
+// @vitest-environment jsdom
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { cleanup, render } from '@testing-library/react';
+
+import { AGENT_COUNT, STATIONS } from '@/data/agent-fleet';
+
+import { AgentFleet } from './AgentFleet';
+
+// jsdom provides neither matchMedia nor IntersectionObserver, and getBoundingClientRect
+// is all-zero — see the repo lesson on jsdom zeroing layout geometry. Every effect
+// under test here (useBeltScrub, useMapLinks, useReveal) reads matchMedia directly on
+// mount, so it must be stubbed for ANY render of AgentFleet/BeltSection/FleetMapSection,
+// not just the reveal-specific tests.
+function stubMatchMedia(opts: { mobile?: boolean; reduce?: boolean } = {}) {
+  vi.stubGlobal('matchMedia', (query: string) => ({
+    matches: query.includes('prefers-reduced-motion')
+      ? (opts.reduce ?? false)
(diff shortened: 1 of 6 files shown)


saeedkolivand:

PR #922, opened 2026-07-30
chore: add robots.txt, sitemap.xml and llms.txt to the landing site

saeedkolivand:
## What

The landing site shipped with no crawl metadata at all — no `robots.txt`, no sitemap, and nothing aimed at LLM crawlers. This adds all three.

- **`src/app/robots.ts`** — allows everything except `/mission-control` (a live GitHub-API dashboard that already sets `robots: { index: false }` in its page metadata; this repeats it at the crawl layer so bots skip the fetch), and points at the sitemap.
- **`src/app/sitemap.ts`** — the 8 indexable routes, each matching that page's own `alternates.canonical`. No `lastModified`/`priority`: Google ignores priority outright, and a build-time timestamp would churn the file on every deploy.
- **`public/llms.txt`** — follows the llms.txt convention (H1, `>` summary, key facts, `## Pages`, `## Optional`). `/world` and `/creature` sit under Optional since they are visual and carry no facts a crawler needs.

## Why llms.txt, given adoption is unproven

No major AI crawler has committed to reading `llms.txt`, and the bots that do hit us (GPTBot, ClaudeBot, PerplexityBot) crawl HTML and obey `robots.txt`. It is included anyway because the file is static, costs one bullet per new page, and pays off if adoption lands. `robots.txt` and the sitemap are the parts that actually get read today.

## Gotcha worth remembering

Next 16 metadata routes default to **dynamic**, and `output: 'export'` hard-errors on them (`export const dynamic = "force-static" ... not configured on route "/robots.txt"`). Both files need `export const dynamic = 'force-static'`. Commented in place.

## Verification

- `pnpm typecheck` — pass
- `pnpm build` — emits `out/robots.txt`, `out/sitemap.xml`, `out/llms.txt`; route table lists both new static routes
- `pnpm check:parity` — pass (27 signature phrases, 13 legacy links, 7 installer URLs, deploy-shape files intact)
- `pnpm lint:strict --max-warnings 0` — clean
- Pre-push gate (parity + agent-system drift + AI review) — clean

## Maintenance

Adding a page means one entry in `ROUTES` and one bullet in `llms.txt`. No drift guard added — `check:parity` already fails loudly if a page disappears, and a guard for two lists of eight felt like more machinery than the problem deserves.


saeedkolivand:
@claude review

saeedkolivand:
All three findings applied in `251cc56b`.

**🟠 MEDIUM — `Disallow` cancelled the `noindex`.** Correct, and it was a genuine error in my reasoning, not just a misleading comment: a crawler that obeys `Disallow` never fetches the page, so the `noindex` is unreachable and the bare URL stays indexable from external links. Dropped the rule — robots.txt is now `Allow: /` with no exceptions, and the comment explains why `/mission-control` must stay *fetchable* for its `noindex` to mean anything.

**🟡 LOW — canonical invariant held for 6 of 8.** Added `alternates.canonical` to `how-it-works/page.tsx` and `creature/page.tsx`, matching the other six pages. The invariant asserted in `sitemap.ts` now actually holds, so the comment stands as written.

**🟡 LOW — the guard already exists at 3 lines.** Fair — I argued against a bespoke guard and missed that `REQUIRED_FILES` was right there. Added `robots.txt`, `sitemap.xml`, `llms.txt`. Confirmed it bites rather than assuming: removing `out/robots.txt` fails `check:parity` with `passthrough file missing from out/: robots.txt`, and it passes again once restored.

**Nit (`SITE_ORIGIN`)** — agreed, left alone. Extracting it while `layout.tsx` and eight page canonicals still hardcode the origin would add an inconsistency rather than remove one.

Re-verified after the changes: `typecheck`, `build`, `check:parity`, `lint:strict --max-warnings 0`, plus the pre-push gate (parity + agent-system drift + AI review) — all clean.

saeedkolivand:

PR #67, opened 2026-05-21
feat: ollama onboarding step with hardware-based model recommendation

saeedkolivand:
## Summary

Adds a new **AI Setup** step between Prefs and Tour in the onboarding wizard:

- **Ollama detection** — checks system health on mount; shows spinner while checking
- **Not installed** — shows install instructions (3 steps) + Download Ollama button + recheck button
- **Installed, no models** — shows 4 model options with recommended badge based on detected RAM
- **Hardware detection** — reads total RAM via `system.getMetrics()` and recommends the best-fit model:
  - < 6 GB → Llama 3.2 (1B)
  - 6–10 GB → Llama 3.2 (3B)  
  - 10–12 GB → Mistral 7B
  - 12 GB+ → Llama 3.1 (8B)
- **Model pull** — animated progress bar while downloading; shows installed checkmark
- **Skip option** — skip with a visible warning that AI features won't work
- **Sets model in preferences** on continue

**AI Settings tab** also gains:
- Ollama status card (running ✅ / not found ⚠️ with download + recheck)
- Quick-pull list when Ollama is running but no models are installed

**Wizard changes:**
- Step 2 (Prefs) next button: "Set up AI" → leads to new Ollama step
- Step 3 (Ollama) next button: "See the features" → leads to tour
- Step dots updated from 3 to 4 across all steps

## Test plan

- [ ] Fresh install — wizard shows Ollama step with "not found" state
- [ ] With Ollama running but no models — shows model list with recommendation for your RAM
- [ ] Download a model from the wizard — progress bar animates, model selected on continue
- [ ] Skip — proceeds to tour with warning shown
- [ ] AI Settings → Ollama card shows correct status

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #923, opened 2026-07-31
ci: let both claude review lanes post inline pr comments

saeedkolivand:
## Why

`.github/workflows/claude-review.yml` has always told Claude to *"Post findings as inline comments on the changed lines"* in both lanes. **Neither lane could.** The instruction was a silent no-op — findings landed in the tag-mode tracking comment (Lane A1) or only in the sticky comment (Lane A2).

Ported from the same fix in [`crosskit#66`](https://github.com/saeedkolivand/crosskit/pull/66), verified against the action source at the SHA this repo already pins (`be7b93b1` = v1.0.183):

- The inline-comment MCP server starts **only** when `allowedTools` contains a `mcp__github_inline_comment__`-prefixed entry — `src/mcp/install-mcp-server.ts:80,138-142`.
- Tag mode's own list is `Glob, Grep, LS, Read, mcp__github_comment__update_claude_comment, mcp__github_ci__*` — `src/modes/tag/index.ts:123-133`. No inline tool.
- Agent mode passes `parseAllowedTools(claude_args)` to the **same** check (`src/modes/agent/index.ts`), and `pull_request` is an entity event with `isPR: true` (`src/github/context.ts:191-201`) — so the automatic gate can post inline too.

So the workflow's two explanatory comments were both wrong: tag mode does *not* grant the tools, and a `prompt` does *not* deny them. Naming the tool is the whole fix.

## What changed

| Lane | Change |
| --- | --- |
| **A1** — `@claude review` on-demand | `--allowedTools "mcp__github_inline_comment__create_inline_comment"` added to `claude_args`. All findings post inline. |
| **A2** — automatic `AI Review OK` gate | Same tool appended to the existing `--allowedTools`, plus one prompt line capping inline to **blocking** findings (HIGH/CRITICAL, confidence ≥ 0.8, `introduced_by_diff`). |

**Why A2 is capped:** that lane re-runs on every `synchronize`, and inline review comments are not deduplicated across runs — an unbounded rule would stack a fresh copy of every MEDIUM/LOW nit on each push. The threshold matches `scripts/ci-review-verdict.mjs:69` (`blockingFindings(findings, 0.8)`) so the inline comments and the red check can never disagree.

Not touched: the sticky comment stays the full-finding-list surface, the check status stays the blocking signal, `permissions:` are unchanged (`pull-requests: write` was already present in both jobs — that is what the inline server needs).

Deliberately **not** ported from crosskit#66: the fork-safety SHA gate, the `ref:` pin, and the install/build/Playwright steps. Those exist there only because that PR also makes the reviewer *run the test suites* — out of scope here.

Both stale comment blocks are replaced with the real mechanism plus source citations, so a future action bump can be re-verified. ADR-0008 repeated the same wrong claim twice (lines 13 and 55) and is corrected with a dated amendment.

## Verification

- `pnpm gen:workflows:check` — no catalog drift ✓
- `prettier --check` on both files ✓
- Pre-push gate (workflow-catalog drift, agent-system drift, AI review of push range) — clean ✓
- `actionlint` via `workflow-lint.yml` ✓
- The inline tool reaches the CLI — `--allowedTools "Read,Grep,Glob,Write,mcp__github_inline_comment__create_inline_comment"` is confirmed present in this PR's run log ✓

### ⚠️ Neither lane can be exercised on this PR

`claude-code-action` refuses to run when the PR modifies the workflow file that invokes it:

> Skipping action due to workflow validation: the workflow file must exist and have identical content to the version on the repository's default branch.

That is a deliberate guard against a PR rewriting its own reviewer, and it hits **both** lanes here — so `@claude review` on this PR will no-op too. The gate fail-opened exactly as designed (`Verdict: fail-open (infra)`, check stayed green, no false block). For contrast, the same job on #922 produced `0 finding(s)` normally — so this is specific to workflow-touching PRs, not a regression.

**Verify after merge, on the next ordinary PR:**

1. Lane A2 runs automatically. A diff with a genuine HIGH/CRITICAL finding should now produce inline comments alongside the sticky comment; a clean diff correctly produces none.
2. Lane A1: comment `@claude review`. Pass = no *"I can only update one comment"* opener, and `gh api "repos/saeedkolivand/ai-job-hunter-app/pulls/<N>/comments"` returns `claude[bot]` rows. This is how the crosskit fix was confirmed — its PR #67 received 6 inline comments once #66 merged.


saeedkolivand:

PR #924, opened 2026-08-01
feat: add an accessibility statement and fix the barriers it found

saeedkolivand:
Adds `/accessibility` — a [WAI-model](https://www.w3.org/WAI/planning/statements/) accessibility statement covering the landing site, the desktop app and the browser extension — **and fixes every WCAG A/AA violation it found**.

## Why, and why it doesn't overclaim

Publishing this isn't legally required here — the EAA's microenterprise exemption covers a solo maintainer's free service, and no US statute requires one from a private site. It ships because this is a job-hunting tool whose users include disabled job seekers, and because a false "fully conformant" claim would be worse than no page at all.

So every claim on the page traces to a measured result, not boilerplate.

## What the audit found, and what it is now

axe-core 4.12 (WCAG 2.0/2.1/2.2 A+AA tags) against the **built** `out/`, plus a scripted keyboard tab-through:

| surface | before | after |
| --- | --- | --- |
| landing, 9 pages | 3 serious rules | **0** |
| desktop home view | 1 critical + 1 serious, 12 nodes | **0**, both colour schemes |

**Landing fixes**

- `/creature` — the faded doodle annotations diluted `--ink` toward the paper background via opacity. Now an explicit warm grey at **4.98:1**, up from 2.46–2.8:1.
- `/world` — the active nav pill darkens its accent with `color-mix`. White on the darkened red is **8.54:1**, up from 3.93:1. The accent cycles per section, and the **blue** state was **1.88:1** — a worse failure the original single-state scan never saw, now 4.86:1. Found by reading the code, not by scanning.
- `/architecture-map` — the sidebar is keyboard-reachable and named. The map's window-level arrow-key pan handler now excludes it, so arrows scroll the sidebar instead of panning the map underneath it. Verified in a browser (focus lands on `#side`, ArrowDown moves `scrollTop` 0 → 80) and covered by a new test that fails against the pre-fix code.

**Desktop fixes**

- The critical `aria-allowed-attr` was `aria-expanded` on a roleless wrapper `div` in the shared `packages/ui` `HoverPopover`. The invalid attribute is now gone from that wrapper everywhere the component is used, and moves onto the trigger where the trigger is a real button — so this cleared the same latent defect at every call site, not just the home view.
- The 11 contrast failures use the existing `text-muted-foreground` token (5.1:1 on white, 4.8:1 on `#f8f8f8`, clear in dark too) instead of low-alpha `text-foreground/30|40|55`.

## What the statement still discloses

Zero automated violations is **not** conformance, so both surfaces stay **partially conformant** and the extension stays **not assessed**. The page states plainly that:

- **the contrast fix is scoped to the home view** — ~460 of the renderer's ~1,100 `text-foreground/NN` usages sit at sub-AA steps, and no other view has been scanned. A global migration is a wide visual diff that needs its own PR and a Lost Pixel review;
- two `best-practice` rules remain on the home view (`landmark-unique`, `region`) — outside WCAG A/AA, but known;
- the three CI a11y gates are **advisory, not blocking**;
- exported PDFs are **not tagged to PDF/UA-1**;
- automated tooling catches roughly a third of WCAG issues, there has been **no third-party audit**, and **no testing with assistive-technology users**.

## The statement page itself

axe 0 violations · 24/24 tab stops with a visible ring · `H1` + 9×`H2`, no skipped levels, no duplicate ids · anchors reveal on focus · no overflow at 320px under `prefers-reduced-motion: reduce`.

Reuses `privacy.css` rather than adding a stylesheet. Linked from the shared footer, the home footer, `sitemap.ts` and `llms.txt`. `check-parity.mjs` untouched — its lists are a frozen legacy snapshot and its checks are presence-only.

## Review

Three cycles with `ui-ux-expert`, which **blocked once**: the copy claimed a sitewide manual keyboard pass that hadn't happened. Rather than soften the sentence, the pass was run. Each cycle caught one further overclaim on the same lines — a reduced-motion count that implied 9 stylesheets were all of them, a focus-ring baseline credited to tokens that define none, and an `aria-expanded` relocation described as universal when two call sites lose the attribute instead. All corrected.

Two apparent findings were discarded as artifacts of the measuring script rather than written into the page: "keyboard traps" on `/privacy` and `/accessibility` were the nine identical `#` heading anchors colliding in a string dedup key, and 142 "ring-less" nodes on `/architecture-map` were the SVG map nodes using the thickened `rect.box` stroke that `architecture-map.css:509` deliberately swaps in for `outline`.

## Follow-ups filed in `NEXT_ISSUES.md`

The `text-foreground/NN` ramp migration · the two landmark rules · **no permanent a11y gate for `apps/landing`** (every scan here ran from throwaway scripts under `apps/desktop/`, so nothing re-checks these pages on future changes) · the unscanned extension · PDF/UA-1 tagging.


saeedkolivand:
@claude review

saeedkolivand:
Applied 4 of the 6 findings (MEDIUM + 3 LOW); the 2 skipped are noted below.

**Applied**

- **MEDIUM · `SECRET_SCAN_FILES`** — the five new accessibility components are registered in `check-landing-drift.mjs` alongside the privacy ones. `pnpm check:landing-drift` passes with them in scope.
- **LOW · `DocShell` footer** — added. Your argument beat my original one: `/architecture-map` is exactly the page that carried the keyboard barrier, so someone who hit it had no in-page route to the statement or the feedback address.
- **LOW · anchor test** — both directions now derived from the DOM, so a typo'd `href` fails instead of staying green.
- **LOW · stale-date risk** — the date is now a single `LAST_UPDATED` constant with a test that fails once it is over 12 months old and tells whoever trips it to re-review the claims before bumping. The rendered prose is unchanged.

**Skipped, deliberately**

- **LOW · route in four places** — a pre-existing divergence (`/world`, `/creature`, `/architecture-map`, `/agent-system` are already sitemap-only). Not worth a refactor for one route.
- **LOW · selector drift on the barrier list** — real, but this is better served by the landing a11y gate already filed in `NEXT_ISSUES`. Worth noting your prediction landed: `#side` did gain `tabindex`, and the copy was updated in the same PR.

**Heads up on scope** — this review ran against the statement-only commit. Since then the PR also fixes every barrier the statement documented (landing 9 pages and the desktop home view are both at 0 WCAG A/AA now, the latter in both colour schemes), which touches `packages/ui/HoverPopover`, the architecture-map keyboard handler, `creature.css`, `scrub-engine.js` and three renderer components. That code has not been reviewed here.

@claude review

saeedkolivand:

PR #925, opened 2026-08-01
ci: add a blocking accessibility gate for the landing site

saeedkolivand:
Follow-up to #924. That PR fixed every WCAG A/AA violation on the landing site and published `/accessibility`, a public page that **states those results**. Nothing re-checked them — so the fixes could regress silently and the statement would start lying. Every scan in #924 ran from throwaway scripts under `apps/desktop/`, the only place `@axe-core/playwright` already resolved. This makes it permanent.

## The gate

`apps/landing/scripts/check-a11y.mjs` serves the built `out/` with `node:http` (no server dep, mirroring `check-parity.mjs`), drives headless Chromium + axe-core, and scans **every route discovered by globbing `out/**/*.html`** — never a hardcoded list. A hardcoded list is the exact failure mode #924's review flagged, where a new route had to be added by hand in four separate places to be covered at all. Auto-discovery means a new page is covered the day it exists.

Tag set is `wcag2a/wcag2aa/wcag21a/wcag21aa/wcag22aa` — the same set `/accessibility`'s published numbers were measured with, deliberately not widened to best-practice rules.

**It blocks.** That differs from the three desktop a11y checks on purpose, and the file header says so in case someone later tries to harmonise it down:

| check | behaviour | why |
| --- | --- | --- |
| `e2e/a11y.spec.ts` | logs, never fails | surface isn't clean yet |
| `eslint.a11y.config.mjs` | warn-only | surface isn't clean yet |
| `.lighthouserc.json` | warn assertion | surface isn't clean yet |
| **`check:a11y`** | **fails the build** | **baseline is 0 — anything found is a regression against a public claim** |

Runs on **pull requests**, inside the existing test job, reusing the Chromium already installed there for the Storybook vitest project. No new job. (Note `check:parity` runs only in `pages.yml` on push-to-main — post-merge, which would be too late for this.)

## What it found on its first run

Auto-discovery turned up **13 routes, not the 9** previously scanned. Three had never been checked:

- **`/404` and `/_not-found`** — clean.
- **`/mission-control`** — `color-contrast` at **4.03:1** on row-title links, 5 nodes. **Fixed.** `docs-tokens.css` already defines `--doc-accent-text` for exactly this case, and `.mc-badge.is-stale` documents the identical failure. Needed `.doc-shell .mc-row__title a` to out-specify the `.doc-shell a` base rule, which wins on source order because the page renders its own `<PageStyle>` before `DocShell`'s. Now **5.7:1**.
- **`/benchmarks/`** — `color-contrast` 3.15:1 on `#dl-button` plus a missing `lang` on `<html>`. **Excluded, not fixed** — it's generated by `benchmark-action` and overwritten wholesale by `quality.yml`'s `benchmark` job on every push-to-main perf run, so an in-repo edit would be reverted by CI. `EXCLUDED_ROUTES` names the generating workflow and both violations, and the gate **logs the exclusion** rather than skipping silently.

## Statement updated to match

A gate finding a barrier it can't fix is exactly the case `/accessibility` exists for, so the page discloses it rather than quietly keeping a clean number:

- "0 violations across all 9 pages" → **12 routes** the permanent gate covers, with `/404`, `/_not-found` and `/mission-control` named.
- `/mission-control`'s contrast failure added to the fixed-barriers list.
- New card disclosing `/benchmarks/` — a public page on the domain with two violations, excluded because it's third-party-generated. Stated as *excluded, not passing*.
- The gate added under "What's in place", explicitly contrasted as **blocking** against the three desktop checks — whose wording is untouched, because they didn't change.

**Every prior disclosure stands.** Both surfaces remain **partially conformant**; the extension is still *not assessed*; PDFs are still not PDF/UA-1 tagged; the desktop contrast fix is still scoped to the home view (~460 sub-AA `text-foreground/NN` usages elsewhere); still no third-party audit and no AT-user testing. No conformance verdict was upgraded — the `partially conformant` and never-`fully conformant` regression tests still pass.

## Proof the gate bites

A gate that can't fail is worse than no gate, so this was verified rather than assumed: lightening `--red` in `privacy.css` made it fail on both `/privacy` and `/accessibility` (which reuses that stylesheet) with exact selectors, fg/bg colours and measured-vs-required ratios, exit code 1. Reverted, back to green.

```
check:a11y — excluding 1 route(s) from the gate (see EXCLUDED_ROUTES): /benchmarks/
check:a11y OK — 12 page(s) scanned, 0 WCAG A/AA violations (1 page(s) excluded, see EXCLUDED_ROUTES).
```

`@axe-core/playwright` is declared explicitly in `apps/landing` rather than relying on `shamefullyHoist: true` — it resolves either way, but a gate's dependencies shouldn't hinge on a global pnpm setting.

## Cost note

The test job now also builds the landing static export. Chromium is already installed and cached there, so the added cost is the Next build. If that becomes a drag, the build output could be cached or the gate split into its own job.

### DIFF
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -524,6 +524,28 @@ jobs:
         continue-on-error: true
         uses: davelosert/vitest-coverage-report-action@8b157684c6a6b259b97d45e72b44242865c0f6a5 # v2.12.2
 
+      # Build the landing static export once, then reuse it for the a11y gate
+      # below — no separate job, reuses the Chromium already installed above
+      # for the Storybook vitest project. Unlike the desktop's advisory a11y
+      # checks, this one BLOCKS the build (see check-a11y.mjs's header for
+      # why: the landing site's WCAG A/AA baseline is 0, so any violation
+      # found here is a regression on the published /accessibility claim).
+      - name: 🏗️ Build landing (static export)
+        id: landing-build
+        continue-on-error: true
+        run: |
+          echo "::group::Building @ajh/landing (static export)"
+          pnpm --filter @ajh/landing build
+          echo "::endgroup::"
+
+      - name: ♿ Landing Accessibility Gate
+        id: landing-a11y
+        continue-on-error: true
+        run: |
+          echo "::group::Scanning out/ with axe-core (WCAG A/AA)"
+          pnpm --filter @ajh/landing check:a11y
+          echo "::endgroup::"
+
       # cargo-llvm-cov collects coverage; cargo-nextest is the test runner
       # (parallel, flaky-retry, faster, clearer output). `cargo llvm-cov nextest`
       # runs the suite ONCE instrumented; `report` formats it afterward. This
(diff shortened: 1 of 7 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #926, opened 2026-08-01
feat: diagnose the role a cover letter answers before drafting it

saeedkolivand:
Folds a company-and-vacancy intelligence pass into the existing single-pass cover-letter generation. No extra AI calls, no new IPC, no new UI — the two research steps already had homes, only their content was thin.

## Company analysis
`commands/ai_provider/research.rs` — the opt-in "Research company" brief (provider-native web search: OpenAI / Anthropic / Gemini natively, Ollama via the Web Search API) now also covers **competitors** and the **challenges / strategic priorities the company is working on**, and carries a fact-vs-inference rule: state a fact plainly only when a search result supports it, hedge what was inferred, omit a facet rather than guess. Brief 120-150 → 150-200 words; the Ollama search query asks for competitors/strategy too.

`buildCompanyResearchBlock`'s fence cap moves 1200 → 1600 chars so the longer brief is not sliced mid-sentence.

## Vacancy analysis + matching
`cover-letter.ts` WRITING NOTES — a private diagnosis before drafting:

1. **why this role is open** — the business problem it exists to solve, read off the ad's own signals (what it repeats, what it lists first, scope, seniority, team, what "you will own") plus the brief when present
2. **the first 6 to 12 months** — the 3 outcomes the hire is judged on
3. **the through-line** — which real résumé achievements make the candidate part of that solution

Every step is bound to evidence in `<job_ad>`, `<company_research>`, or `<candidate_resume>`. Thin evidence keeps the diagnosis broad instead of guessing — an invented problem or piece of company news is worse than a general one — and the letter voices it as the candidate's own reading of the role, never insider knowledge. The "heart" movement now asks achievements to prove the candidate can solve that problem. The same one-liner goes into the agent-mode `draft_cover_letter` prompt.

The diagnosis names `<company_research>` only when a brief is actually fenced above; without one it runs on the ad + résumé alone.

## Drive-by
Drops the hardcoded "200 to 300 words" from four places — it contradicted `<market_conventions>`, which already carries the per-market band (intl 200-350, DE 250-400).

## Deliberately not done
- **No interactive 3-question intake** — the app already has company / JD / CV / country / language as structured inputs.
- **No second AI call for the vacancy analysis** — the model has the ad + brief in context; a round-trip buys latency, not insight. Worth adding only if the diagnosis should be *shown* to the user rather than only used.
- **No follow-up questions on missing evidence** — that needs a UI turn mid-generation; the prompt degrades to "keep it broad" instead.

## Verification
`@ajh/prompts` 555 tests · `cargo test --lib` (research + agent::tools) · `pnpm typecheck` · `pnpm lint` · `cargo clippy --all-features` · `cargo fmt --check` · pre-push gate incl. AI review: clean.

## Summary by CodeRabbit

* **New Features**
  * Improved cover-letter generation by tailoring content to the role’s business needs and likely success criteria.
  * Added competitor and strategic-priority insights to company research.
  * Expanded research context and adapted letter length to market conventions.

* **Bug Fixes**
  * Improved handling of unsupported company information by distinguishing verified facts from inferences.

### DIFF
--- a/apps/desktop/src-tauri/src/agent/tools.rs
+++ b/apps/desktop/src-tauri/src/agent/tools.rs
@@ -311,6 +311,13 @@ fn match_resume_handler(
 /// agent-context version of the `@ajh/prompts` cover-letter builder — the
 /// honesty/no-fabrication spine, grounded in the fenced résumé, untrusted brief
 /// used for company facts only.
+///
+/// The 200-300 word band here is deliberately market-agnostic and deliberately
+/// NOT the per-market band the TS builder uses: this path has no
+/// `<market_conventions>` block to defer to (no resolved market reaches the
+/// agent tool), so a fixed, safe middle is the only option. Expect the same
+/// posting to yield a slightly different length through the desktop flow —
+/// that is the intended trade, not drift to "fix".
 const COVER_LETTER_SYSTEM: &str = "\
 You are a cover-letter writer. Write ONE focused, specific cover letter (about 200-300 \
 words of body) that reads like a real person wrote it: flowing prose, not a list of \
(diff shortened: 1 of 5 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #927, opened 2026-08-02
feat: add opt-out crash reporting via sentry

saeedkolivand:
Adds remote crash reporting to the desktop app. Plan: `.claude/plans/polished-singing-graham.md`, decision recorded in **ADR-0020**.

**This reverses a published promise.** The privacy policy named Sentry, by name, as absent in three places, and ADR-0005 defined the local-first guarantee as "no telemetry". That was a deliberate product decision by the owner, not an oversight — the docs are rewritten here in the same PR because `check:parity` and `check:landing-drift` fail if code and copy disagree.

## Why

The app was blind in production. The panic hook wrote `crashes.log` and the diagnostics zip could export it, but both required the user to notice a failure, find the export, and file an issue by hand. React errors caught by a boundary never reached `window` at all, so they were invisible even in principle.

## Scope

Desktop app **only**. The extension keeps its AMO `data_collection_permissions: ['none']` declaration true, and the landing site keeps its no-third-party-JS origin invariant (a foreign script there could read the mission-control PAT from `localStorage`).

## The four things that make "default ON" defensible

1. **Consent gate.** Transmission requires `enabled && consentShown`, not just `enabled`. The default is on, but nothing is sent until the setup wizard has actually shown the choice. A user who never finishes onboarding never reports anything — failing closed.
2. **Redaction.** Every event is serialized, every string passed through the same `redact_token` pipeline the diagnostics bundle uses (ADR-027), then deserialized. Whole-event rather than field-by-field, because a field list is a denylist that rots as the SDK adds fields. `send_default_pii` off; `server_name` overridden (the SDK otherwise sends the machine hostname, often the owner's real name).
3. **No DSN outside release.** `build.rs` bakes it from an env var set only in the signed release job, read via `option_env!`. Local builds, contributor clones and every CI check compile to `None` and cannot transmit.
4. **Extension excluded**, as above.

## Notable implementation decisions

- **Consent is Rust-owned, not a renderer preference.** `sentry::init` must run before `tauri::Builder` because `sentry-rust-minidump` forks the crash-reporter process at startup and `panic = "abort"` leaves nothing in-process able to flush a crash. There is no WebView that early, so no `localStorage` — hence a small JSON file plus an IPC pair.
- **No CSP change.** The plugin's browser side uses a custom transport that forwards WebView events to the Rust SDK over `invoke`, so the renderer never egresses. Widening `connect-src` is rated HIGH/CRITICAL in this repo; it stays untouched.
- **`sentry` is pinned to 0.42, not 0.49.** `tauri-plugin-sentry` 0.5 (the latest) builds against 0.42; a newer `sentry` links a second, type-incompatible copy and `init(&client)` stops accepting our client.
- **No hand-rolled `window.onerror`.** The injected browser SDK already installs global handlers; a second listener would double-report. Only React 19's `onCaughtError` is wired, because boundary-caught errors never reach `window`.
- **Release profile now keeps symbols** (`debug = "line-tables-only"`, `split-debuginfo = "packed"`, `strip = "debuginfo"`) and CI uploads sidecars to Sentry — never attached to the release, since full DWARF for this dep tree is large.

## Docs rewritten

All twelve claim sites, plus ADR-0005 (eighth egress class) and new ADR-0020. Two worth calling out:

- **`Footer.tsx`** claimed *"Because there's no server-side copy, deleting locally deletes it everywhere"* — flatly false once reports exist. Replaced with the erasure path.
- **`/privacy` is the policy URL filed with both Chrome Web Store and Firefox AMO.** It now states desktop collection and extension non-collection as separate sections, so a reviewer cannot read one as the other while the manifest still declares `['none']`.

## Verification

`cargo test --lib` (5 new crash_reporting tests incl. the redaction gate) · `cargo test --test architecture` · `cargo clippy --all-features` · `cargo fmt --check` · `pnpm test` 4636 passed · `pnpm typecheck` · `pnpm lint` · `check:parity` against a fresh landing build · `check:landing-drift` · pre-push gate incl. cargo-deny and cargo-machete.

Claim sweep (`rg -i "no telemetry|no analytics|phones home|No Sentry"`) returns only extension-scoped hits and historical commit data.

## Known limitations, stated not hidden

- Turning reporting off unbinds the client immediately, but the minidump supervisor is a separate process forked at launch and only stops at restart. The settings copy says so.
- **5k errors/month is a quota on errors, not users.** One crash loop can exhaust it in a day. Worth setting a client `sample_rate` and per-issue server-side rate limiting before the user base grows — noted in ADR-0020, not done here.

## Needs before this actually reports

Three repo secrets: `AJH_SENTRY_DSN`, `SENTRY_AUTH_TOKEN`, `SENTRY_ORG`, `SENTRY_PROJECT`. Without them the build compiles inert and the symbol-upload step skips itself — nothing breaks.


saeedkolivand:
@claude review

saeedkolivand:
Fixed in `848c0253`.

Both lanes flagged the same thing and it was real: `init()` resolved the consent file through `platform::config::data_dir()` before `setup()` had exported `AJH_DATA_DIR`, so it read `$HOME/.ajh` while the commands wrote to Tauri's app-data dir. No user choice would ever have been observed, and it would have failed **silently** — a missing file is indistinguishable from "user said no", so the feature would have looked fine while never activating.

Rather than making the two call sites resolve alike, the directory is now resolved once into a `OnceLock` and `load`/`save`/`clear` all funnel through it, so agreement is structural. The commands no longer take an `AppHandle`.

Added `every_accessor_shares_one_resolved_directory` as a regression test: it pins that repeated resolution is identity-stable and that a `save()` is observable through `load()` — the exact property that was broken.

Verified: `cargo test --lib crash_reporting` (6 passed), `cargo test --test architecture`, `clippy --all-features`, `fmt`, `pnpm typecheck`.

@claude review

saeedkolivand:
Both fixed in `2a6d8778`.

**HIGH — factory reset kept transmitting.** Correct and it directly contradicted my own settings copy. `clear()` only affects the *next* launch, and a factory reset doesn't restart the process, so a consented user kept reporting for the rest of the session right after what reads as a full privacy wipe. Now calls `disable_current()` alongside the clear, mirroring the opt-out path.

**MEDIUM — missing capability entry.** This was worse than "unlike every other plugin": I checked the plugin's generated schema and its browser side forwards events to Rust via two permission-gated `invoke` commands (`envelope`, `breadcrumb`). Without `sentry:default` every renderer error would have been denied at the IPC boundary and silently never reported — half the feature dead while looking correctly wired. Added to `capabilities/default.json`; `tauri-build` validates the identifier at build time, so a wrong one fails the build.

Verified: 6 crash-reporting tests, architecture test, `clippy --all-features`, `fmt`, and a build-script rerun to exercise capability validation.

Worth noting both of this round's findings shared a failure mode with the first one: the feature would have looked fine and reported nothing. That's the risk profile of a fail-closed design, and it's why I'd treat a manual end-to-end check — throw a renderer error against a real DSN and confirm it lands — as the gate before trusting this in production, not the test suite.

@claude review

saeedkolivand:
Round three addressed in `e0938928` — one accepted, one partly wrong, one rejected with reasons.

**MEDIUM (dependency bloat) — half right, and the right half was worth fixing.** I verified before acting: `cargo tree -i actix-web` returns nothing, so the actix integration was never pulled — defaults don't enable it. The duplicate `reqwest 0.12.28` alongside the app's 0.13.1 was real, and my `sentry = "0.42"` was the only thing pulling it, since both `tauri-plugin-sentry` and `sentry-rust-minidump` declare sentry with `default-features = false`.

Fixed by switching to the `ureq` transport rather than just trimming features, because dropping `transport` without a replacement leaves no transport at all. That is also strictly more correct here: the reqwest transport requires tokio, and the minidump supervisor is a bare forked process with **no async runtime to send on**. Verified `reqwest@0.12` is gone and `ureq 3.3` replaces it.

**HIGH (STATE_DIR pins the fallback dir) — the fact is correct, the suggested fixes are not possible.** The file does land in `$HOME/.ajh` rather than the per-OS app-data dir. All three proposed remedies require an `AppHandle` that does not exist yet:

- resolving from `resolve_and_export_data_dir` — that runs inside `setup`, strictly after `init`;
- deferring the OnceLock until after setup — `init` needs the value *before* setup to decide whether to construct the client at all;
- moving `init` later — the minidump supervisor re-executes everything above its own call in the forked child, so a late fork makes the child build a second Tauri app.

So the placement is forced. What was genuinely defective was my module doc claiming the file sits "next to the other app data" — that was false. Corrected it, documented why the location is forced, and recorded the real consequence in ADR-0020: **deleting the app-data directory by hand leaves this file behind.** It holds two booleans and no personal data, and the factory reset clears it, so nothing identifying survives — but the asymmetry is now written down instead of implied away.

If you'd rather the file live in the canonical app-data dir, the only honest route is duplicating Tauri's per-OS path algorithm in `platform/config.rs` — which is exactly the fragile duplication that module exists to prevent, and would reintroduce a silent split-brain the day the two implementations diverge. I'd take the documented asymmetry over that.

Verified: 6 crash-reporting tests, `clippy --all-features`, `fmt`, dependency tree.

@claude review

saeedkolivand:
Dismissing the remaining MEDIUM as factually incorrect, with the evidence it asked for.

Ran exactly the suggested command plus `--target all` (my earlier check only covered the host target, so the reviewer was right to push on it):

```
$ cargo tree --target all | rg -o 'reqwest v[0-9.]+' | sort -u
reqwest v0.13.1

$ cargo tree --target all -i actix-web
warning: nothing to print.

$ cargo tree --target all -p sentry --depth 1 | rg 'reqwest|ureq|actix'
└── ureq v3.3.0
```

And the resolved feature set for sentry is `backtrace, contexts, debug-images, httpdate→ureq, native-tls, panic, release-health` — no `reqwest` feature, no `transport`, no actix. Across every target there is exactly one reqwest in the tree (0.13.1, the app's own) and sentry's only HTTP dependency is ureq.

So the comment's claim holds as written: the `ureq`-only selection does keep a second stack out. The finding looks like it was carried forward from the pre-fix state.

---

**Gate is now green** — `AI Review OK`, no blocking findings, `mergeStateStatus: CLEAN`.

Summary of the three rounds, all of which shared one failure mode — the feature would have looked correct and reported nothing:

1. **Consent directory split-brain** — written to app-data, read from `$HOME/.ajh`. Fixed by resolving once into a `OnceLock` so agreement is structural, with a regression test pinning it.
2. **Factory reset kept transmitting** for the rest of the session, contradicting the settings copy. Now unbinds.
3. **Missing `sentry:default` capability** — the plugin's `envelope`/`breadcrumb` invoke commands are permission-gated, so every renderer error would have been denied at the IPC boundary. Half the feature was dead while looking wired.

Plus the transport swap, which removed a duplicate reqwest and fixed a latent problem nobody flagged: the default transport needs tokio, and the minidump supervisor is a forked process with no runtime to send on.

**Still unverifiable by tests:** none of this proves an event actually arrives. That needs the `AJH_SENTRY_DSN` / `SENTRY_AUTH_TOKEN` / `SENTRY_ORG` / `SENTRY_PROJECT` secrets and one manual end-to-end check. Until then the build compiles inert by design.

saeedkolivand:

PR #928, opened 2026-08-02
fix: bump the privacy policy last-updated date and centralize it

saeedkolivand:
Follow-up to #927. I rewrote what `/privacy` says the app collects and left **`Last updated: 30 June 2026`** untouched.

That page is the privacy-policy URL filed with **both** the Chrome Web Store and Firefox AMO. A stale date there isn't cosmetic — it asserts a review that predates the change it describes, on a document store reviewers read against the code.

## Fix

Bumped to **2 August 2026**, and moved it into a single exported `LAST_UPDATED` constant in `sections/Footer.tsx`, right next to the "we'll bump the date" promise it backs. That mirrors `components/accessibility/sections/Footer.tsx`, which already had this shape — the privacy page was the odd one out, hardcoding the date in a different file from the promise about it.

Added a test pinning that the page renders the constant in `D Month YYYY` form, so a hardcoded date can't creep back in.

## What the test deliberately does not claim

The accessibility page enforces a 12-month staleness cutoff because it *promises* annual review. The privacy page promises something different — "if this policy changes, we'll bump the date" — which is change-triggered, not cadence-based. Nothing mechanical ties a copy edit to a date bump, so I didn't fake a guard that would either fail spuriously after 12 quiet months or give false confidence. The test comment says so outright, and the miss it guards against is named in it.

## Verification

`pnpm --filter @ajh/landing test` 325 passed · fresh landing build with the date confirmed in `out/privacy.html` · `check:parity` · `pnpm typecheck` · `pnpm lint`.


saeedkolivand:

PR #929, opened 2026-08-02
fix: fail the release when a configured sentry symbol upload breaks

saeedkolivand:
The symbol upload was `continue-on-error: true`, so a wrong token or a rejected upload would skip silently — and the first you'd know is a crash report full of unresolved addresses. That is the *same* silent-failure mode this feature was reviewed for three separate times in #927.

Now that the Sentry secrets are actually provisioned, "skip quietly" only makes sense for forks.

## Changes

- **Dropped `continue-on-error`** and added `set -euo pipefail`, so any failure past the secrets guard turns the run red.
- **Moved the step after `Upload Release Assets`.** The installers are published first, so a Sentry outage or an expired token never costs a release — but the run still fails visibly afterwards. Symbols associate by debug-id, not by upload order, so nothing is lost by uploading second.
- **Absent secrets remain a legitimate skip** (a fork, or a release cut before provisioning), now emitted as a `::warning::` annotation rather than a plain log line — so **"not configured" stays distinguishable from "configured wrong"**, which is the distinction that was collapsed before.

## Verification

Parsed the workflow and asserted the invariants rather than eyeballing them:

```
YAML parses OK
assets step index: 19 | symbols step index: 20
ORDER OK: symbols upload after assets
continue-on-error: (absent - fails loudly)
```

ADR-0020 records the trade-off under Consequences.

## Note

This makes a *configured-but-broken* pipeline loud, which is the point. It cannot make the first release pass or fail on its own — whether the pasted token value is correct is still only knowable from that first run, where Sentry's "Last access" column flips from *never used*.

## Summary by CodeRabbit

* **Release Process**
  * Release assets are uploaded only after successful builds.
  * Debug symbol uploads now provide clearer handling for missing or invalid Sentry configuration.
  * Configured upload failures are reported and can fail the release job, while missing credentials are skipped with a warning.

* **Documentation**
  * Updated crash-reporting guidance to reflect the release and symbol-upload behavior.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -412,39 +412,6 @@ jobs:
       # for this dep tree (typst, chromiumoxide, rusqlite) is very large, which
       # is the same reason the failure-log upload below is narrowly scoped.
       #
-      # Skipped rather than failed when the org/token secrets are absent, so a
-      # fork or a release cut before Sentry is provisioned still ships installers.
-      - name: 🔍 Upload debug symbols to Sentry
-        if: success()
-        continue-on-error: true
-        env:
-          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
-          SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
-          SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
-        run: |
-          # Guarded in the shell, not in `if:` — neither the `secrets` nor a
-          # step's own `env` context is available to a step-level `if`.
-          if [ -z "$SENTRY_AUTH_TOKEN" ] || [ -z "$SENTRY_ORG" ] || [ -z "$SENTRY_PROJECT" ]; then
-            echo "Sentry secrets not configured — skipping symbol upload."
-            exit 0
-          fi
-          echo "::group::Uploading debug files + sourcemaps"
-          VERSION="${{ steps.resolve.outputs.version }}"
-          npx --yes @sentry/cli@^2 releases new "${VERSION#v}"
-          # Native debug files: PDB (Windows), dSYM (macOS), DWARF/dwp (Linux).
-          # `split-debuginfo = "packed"` puts them beside the binary in target/.
-          # No `--include-sources`: that would ship our source into Sentry, and
-          # line tables alone are enough to resolve a frame.
-          npx --yes @sentry/cli@^2 debug-files upload \
-            apps/desktop/src-tauri/target/
-          # Renderer sourcemaps — emitted as 'hidden', so they exist on disk here
-          # but were never referenced from the shipped bundle.
-          npx --yes @sentry/cli@^2 sourcemaps inject apps/desktop/dist/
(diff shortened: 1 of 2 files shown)


saeedkolivand:
The review gate caught a real regression in my own fix — pushed `b5176a40`.

**Hard-failing the step cascades into a broken updater.** It sits inside a 4-leg matrix, and `generate-update-manifest`, `update-cask` and `update-download-page` all hang off a bare `needs: build` with no `if:`. So a Sentry hiccup on a single leg flips the aggregate result and **skips all three** — `latest.json` is never written, auto-update breaks for every existing user, and the installers sit attached to the release looking fine.

I traded unreadable stack traces for a broken updater. That is strictly worse and the exact opposite of the intent I wrote in the PR description.

**Corrected approach:** `continue-on-error` is restored, and failure is surfaced by an `::error::` annotation plus a `$GITHUB_STEP_SUMMARY` entry naming the platform, the version, and which secrets to check. Loud and unmissable, but non-gating.

**I did not take the gate's second suggestion** — loosening the three downstream jobs to `if: always() && needs.build.result != …`. That would publish an updater manifest even when a platform's build *genuinely* failed, pointing users at artifacts that don't exist. Trading a Sentry-shaped failure for a build-shaped one isn't an improvement.

Asserted the invariants rather than eyeballing:

```
order assets<symbols<flag: true (19,20,21)
symbols continue-on-error: true
flag if: always() && steps.sentry_symbols.outcome == 'failure'
flag has no exit 1: true
downstream still strictly gated: true
```

ADR-0020 now records why the asymmetry exists, so the next person doesn't 'fix' the `continue-on-error` back off — which is precisely what I did an hour ago.

saeedkolivand:
Both review comments handled.

**@claude HIGH (cascade into the release pipeline)** — already fixed in `b5176a40`, pushed before this review landed; it was reviewing the earlier commit. Same finding the Stop gate raised. `continue-on-error` is restored and failure is surfaced through an `::error::` annotation plus a step-summary entry, so the matrix leg stays green and `generate-update-manifest` / `update-cask` / `update-download-page` still run. I did not loosen those three to `if: always()` — that would publish an updater manifest even when a platform's build genuinely failed.

**CodeRabbit 🟠 Major (ADR duplicating the workflow)** — fair, and fixed in `bd8ce17e`. The `docs/**` thin-pointer rule is real and I'd copied the execution contract into the ADR where it can drift.

Applied it surgically rather than wholesale, though: I cut the *mechanics* — ordering, `continue-on-error`, the annotation plumbing — and kept the *rationale*, because that is what an ADR is for and it is the load-bearing part here. The bullet now says why a failed upload is loud rather than fatal, and why making the downstream jobs tolerant was rejected instead, then points at the workflow as authoritative.

Replacing the whole bullet with a bare link, as the suggested diff proposed, would have deleted the reasoning — and this is precisely a decision someone will otherwise 'clean up'. I know, because I removed that `continue-on-error` myself an hour ago for exactly the reason the ADR now warns against.

Verified no workflow mechanics remain in the ADR:

```
$ rg -n 'continue-on-error|::error::|step-summary|4-leg matrix|needs: build' docs/adr/0020-crash-reporting.md
(no matches)
```

saeedkolivand:

PR #930, opened 2026-08-02
fix: upload only our own debug files to sentry

saeedkolivand:
Found by inspecting the first real release run ([30741386889](https://github.com/saeedkolivand/ai-job-hunter-app/actions/runs/30741386889)).

## The good news first

Everything from #927/#929 worked on all four matrix legs: symbol upload `success`, flag step correctly `skipped`, assets published before symbols, no cascade into the updater jobs, `Created release 0.132.0`. **The auth token is valid** — that was the one thing tests could not prove.

## The bug

```
> Uploaded 474 missing debug information files
```

Per leg. ~1900 per release. What actually went up:

```
release/build/aws-lc-sys-.../out/chacha20_poly1305_x86_64.o
release/build/proc-macro2-.../build-script-build
release/deps/libscroll_derive-....so
AI Job Hunter.AppDir/usr/lib/libicudata.so.70
AI Job Hunter.AppDir/usr/lib/libudev.so.1
AI Job Hunter.AppDir/usr/bin/ajh-tauri          ← the only one that matters
```

Dependency build scripts, proc-macro `.so`s, `ring`/`aws-lc-sys` object files, and the third-party system libraries bundled into the AppImage. I had pointed `sentry-cli debug-files upload` at the whole `target/` tree.

## The fix, and why it is derived rather than guessed

Scoped to `ajh_tauri*` under `release/deps`. I took the paths **from that run's logs instead of inferring them**, which turned out to matter — my first attempt at this patch guessed `target/<triple>/release/ajh-tauri.dSYM` and would have been wrong on two platforms:

| Platform | Actual |
| --- | --- |
| Linux | `release/deps/ajh_tauri-<hash>` |
| Windows | `release\deps\ajh_tauri.exe` + **`ajh_tauri.pdb`** |
| macOS | `<triple>/release/deps/ajh_tauri-<hash>` + `.dSYM` |

On every platform the artifacts carry the **library** name (`ajh_tauri`, underscore) and sit in `deps/` — not the bin name at `release/ajh-tauri`. The guessed path would have uploaded the macOS binary **with no debug companion**: symbolication silently degraded, which is this feature's signature failure.

The copies under `release/`, `bundle/appimage/` and `bundle/deb/` are hardlinks that dedupe to the same debug id, so nothing is lost by ignoring them.

Also guarded the per-triple directory behind a non-empty check — with `matrix.target` empty the path collapses to `target//release/deps`, the same directory, enumerating everything twice.

## Verification

Simulated all three real layouts plus the empty case:

```
linux   -> 2: deps/ajh_tauri-fa4fa53a  deps/ajh_tauri-fa4fa53a.d
windows -> 2: deps/ajh_tauri.exe       deps/ajh_tauri.pdb
macos   -> 2: <triple>/deps/ajh_tauri-d8c3e74  <triple>/deps/ajh_tauri-d8c3e74.dSYM
empty   -> 0  (fails loudly, per the existing guard)
```

474 → 2 per leg, debug companion included in every case. YAML parses, step body passes `bash -n`, `continue-on-error` untouched (see ADR-0020 for why it must stay).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -517,6 +517,25 @@ jobs:
             echo "::error title=No debug files found::No ajh_tauri* artifacts under $BASE/**/release/deps on ${{ matrix.platform }} — nothing was uploaded, so crash reports from this release will show unresolved addresses."
             exit 1
           fi
+
+          # "Non-empty" is not the same as "symbolicatable". Finding
+          # `ajh_tauri.exe` while its `.pdb` is missing would pass the check
+          # above and still ship unresolvable stack traces — this feature's
+          # recurring failure shape. So assert the DEBUG COMPANION each platform
+          # actually produces, as observed in a real release run:
+          #   windows  a separate .pdb
+          #   macos    a .dSYM bundle
+          #   linux    none — DWARF stays inside the binary, nothing to require
+          case "${{ runner.os }}" in
+            Windows) NEED='\.pdb$';  NEED_DESC='a .pdb debug companion' ;;
+            macOS)   NEED='\.dSYM$'; NEED_DESC='a .dSYM debug companion' ;;
+            *)       NEED='';        NEED_DESC='' ;;
+          esac
+          if [ -n "$NEED" ] && ! printf '%s\n' "${DIFS[@]}" | grep -qE "$NEED"; then
+            echo "::error title=Debug companion missing::Found ajh_tauri artifacts on ${{ matrix.platform }} but not $NEED_DESC, so frames from this release will not resolve. Check the release profile's debug/split-debuginfo settings."
+            exit 1
+          fi
+
           printf 'Uploading debug files:\n'
           printf '  %s\n' "${DIFS[@]}"
           npx --yes @sentry/cli@^2 debug-files upload "${DIFS[@]}"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -480,8 +480,46 @@ jobs:
           # `split-debuginfo = "packed"` puts them beside the binary in target/.
           # No `--include-sources`: that would ship our source into Sentry, and
           # line tables alone are enough to resolve a frame.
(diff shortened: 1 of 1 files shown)


saeedkolivand:
Addressed in `8cd43632`.

**🟡 Medium — the empty-check isn't a completeness signal.** Correct, and worth doing now rather than deferring: `DIFS` being non-empty only proves *an* `ajh_tauri*` file exists. Finding `ajh_tauri.exe` while its `.pdb` is missing would pass and still ship unresolvable stack traces — the same looks-fine-reports-nothing shape this feature has produced at every stage.

Added a per-platform assertion on the **debug companion**, using the companions each platform actually emitted in run [30741386889](https://github.com/saeedkolivand/ai-job-hunter-app/actions/runs/30741386889) rather than assumed ones:

| Platform | Required |
| --- | --- |
| Windows | a separate `.pdb` |
| macOS | a `.dSYM` bundle |
| Linux | none — DWARF stays inside the binary, so requiring a companion would false-fail |

Tested both directions per platform:

```
windows-good : PASS (2 files)
windows-nopdb: FAIL (missing a .pdb debug companion)
macos-good   : PASS (2 files)
macos-nodsym : FAIL (missing a .dSYM debug companion)
linux-good   : PASS (2 files)
```

**Not doing** the second half of the suggestion — cross-checking that the uploaded debug ID matches the shipped binary's build-id/PDB GUID. That needs parsing sentry-cli's accepted-ID output and re-deriving the shipped artifact's ID, which is a meaningful chunk of workflow logic to catch a case that the companion assertion already covers in practice (a mismatched ID would mean cargo emitted a companion for a different build, which the same `deps/` directory makes near-impossible). Recording it as a deliberate skip rather than an oversight.

The honest limit stays what it was: the definitive check is a real crash resolving to named frames in Sentry. This assertion narrows what can silently go wrong on the way there; it doesn't replace it.

saeedkolivand:

PR #931, opened 2026-08-04
fix: keep project links out of the resume header and make them work

saeedkolivand:
Fixes three defects reported from one generated résumé.

## Project URLs rendered on the contact line

Project sites were swept into the contact profile's extra links at import, so the exported header rendered them on the contact line **and** again in PROJECTS. Only platform profiles may seed extra links now, and website selection prefers an apex over its own subdomains, so it no longer depends on the order links happen to appear in the source PDF.

## PROJECTS read as `Name — same-name-slug`, with no links

The prompt asked for that shape, and link injection matched the machine label literally while the PDF extractor humanised it differently — so the label never matched and nothing linked. Items now carry their real name; matching is normalised and assigns each link to its best line; job-board and non-personal profile URLs are dropped rather than becoming invented projects; and a link that still matches nothing is placed in the localized projects section or left unplaced, never appended as an orphan.

## The header in the editor differed from the exported one

Edits to it were silently discarded from PDF and DOCX while still shipping in TXT and clipboard copy. The document text now owns the header at export; the profile seeds it at generation and fills only what the text leaves blank. See [ADR-0021](docs/adr/0021-editor-owns-resume-header.md).

## Review notes

This went through three security rounds and a three-pass review ensemble before opening. Findings worth calling out, because each was invisible in the diff and only appeared when a reviewer executed the code:

- The header-seeding scan could **delete résumé content** — an entire job, a skills line, a project link — for four of the seven locales the app generates for, and for English résumés whose first heading wasn't in the recognised list. Seeding now only ever replaces one line or inserts; removal is structurally impossible.
- The Resume Builder path was never seeded, so making export fallback-only would have shipped the model's invented contact line. Both generation paths now seed through one helper, and the prompt's false promise that the line "will be replaced on export" is gone.
- A profile with only a name never seeded it, and export could not rescue it.
- The two profile lookups were unguarded and could discard a completed, already-paid-for generation.

Rules that exist in both Rust and TypeScript are now gated by shared JSON fixtures asserted from both sides, and the section-name list is read from its fixture by TypeScript rather than duplicated — after three separate bugs this session traced to a mirrored rule whose comment claimed a parity it did not have.

## Verification

`cargo test --lib` 2915 · `cargo clippy --all-features -D warnings` clean · `cargo fmt` clean · `pnpm typecheck` 13/13 · `pnpm lint:strict` clean · `pnpm gen:ipc:check` up to date · `pnpm -C packages/prompts test` 597 · `pnpm -C apps/desktop test` 2772

## Summary by CodeRabbit

* **New Features**
  * Résumé generation now seeds headers with saved contact details, including localized contact lines.
  * Improved link handling selects relevant personal websites and profile links while filtering job boards and unsafe URLs.
  * Added recognition for résumé sections and contact information across more languages.

* **Bug Fixes**
  * Preserves edited résumé header names and contact details during DOCX and PDF export.
  * Prevents malformed links, duplicate contact content, and incorrect header validation results.
  * Contact profile save failures now display a localized notification.


saeedkolivand:

PR #932, opened 2026-08-04
fix: warn when local ollama has no account key for company research

saeedkolivand:
Company research on local Ollama needs a free Ollama **account key** (credential slot `ollama-cloud`), because Ollama has no model-side web search — the app calls its web search API instead. Without the key the backend returns nothing and the renderer swallows the failure, so a missing key looked exactly like *"found nothing"*, and the cover letter was generated ungrounded with no signal to the user.

## What changed

The interview-questions modal already warned about this, and the AI-Generate fine-tune step carried its own guarded copy of the same rule. Both now consume one shared hook, and the hint appears beside the "research company" toggle in the tailor flow too. No new strings — the existing translation is reused.

## Review notes

The pre-PR review caught that the repo had **two** implementations of this predicate, and the first attempt at consolidation extracted the wrong one — dropping a cold-boot guard added by #682. Both defaults fail *into* the hint (`activeProvider` defaults to `ollama` before the config resolves, `has` defaults to `false` before the key query resolves), so an unguarded check tells a user who **already has** the key that they need to add it, on first paint of a cold session. Reproduced by executing the hook against a stub client.

The shared hook now guards both pending windows — including the key-query one, which the fine-tune copy never covered — and the fine-tune step was migrated, so there is genuinely one predicate in one place rather than two.

The hint is announced (`role="status"`, `aria-live="polite"`): the provider picker sits directly above it, so switching providers makes it appear in response to the user's own action, and `TailorWizard` has no ambient live region the way its sibling wizards do.

Tests were tightened alongside: the a11y attributes are now asserted by role rather than text (stripping them fails the suite), and the credential-slot argument is asserted rather than discarded by the mock — previously, changing `ollama-cloud` to `ollama` would have kept every test green while pinning the hint on for every Ollama user in production.

## Verification

`pnpm typecheck` 13/13 · `pnpm lint:strict` clean · `pnpm -C apps/desktop test` 240 files / 2792 tests

## Summary by CodeRabbit

- **New Features**
  - Added clearer guidance when Ollama-based research features require a missing research key.
  - Added accessible status messaging without blocking model selection, research toggles, or AI generation.
  - Suppressed key warnings while provider and key information is still loading.

- **Bug Fixes**
  - Improved consistency of research-key detection across AI generation, interviews, and document tailoring flows.

- **Tests**
  - Added coverage for provider selection, loading states, missing keys, and interaction behavior.


saeedkolivand:
@coderabbitai

saeedkolivand:

PR #933, opened 2026-08-04
feat: offer reasoning effort per model and fix embedding re-indexing

saeedkolivand:
Two AI-provider fixes that share the adapter files, so they ship together.

## Re-indexing reported success while writing nothing

Input was truncated to a fixed character budget that overflows a dense language's token window, every provider error was swallowed, and the batch completed regardless — so the status strip correctly read `0/1 indexed` underneath two success toasts.

Input is now chunked and mean-pooled so the whole document is embedded, the working chunk size is learned once per document and bounded, and a run that embeds nothing fails instead of reporting success. Gemini's default embedding model had been **retired by Google in January**; changing the code default fixed nothing because the resolved name is persisted, so a migration rewrites it and evicts the caches keyed on it.

## Reasoning effort reached only the Codex agent

Every HTTP adapter dropped the setting. It is now gated on the per-model reasoning capability, and the levels a picker offers come from the backend **per model** — providers publish different vocabularies per model, and an unsupported value is rejected outright.

## ⚠️ One-time effect on first launch

Stored vectors carry a format version, and this bumps it. Vectors written before this change were truncated-prefix embeddings and are **not comparable** with pooled ones, so every stored vector reads stale and lazily re-embeds. Nothing unrecoverable is deleted — only regenerable caches (`posting_vectors`, `match_scores`). This is intended and one-way.

## Review notes

A security gate plus a three-pass review ensemble ran before this opened. Findings worth calling out, because each needed the code executed rather than read:

- **The learned chunk size was poisoned by each chunk's leftover tail**, so one short remainder shrank the working size for the rest of the document. A 300,000-character résumé burned its entire call budget and failed after embedding 3% of its text. Fixed; the same document now completes in 97 calls.
- **Gemini's per-model level table was built, tested and advertised to the UI, but never consulted on the send path.** Since the setting is stored per *provider*, switching Gemini models carried a level the new model rejects straight to the API.
- **The spend ledger recorded nothing when a multi-chunk embed failed partway**, discarding calls already billed — a regression introduced by moving the paid call inside a retry loop.
- **This branch had added a model to the one-click list that Google has since shut down.** All three stale-model defects here came from hardcoded lists that were correct when written, so each such list now records the vendor page it was checked against and the date.

The adaptive embedding path moved to its own module: adding it had nearly doubled the provider module, tripping the architecture LOC rule. Split rather than allowlisted — the rule exists to report exactly that.

## Verification

`cargo test --lib` 3010 · `cargo test --test architecture` 11/11 · `cargo clippy --all-features -D warnings` clean · `cargo fmt` clean · `pnpm typecheck` 13/13 · `pnpm lint:strict` clean · `pnpm gen:ipc:check` up to date · `pnpm test` 413 files / 4747 tests

## Summary by CodeRabbit

* **New Features**
  * Added model-aware reasoning controls that show supported effort levels for compatible AI providers and models.
  * Added model capability reporting for reasoning and web search.
  * Improved embedding for long documents with adaptive chunking, retries, pooling, and normalization.
  * Updated Gemini defaults to newer embedding and generation models.
* **Bug Fixes**
  * Embedding errors now provide clear failure details instead of silent null results.
  * Re-indexing now reports success, partial completion, cancellation, and failure outcomes accurately.
  * Resume matching falls back to keyword scoring when embeddings are unavailable.
* **Documentation**
  * Expanded API documentation for AI capabilities, embeddings, and contact profiles.


saeedkolivand:

PR #935, opened 2026-08-05
feat: make provider model listing report failures and carry model metadata

saeedkolivand:
PR 1 of 3 in a programme to stop hardcoding provider model lists. **No user-visible behaviour change** — this is the enabling change.

## Why

`provider-meta.ts` holds hand-curated model-id arrays. They rot silently: **four stale-model defects shipped in one session**, including a shut-down model serving as the Gemini onboarding default. Live loading already exists (`AiProvider::list_models`, all providers) and already wins at runtime — `build-options.ts` is `names = live.length > 0 ? live : m.models`, so the curated array is only a fallback.

PR 2 deletes those cloud arrays. That cannot land first, because `list_models` collapsed five distinct outcomes into `vec![]` — missing key, request failure, non-2xx, unparseable body, genuinely-empty catalogue — with `ai_list_provider_models` adding a sixth via `resolve_by_name`. Remove the fallback with that in place and every offline or erroring user gets an empty model picker and no explanation.

So the trait now returns `AppResult<Vec<Value>>`, adapters return `Err` for real failures and `Ok(vec![])` only for an empty catalogue, and the command **rejects the invoke** rather than returning an in-band error object no caller reads.

## What review found that the spec didn't

Each of these was live, and each was invisible only because the curated fallback was covering it:

| Found by | Defect |
|---|---|
| correctness | Gemini listed from `/v1` while every generation path uses `/v1beta`, so **every `-preview` model was missing** — including `gemini-3.1-pro-preview`, the curated Pro-tier entry that would have silently vanished in PR 2 |
| correctness | Both cloud list endpoints paginate; only page one was read |
| correctness | A keyless OpenAI-compatible server (LM Studio, vLLM) could generate but never list |
| security | `format!("{base_url}/models")` sent requests to the **wrong path** and leaked a query-string API key into the error this PR now surfaces |

On the last one: `Url::join` is *not* the fix — a reference with no query clears the base's query, dropping the auth entirely. Resolution goes through `path_segments_mut()`, and since `reqwest::Error`'s `Display` embeds the URL and strips only userinfo, errors additionally pass through `scrub_url_secret`. `scraping/http.rs` already defends this exact class.

## Model metadata

Entries widen from `{name}` to `{name, displayName?, createdAt?, contextLength?}`. `name` is byte-identical — it is what every selection and stored preference keys on.

Every field is optional because **no field is available from every provider**, verified against each vendor's docs rather than assumed:

| | displayName | createdAt | contextLength |
|---|---|---|---|
| Anthropic | `display_name` | `created_at` | `max_input_tokens` |
| Gemini | `displayName` | **none** | `inputTokenLimit` |
| OpenAI | — | `created` (epoch) | — |
| Ollama | — | `modified_at` | — |

Gemini returns no creation date, so newest-first ordering can never be universal — which is why PR 3 keeps the onboarding default a user choice rather than a ranking rule. Absent fields are omitted, never zero-filled: a fabricated timestamp sorts wrongly and silently.

Anthropic's `capabilities` object is deliberately **not** carried — reasoning/effort gates already resolve per model via `ai_model_capabilities`, and a second provider-shaped capability source is the duplicated-rule divergence this programme keeps paying for.

## Deliberately deferred

- `base_url` arrives from the renderer on `ai_list_provider_models`; ADR-0012 says it should resolve server-side from `AiConfigStore`. Pre-existing, and fixing it properly changes several command signatures and their IPC contracts.
- Unbounded `resp.json::<Value>()` body reads — fleet-wide, belongs in `net/http.rs`.

## Verification

`cargo clippy --all-targets --all-features --locked -D warnings` clean · `cargo test --lib` 3082 passed · `cargo test --test architecture` 11/11 · `pnpm gen:ipc:check` no drift · `typecheck` 13/13 · `lint:strict` clean · desktop suite 2811 passed.

Security review: **no HIGH/CRITICAL**. Correctness review: **no HIGH/CRITICAL**.

Behaviour-neutrality is covered by a new test that drives the real `ModelSelector` with a rejected query and asserts the curated fallback still renders — previously the component's `useQueries` stub made that path untestable.

## Summary by CodeRabbit

* **New Features**
  * Provider model lists now include display names, creation dates, and context-length details where available.
  * Model discovery supports complete paginated catalogs across supported providers, including compatible and keyless configurations.
* **Bug Fixes**
  * Model-loading failures now show meaningful errors instead of appearing as empty lists.
  * Improved handling of malformed responses, invalid credentials, network issues, and stalled pagination.
  * Sensitive URL and credential details are protected in error messages.
  * Failed cloud-provider model requests no longer retry repeatedly.

### DIFF
--- a/apps/desktop/src-tauri/src/commands/ai.rs
+++ b/apps/desktop/src-tauri/src/commands/ai.rs
@@ -5,6 +5,7 @@ use tauri::{AppHandle, Manager};
 use crate::credentials::CredentialStore;
 use crate::db::new_job_id;
 use crate::documents::{embedding_space_changed, DocumentStore, EmbeddingConfig};
+use crate::error::AppResult;
 use crate::events::{emit_event, JobEvent, JOBS_EVENT};
 use crate::ipc_contracts::ai::AiEmbedRequest;
 use crate::jobs::{JobStatus, JobTracker};
@@ -143,12 +144,9 @@ pub async fn ai_list_provider_models(
     app: AppHandle,
     provider: String,
     base_url: Option<String>,
-) -> Value {
-    let provider_client = match resolve_by_name(&provider, base_url) {
-        Ok(p) => p,
-        Err(_) => return json!([]),
-    };
-    json!(provider_client.list_models(&app).await)
+) -> AppResult<Value> {
+    let provider_client = resolve_by_name(&provider, base_url)?;
+    Ok(json!(provider_client.list_models(&app).await?))
 }
 
 /// Static, network-free capability probe for a provider/model — whether it can
--- a/apps/desktop/src-tauri/src/commands/ai_provider/anthropic.rs
+++ b/apps/desktop/src-tauri/src/commands/ai_provider/anthropic.rs
@@ -26,14 +26,22 @@ const VERSION: &str = "2023-06-01";
 /// `has_more` forever.
 const MAX_LIST_MODELS_PAGES: usize = 50;
 
-/// Resolve the stored Anthropic key, erroring on a missing/blank one. Shared
-/// by `list_models` and `test_key` so the two structurally agree on what
-/// counts as "no key" (previously `test_key` alone accepted a whitespace-only
-/// key and burned a 401 round-trip on it). Pure (no `AppHandle`) so it's
-/// unit-testable without a mock-app harness.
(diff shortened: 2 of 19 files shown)


saeedkolivand:

PR #936, opened 2026-08-05
feat: load cloud model lists from the provider instead of a curated array

saeedkolivand:
The payoff of the programme #935 enabled. **Deletes the hand-maintained cloud model arrays and the hardcoded onboarding defaults.**

Four stale-model defects shipped in a single session from those two places — including `gemini-2.0-flash` serving as the Gemini onboarding default after Google shut it down.

## Why both, in one PR

I originally planned these as two PRs. They aren't two concerns. They're one — **hardcoded cloud model ids that rot silently** — and splitting them ships an intermediate state that is *worse than either endpoint*: `provider-meta.test.ts` asserted that `CLOUD_DEFAULT_MODELS` pointed at a model in the curated list, so deleting the arrays alone removes that guard while leaving the hardcoded default in place, unguarded. That is precisely the hole the original bug came through.

## What replaces them

- **Last-good cache** per provider + base URL, served when a fetch fails and **disclosed as cached** rather than presented as fresh.
- **Three states that no longer look alike**: no key stored (neutral, not an error), cached list served, and a real failure showing the provider's own classified message. #935 produced those messages — 401/403 → "invalid or unauthorized API key", 429/5xx → rate limit / service error — and *nothing rendered them*. Now something does.
- **Onboarding defers the model choice** until a key exists, then the user picks from the live list. Continue blocks until they do; Skip stays always available.
- **No auto-ranked default.** Gemini's list endpoint returns no creation date at all, so a newest-first rule would silently fail for the one provider whose default actually rotted.
- Labels prefer `displayName` (`Gemini 3.6 Flash`, not `gemini-3.6-flash`); sort is stable, so a provider supplying no dates keeps its own order.
- **Keyless OpenAI-compatible can list models.** #935 allowed this in the backend for LM Studio / vLLM; the renderer gate still blocked it, so that fix was unreachable.

CLI-agent aliases (`sonnet`, `gpt-5-codex`, …) **stay hardcoded** — a local CLI binary exposes no catalogue endpoint, so they are not the same rot.

## Two regressions the deleted fallback had been hiding

**`isModelList([])` is `true`** — `every` on an empty array is vacuously true. So an empty cache entry read back as `[]`, and `if (cached)` accepted it, meaning a revoked key surfaced as *"this provider's catalogue is empty"* instead of an auth error. An operation that failed, reported as success — in the code written to make failures visible. Fixed read-side rather than write-side, so entries already written by a pre-fix build are also handled.

**An empty dropdown during load.** With the curated array gone, `options.length === 0` became reachable mid-fetch for the first time, and the branch order rendered a bare dropdown with no spinner. The test that should have caught it asserted only that the *wrong* state was absent, never that the *right* one was present — a lesson both reviewers proposed independently.

Also fixed: **Enter never configured the provider.** The onboarding wrapper's key handler took the raw `onNext`, bypassing the branch that persists the choice. Harmless when a default was auto-applied; it now discards a model the user deliberately picked. Repaired for all three modes (cloud, CLI, local Ollama).

## Accessibility

`ErrorState` and `EmptyState` in `packages/ui` had no `role`/`aria-live` — a shared-primitive fix benefiting every consumer. The new status messages are announced; focus moves to the model picker when it appears after a key save; the newly-conditional Continue explains why it's disabled instead of just dimming.

## Known, filed separately

`WelcomeStep` has the same raw-`onNext` wiring, reachable by an ordinary flow (type a name → Tab to a language tile → Enter) and it silently discards the name. Out of scope here; queued as its own fix.

## Verification

`pnpm typecheck` clean · `pnpm lint:strict` clean · desktop 246 files / **2851 tests** · `@ajh/ui` 60 files / **530 tests**.

Reviewed by `frontend-reviewer` (2 HIGH, both fixed) and `ui-ux-expert` (4 a11y HIGH, all fixed).

## Summary by CodeRabbit

* **New Features**
  * Cloud provider model lists are now fetched live and sorted by newest available models.
  * Added cached model-list fallback when live retrieval fails.
  * Onboarding and settings now require selecting an available cloud model.
  * Added clear loading, error, empty-list, cached-results, and API-key guidance states.
  * Keyless compatible providers can discover and select models.
* **Accessibility**
  * Improved screen-reader announcements for status, errors, and onboarding guidance.
* **Translations**
  * Added German and English text for model availability and retrieval states.


saeedkolivand:

PR #937, opened 2026-08-05
fix: stop probing api.openai.com for users who never configured it

saeedkolivand:
**Privacy regression introduced by #936 and currently on `main`.** Found by the post-merge review on #936.

## What happens today

#936 added a keyless carve-out so LM Studio / vLLM users could list models without storing a key:

```ts
const canFetchModels = (p: AiProvider): boolean =>
  (connected.get(p) ?? false) || p === 'openai-compatible';
```

It requires only that the provider *be* `openai-compatible` — not that it be pointed anywhere.

- `baseUrlFor('openai-compatible')` is `undefined` for anyone who never configured it.
- `OpenAiClient::new` falls back to `DEFAULT_BASE = "https://api.openai.com/v1"` on a blank base URL (`openai.rs:463-465`), sending no `Authorization` header.
- `ModelSelector` mounts on six surfaces: `ResumeBuilderPage`, `TailorFlow/StepModel`, `ReferralModal`, `JobAdView`, `AnalyzeLeftPanel`, `ai-generate/LeftPanel`.

So **every user — including a fully local, offline-by-choice Ollama user who has never touched this provider — makes an outbound HTTPS request to `api.openai.com` on every relevant page load.** For an app whose entire posture is local-first, that is the specific thing it promises not to do.

Before #936 the gate was `connected.get(p) ?? false` for every provider, so nothing fired until the user had stored something.

## Fix

`openai-compatible` fetches only once it has actually been configured — a stored base URL **or** a stored key. Every other provider keeps its stored-key gate unchanged.

Expressed as a single `isProviderConfigured(provider, connected, baseUrl)` predicate rather than a boolean special-cased per site, because the carve-out lived at **three** call sites:

- `ModelSelector/index.tsx` — `canFetchModels`
- `AISettingsTab/useProviderKeys.ts` — `expandedCanFetchModels`
- `CloudProviderConfig/index.tsx` — `canPickModel`

A change reaching some siblings and not all has produced this programme's last four defects, so the predicate exists once and the exhaustiveness was verified by grep rather than assumed.

## Preserved

- A **configured keyless** server (base URL set, no key) still lists models — the case #936 existed to enable. Covered by a test.
- A user with a key but no base URL keeps today's behaviour (resolves to the default base with their key). That predates #936 and is untouched.

## Note on the tests

**Five existing tests asserted the buggy behaviour** — fetching with nothing configured — so they were written to match the bug and would have preserved it. They now cover the configured-keyless case they were meant to protect. That's why this diff touches more test files than source files.

The new regression test asserts `listProviderModels` is never called with no key and no base URL, gated on query settlement rather than a timer (a negative assertion on a `setTimeout` fails in the direction that looks like success).

## Verification

`pnpm typecheck` 13/13 · `pnpm lint:strict` clean · desktop 249 files / **2870 tests**.

### DIFF
--- a/apps/desktop/src/renderer/components/ui/ModelSelector/ModelSelector.privacy.test.tsx
+++ b/apps/desktop/src/renderer/components/ui/ModelSelector/ModelSelector.privacy.test.tsx
@@ -1,0 +1,80 @@
+/**
+ * ModelSelector — keyless-provider privacy regression tests
+ * (fix/no-unconfigured-openai-probe).
+ *
+ * ModelSelector builds a `useQueries` model fetch for EVERY cloud provider on
+ * every mount (`cloudProviders.map`), gated per-provider by `canFetchModels`.
+ * The #936 carve-out — `connected || provider === 'openai-compatible'` — had
+ * no requirement that `openai-compatible` be pointed anywhere, so a user who
+ * never touched that provider (e.g. a fully-local Ollama user) still fired
+ * `listProviderModels` for it on every page mounting this component; the Rust
+ * side falls back to `https://api.openai.com/v1` when no base URL is
+ * configured, sending an outbound request with no `Authorization` header.
+ *
+ * Unlike `ModelSelector.test.tsx` (which stubs `@tanstack/react-query`'s
+ * `useQueries` entirely, bypassing `enabled`), this file mounts a real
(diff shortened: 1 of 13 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #938, opened 2026-08-05
docs: record the no-fallback-model-list decision, and post medium review findings inline

saeedkolivand:
Two commits, separate concerns, both meta-level (no app code).

## 1. `docs:` — ADR-0022 + `docs/API.md`

Closes out the programme that deleted every hardcoded cloud model id (#935, #936). The decision is hard to reverse and genuinely surprising to a future reader — *"why does this app ship no model list at all?"* — so the reasoning belongs somewhere other than a diff.

Records: the four stale-model defects that prompted it; why the onboarding default is a **user choice** rather than auto-ranked (Gemini's catalogue endpoint returns no creation date, so a newest-first rule would silently fail for the one provider whose default actually rotted); why CLI-agent aliases stay hardcoded; and why the scheduled CI staleness check from the original design was **dropped** — it would have false-positived on `RATES` rows that are deliberately retained so historical spend still prices correctly.

Costs are recorded honestly too: onboarding now blocks on a network round-trip that used to be instant, and #937's regression (the keyless carve-out probing `api.openai.com` for unconfigured users) is included as the clearest illustration of what removing a fallback exposes.

`docs/API.md` gains the `listProviderModels` entry it never had, including that it now **rejects** rather than resolving an empty list.

## 2. `ci:` — inline review findings, deduped

The AI review gate could already post inline comments, but only for blocking findings (HIGH/CRITICAL at confidence ≥ 0.8). That tier almost never fires — every finding this gate has produced recently was MEDIUM or LOW, living only in the sticky comment, away from the code it describes. The capability looked absent because it almost never qualified.

**The cap was load-bearing, though.** The action's inline tool does not deduplicate across runs and this lane re-runs on every push, so an unbounded instruction stacked a fresh copy of every finding each time — noise proportional to how long a PR stays open, worst on exactly the PRs getting the most review.

So the model no longer posts at all. It writes findings; a **deterministic step** decides what gets pinned, from the same JSON the verdict already uses — the same split that keeps model prose out of the merge decision.

Dedup reuses the existing `ledgerKey`: file + category + hash of summary, **with no line number**. That omission is the point — an edit above a finding shifts its line, and a line-keyed identity would repost it as new on the next push.

A finding whose line isn't in the diff makes GitHub reject the comment, so those are skipped with a warning rather than failing the step; the sticky comment carries every finding regardless.

Lane A1 (`@claude review`) keeps its inline tool — owner-triggered, runs once, nothing to deduplicate.

### ⚠️ This PR cannot self-test the review

`claude-code-action` refuses to run when a PR edits its own workflow file (the file must match the default branch). So the **AI Review check on this PR will fail open and prove nothing** — that's the documented guard against a PR rewriting its own reviewer, not a failure. Inline posting has to be verified on the next ordinary PR after this merges.

## Verification

`pnpm gen:workflows:check` in sync · `pnpm lint:strict` clean · `scripts/ci-review-verdict.test.mjs` 10 passed (4 existing + 6 new).

The tier rule and the key's line-independence are unit-tested, since neither is observable from workflow YAML.

### DIFF
--- a/.claude/hooks/review-lib.mjs
+++ b/.claude/hooks/review-lib.mjs
@@ -234,6 +234,27 @@ export const blockingFindings = (findings, minConfidence) =>
       f.introduced_by_diff !== false
   );
 
+/**
+ * Findings worth pinning to a line in the PR diff.
+ *
+ * LOW is deliberately excluded: it is advisory by definition, and the sticky
+ * comment already lists every finding, so pinning nits to the diff costs
+ * attention without adding information.
+ *
+ * A finding with no `file`/`line` cannot be anchored at all — GitHub rejects a
+ * review comment without a position, so those stay sticky-only rather than
+ * failing the post.
+ */
+export const inlineFindings = (findings) =>
+  (findings || []).filter(
+    (f) =>
+      (f.severity === 'CRITICAL' || f.severity === 'HIGH' || f.severity === 'MEDIUM') &&
+      f.introduced_by_diff !== false &&
+      f.file &&
+      Number.isInteger(f.line) &&
+      f.line > 0
+  );
+
 export const formatFinding = (f) =>
   `${f.severity} · ${f.file}${f.line ? ':' + f.line : ''} · ${f.summary}${f.fix ? ' · ' + f.fix : ''}`;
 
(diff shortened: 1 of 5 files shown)


saeedkolivand:

PR #939, opened 2026-08-05
fix: keep the typed name when onboarding advances by keyboard

saeedkolivand:
Last of the onboarding steps carrying the wiring defect fixed in `AISelectionStep` (#936).

## The bug

`WelcomeStep` passed the raw `onNext` prop to `OnboardingStepWrapper`. That prop drives exactly one thing — the wrapper's global Enter-key listener — so pressing Enter bypassed `handleNext`, which is what calls `setUserName`.

The name `<Input>` stops propagation on its own Enter handler, so this only fires once focus leaves the field. The reachable path is ordinary:

**type a name → Tab to a language tile → press Enter** → onboarding advances and the name is silently discarded.

## Fix

One line: pass `handleNext` instead of `onNext`, so the keyboard path and the button path are the same path.

## The test

Drives the real sequence — type, `user.tab()` out of the input, Enter — and asserts the name is persisted *and* the step advances.

**It was verified against the unfixed component before being accepted**: reverting the one-line fix makes it fail with `expected '' to be 'Grace Hopper'`. A regression test that has never been seen to fail isn't evidence of anything.

`WelcomeStep` had no prior test file, so the style mirrors `AISelectionStep.test.tsx`'s Enter-path block and uses the real `usePreferencesStore` (reset via `resetPreferences()`) rather than mocking it.

Two supporting cases: click-Continue persists the trimmed name, and Enter with no name typed does nothing (`canAdvance` still gates the keyboard path).

## Scope

The other onboarding steps were audited during #936 — `ResearchStep`, `AdzunaKeyStep`, `ExtensionStep`, `ResumeStep`, `AppearanceStep`, `BrowserStep` and `CrashReportingStep` are each either already wired correctly or have nothing to persist before advancing. Untouched here.

## Verification

`pnpm typecheck` 13/13 · `pnpm lint:strict` clean · desktop 250 files / **2880 tests**.

### DIFF
--- a/apps/desktop/src/renderer/features/onboarding/components/OnboardingStepWrapper/index.tsx
+++ b/apps/desktop/src/renderer/features/onboarding/components/OnboardingStepWrapper/index.tsx
@@ -26,25 +26,32 @@ export function OnboardingStepWrapper({
   showStepDots = true,
   className = '',
 }: OnboardingStepWrapperProps) {
-  // Global keyboard nav for the onboarding flow: Enter advances (unless typing
-  // in an input/textarea), Escape goes back. A window listener keeps the
-  // interaction off the non-interactive step container (jsx-a11y).
+  // Global keyboard nav for the onboarding flow: Enter advances, Escape goes
+  // back — but only when focus isn't on something that already owns Enter
+  // itself. A window listener keeps the interaction off the non-interactive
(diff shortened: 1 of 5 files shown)


saeedkolivand:
@claude review

saeedkolivand:

PR #940, opened 2026-08-05
fix: import linkedin profiles anonymously, because signing in is what breaks it

saeedkolivand:
Reported as *"LinkedIn import on welcome step imports resume, but in resume settings returns an error"* — while signed in to LinkedIn.

## The direction is backwards from how it reads

Both surfaces render the same `ProfileUrlImport` and call the same `profile_import_from_url`. **There is no code-level divergence.** The difference is runtime state, and **being signed in is what breaks it**:

- The importer fetches over plain `reqwest` + `scraper` — **no browser** — and needs the `ld+json` Person block LinkedIn serves only to **anonymous crawlers**.
- It built its client from `<app_data>/browser-state/linkedin/cookies.json` whenever that existed.
- Onboarding has no LinkedIn-connect step, so a fresh install fetched as a guest → public page → parsed fine. Connect LinkedIn, and the same URL returns the authenticated SPA shell → no `ld+json` → failure.

An authenticated fetch never yields a parseable profile, so the credentials bought nothing and broke the only case that worked. The profile-import GET now always fetches anonymously, **deliberately**, with the reason in a comment — "don't send the user's credentials" reads as an oversight otherwise.

Scoped to that GET. Board scraping and login keep their cookie jar; `build_authed_client`'s other caller is untouched.

## It was failing invisibly, twice

**The success toast lied.** The *import* result was error-checked; the *save* was cast to `{ id?: string }` and `notify.success` fired unconditionally. But `documents_import` returns `json!({ "error": ... })` **in-band rather than rejecting** — so a scanned PDF or failed extraction resolved, "Profile imported" appeared, and `onImported` ran with an undefined id, silently skipping the résumé write. **The bug report's own premise may never have been true.**

**The parse reported success on garbage.** It failed only when name AND experience AND skills were all empty, with `name` falling back to `og:title` — so an authwall page with a title produced a "successful" import of an empty résumé. Success now requires a real `ld+json` name, which is precisely the field an authenticated fetch withholds.

## The error advice was pointing at the cause

`isProfileAuthError` matched `login|log in|sign in|…` and replaced the backend message with *"Connect your LinkedIn account in Settings → Accounts."* That now tells the user to do the exact thing that breaks their import — and no backend string matches it any more, so it was dead as well as wrong.

**Deleted outright rather than rewritten.** Keeping a substring classifier over messages the backend owns would rebuild the two-places-encode-one-rule drift this repo keeps paying for. The backend strings were rewritten to be user-readable, so they're shown directly, and a test asserts none of them ever says "log in" again. `resumeInput.profileLoginRequired` removed from `en` + `de`.

## Logging

The path had **zero** `tracing` calls — which is why the reporter's diagnostics bundle contained nothing about their own bug (one app line, plus ~50 `chromiumoxide` warnings that are the *login window*, not the import). It now records platform, HTTP status, whether a session was attached, and the error — but **never the URL or slug**: bundles get shared, and a profile URL identifies a person.

## Found, not fixed (out of scope)

- `board_login/mod.rs` builds cookies with no `Domain=`, so reqwest stores them host-only — a cookie for `www.linkedin.com` isn't sent to `linkedin.com`, and both spellings pass `detect_platform`. Confined to the `scrape_url::try_linkedin` paste flow; the main board scraper bypasses the jar entirely.
- `use-import-with-ocr` checks only for `scanned_pdf`, so other `{ error }` shapes are swallowed with no feedback either way.

## Verification

`cargo clippy --all-features -D warnings` clean · `cargo test --lib` 3107 · `cargo test --test architecture` 11/11 · `pnpm typecheck` 13/13 · `pnpm lint:strict` clean · desktop 250 files / 2882 tests.

New tests cover the authwall page no longer parsing, the failed save no longer reporting success, `scanned_pdf` surfacing its message, and the rate-limited / no-public-data strings reaching the user.

## Summary by CodeRabbit

* **New Features**
  * LinkedIn profile imports now work more reliably with stricter profile detection and clearer handling of supported profile URLs.

* **Bug Fixes**
  * Improved import error messages and removed login-required wording for profile imports.
  * LinkedIn imports now fail gracefully when a profile isn’t publicly accessible, with better handling for rate limits and other fetch issues.
  * Successful imports and save failures now surface more consistently in the resume import flow.

* **Chores**
  * Updated translations to remove an obsolete profile login message.

### DIFF
--- a/apps/desktop/src-tauri/src/commands/profile_import.rs
+++ b/apps/desktop/src-tauri/src/commands/profile_import.rs
@@ -9,6 +9,16 @@ pub async fn profile_import_from_url(_app: AppHandle, url: String) -> Value {
             "name": profile.name,
             "platform": profile.platform,
         }),
-        Err(e) => json!({ "error": e }),
+        Err(e) => {
+            // Host only — never the profile path/slug (PII; diagnostics bundles
+            // get shared with support). The deeper fetch/parse layers already log
+            // HTTP status and session state for the LinkedIn path.
+            let host = reqwest::Url::parse(&url)
+                .ok()
+                .and_then(|u| u.host_str().map(str::to_string))
+                .unwrap_or_default();
+            log::warn!("[profile_import] import failed: host={host} error={e}");
+            json!({ "error": e })
+        }
     }
 }
--- a/apps/desktop/src-tauri/src/profile_import/linkedin.rs
+++ b/apps/desktop/src-tauri/src/profile_import/linkedin.rs
@@ -11,15 +11,36 @@ use crate::error::{AppError, AppResult};
 /// as a fallback for experience, education, and skills.
 pub async fn import(url: &str) -> AppResult<ProfileData> {
     let html = fetch_page(url).await?;
-    let document = Html::parse_document(&html);
+    parse_profile(&html)
+}
 
-    // Try JSON-LD first for name/headline/summary
+/// Parse a fetched LinkedIn page into [`ProfileData`]. Pure (no network), so
+/// the parse-strictness gate below is unit-testable without a fetch mock.
+fn parse_profile(html: &str) -> AppResult<ProfileData> {
+    let document = Html::parse_document(html);
+
     let ld = extract_ld_json(&document);
-    let name = ld
+
+    // The `ld+json` Person block is LinkedIn's own signal that this response is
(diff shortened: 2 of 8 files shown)


saeedkolivand:

PR #934, opened 2026-08-04
fix: make the how-it-works page reachable and add structured data

saeedkolivand:
Diagnosed from the live Search Console for `https://aijobhunter.app/`, not from a checklist.

## The problem

Google indexes **4 of 9 pages**. Absent: `/how-it-works`, `/agent-system`, `/architecture-map`, `/world`, `/accessibility`.

Indexing tracks internal link count almost monotonically:

| Route | Inbound internal links | Indexed |
|---|---|---|
| `/download` | 7 | ✅ |
| `/privacy` | 5 | ✅ |
| `/creature` | 4 | ✅ |
| `/accessibility` | 3 | ❌ |
| `/world`, `/architecture-map`, `/agent-system` | 1 each | ❌ |
| `/how-it-works` | **0** | ❌ |

`/how-it-works` had **no `<a href="/how-it-works">` anywhere on the site**. Every other occurrence of the string is a stylesheet name, a test, a comment, or `<code>` prose. The most commercially useful page was unreachable by crawl, while a doodle short film and the privacy policy were indexed.

Compounding it, the sitemap had **never been read** — Search Console listed zero submitted or discovered sitemaps, even though `robots.ts` correctly declares it and #922 ships the file. Declaring a sitemap in `robots.txt` is evidently not sufficient. It has now been submitted manually, so the orphans will be discovered regardless; internal links still matter for crawl priority and link equity.

For context on why this matters: 3 months of traffic is **84 impressions, 2 clicks, average position 39.1**, and all 12 queries are brand-navigational. Nothing in the index answers a question anyone searches.

## Changes

- `/how-it-works` now appears in the shared footer, the homepage's own footer row, and one descriptive in-content hero link. `/agent-system` and `/architecture-map` gain footer links.
- `SoftwareApplication` JSON-LD on the homepage — the site had **no** structured data. No `aggregateRating`: there are no real reviews, and inventing one is both a Google penalty and a lie.
- Titles carry a searchable phrase; the joke moves to `openGraph.title`, which is what people actually see when a link is shared.
- The six pages that declared no `openGraph.images` now reuse the existing card.

### Titles changed — easy to veto

| Page | Was | Now |
|---|---|---|
| `/` | `AI Job Hunter — please hire him` | `AI Job Hunter — Job Search, Résumé & Cover Letters` |
| `/how-it-works` | `— How It Works (End to End)` | `— How It Works, Step by Step` |
| `/download` | `— Download` | `— Download for macOS, Windows, Linux` |

[...]

PR #941, opened 2026-08-05
feat: add a tech radar that fails the build when an entry goes stale

saeedkolivand:
A public `/tech-radar` page — what this project adopts, trials, assesses and holds, with the reasoning, linked to the ADR where one genuinely covers that choice.

## The interesting problem isn't the drawing

A radar is a **curated judgement artifact** — the same shape as the hardcoded model lists this repo just spent #935/#936 deleting because they rot silently, after four stale-model defects in a single session.

Deriving it from the manifests would remove the judgement that makes a radar worth reading. So entries stay curated, and **staleness is made to fail loudly**: every entry carries a `lastReviewed` date, and `check:tech-radar` fails when an entry names a dependency no longer present in any `package.json` or `Cargo.toml`.

**Note the deliberate asymmetry** with the CI check I argued *against* for model lists two days ago. That one would have fired on `RATES` rows retained on purpose so historical spend still prices correctly. A radar entry naming a removed dependency has no such justification. Same mechanism, opposite verdict, because the underlying data has opposite semantics.

Entries that aren't dependencies at all — techniques, hosted services, and the deliberately-not-adopted — declare a `subjectKind` and are exempt. **A check that cries wolf gets ignored**, which is precisely why the model-list version was dropped.

### Verified in both directions

```
✓ check:tech-radar OK — 40 entries checked against 179 known dependency names
✗ "Tauri 2" names dependency "zzz-not-a-real-package-tauri" — not found in any
  package.json / Cargo.toml on disk. Update the entry's dependencyName…, or move
  it to ring: 'hold' with subjectKind: 'not-adopted'…, or delete the entry entirely.
```
Exit code 1 on failure, 0 restored. A staleness check nobody has seen fail is not known to work.

## Rendering — no new dependency

`apps/landing` has exactly three runtime deps (`next`, `react`, `react-dom`) and is a static export. Zalando's radar needs d3; Thoughtworks' BYOR is canvas plus a Google Sheet. A radar is polar-to-cartesian arcs and dots, so it's **hand-written inline SVG** — d3 would have cost more than it saved.

## Accessibility

An SVG of dots is invisible to a screen reader, so the full data is **always present as a text list** — the SVG is decoration over it, not the only copy. Rings are encoded by **shape as well as colour**, every blip is a focusable anchor, and the SVG is hidden below 560px in favour of the list that was already there.

The a11y gate caught a real WCAG AA contrast failure during development (`--doc-accent` at 4.46:1) plus a heading-hierarchy skip; both fixed, and `check:a11y` is now 0 violations across 13 routes.

## Content

40 entries across four quadrants named for **this project's real shape** — Renderer & UI, Rust Core & Data, Documents & Export, Build/Ship/Trust — rather than Thoughtworks' generic set, which would have left two of four nearly empty. Verified name-by-name against the manifests. Six link a real ADR.

The **Hold** entries are real, sourced decisions (XState, a dedicated vector DB, Nominatim, Codecov + SonarCloud, a WebGL-driven landing) — that's the part of any radar that reads as credible rather than promotional, and it's a public statement about the stack, so it's worth reading closely.

## Not orphaned

In `sitemap.ts`, linked from `SiteFooter` and `DocShell`, registered in `check-parity.mjs`. #934 exists because `/how-it-works` had zero inbound links and was never indexed — a brand-new route is exactly where that repeats.

## Verification

`check:tech-radar` both directions · landing build emits the route · `check:parity` OK · landing 36 files / 331 tests · `check:a11y` 0 violations / 13 routes · `pnpm typecheck` · `pnpm lint:strict` · 7 unit tests on the check itself.


saeedkolivand:
@claude review

saeedkolivand:

PR #946, opened 2026-08-05
fix: pay off three deferred correctness defects

saeedkolivand:
Three independent fixes, each deferred from an earlier PR's review. Every one produces a **wrong answer without producing an error**, so none would have surfaced from ordinary use.

## 1. Cached match scores ignored the embedding vector format

`MatchScoreKey` carried provider, model and `formula_version` — not the vector version. Semantic scores are computed *from* embeddings, so a format change left every cached score a valid hit against vectors that no longer mean the same thing.

**It has been safe only by coincidence.** #933's format bump happened to land alongside a `MATCH_FORMULA_VERSION` bump, and *that* field is in the key — so scores were invalidated on the formula axis, by luck rather than by anything reasoning about vectors. The `evict_..._format_v2` migration clears `posting_vectors`, not `match_scores`.

The version is now part of the key **and** the table's primary key, so a format change invalidates by construction. Adding a `DELETE` to each future migration was the alternative, and it depends on whoever writes the next migration remembering — which is exactly what this already survived once on luck.

Raised by CodeRabbit on #933, correctly about the shape and incorrectly that it was live.

> **Found, not fixed:** `posting_vectors` has no persisted version column at all — it synthesises the current one on read, so it can never self-detect staleness and still needs a hand-written eviction per bump. That's a schema change; separate PR.

## 2. Authentication depended on how the URL was spelled

`board_login` rebuilt stored cookies as `name=value; Path=…` with **no `Domain`**, added against the stored domain minus its leading dot. Per RFC 6265 that makes them host-only — discarding the leading dot, which is the very thing marking a cookie as applying to subdomains. A session captured for `.linkedin.com` then never reached `www.linkedin.com`.

Both spellings pass `detect_platform`, so the same account silently produced an authenticated or a guest fetch depending on URL form.

A stored domain with the leading dot now rebuilds with `Domain`; one without stays host-only. **Deliberately narrow** — `Domain` is only ever the cookie's own host. Widening a session credential's scope would send it to a host it was never issued for, which is worse than the bug; where the stored shape is ambiguous it stays host-only.

The safety test (never attached to an unrelated host) was confirmed load-bearing by temporarily broadening the code and watching it fail.

Found during #940's investigation, correctly left out of that PR.

## 3. Gemini's `top_p` escaped the gate `temperature` got

Google deprecates `temperature`, `top_p` and `top_k` **together** for Gemini 3.x — ignored today, documented as a future 400. `temperature` received a per-model gate during the reasoning-effort work; `top_p` was left behind in the same commit, so two halves of one API rule disagreed about the same models.

Both now run through one `gemini_omits_sampling_params` predicate, so a future model can't be gated for one parameter and not the other. **Sharing the rule is the fix, and not sharing it is how the defect was created.**

`top_p` is omitted on v3+ *even when explicit*, deliberately unlike temperature's explicit-always-wins carve-out: temperature is a user dial, so an explicit value is intent worth honouring, whereas `top_p` here is set by our own anti-detection code — no user intent to preserve, nothing worth risking a 400 for. The reasoning is in the doc comments, because a reader would otherwise "correct" the asymmetry.

Verified against Google's current docs rather than inherited from the earlier review. `top_k` is never sent by this adapter, so it needs no gate today.

Raised on #933's review and deferred.

## Verification

`cargo fmt --all --check` clean · `cargo clippy --all-targets --all-features --locked -D warnings` 0 warnings · `cargo test --lib` **3115 passed** · `cargo test --test architecture` 11/11 · `pnpm typecheck` 13/13 · `pnpm lint:strict` clean · `pnpm gen:ipc:check` no drift.

### DIFF
--- a/apps/desktop/src-tauri/src/commands/ai_provider/gemini.rs
+++ b/apps/desktop/src-tauri/src/commands/ai_provider/gemini.rs
@@ -443,27 +443,53 @@ fn parse_gemini_frames(buf: &mut String, state: &mut GeminiScanner) -> Vec<Strea
     out
 }
 
+/// Whether Gemini's deprecated sampling knobs — `temperature`, `topP`, and
+/// `topK` (never wired up in this file — `AiGenerateRequest` has no `top_k`
+/// field, so there is nothing to gate for it) — should be withheld for
+/// `model`. ONE predicate decides all of them so a future model can't end up
+/// gated for one and not another: that exact drift is how `topP` shipped
+/// ungated in the first place while `temperature` already had this check.
+///
+/// Verified against two live Google references (fetched 2026-08-05), which
+/// group all three identically rather than singling out `temperature`:
+/// - `ai.google.dev/gemini-api/docs/whats-new-gemini-3.5`, "Parameter updates
+///   and best practices in Gemini 3.x" — explicitly scoped to "all Gemini
+///   3.x models", not just the newest releases: "`temperature`, `top_p`,
+///   `top_k`: we strongly recommend not changing the default values."
+/// - `ai.google.dev/gemini-api/docs/latest-model#sampling-parameter-deprecation`
+///   (scoped to Gemini 3.6 Flash / 3.5 Flash-Lite "and all future Gemini
+///   model releases"): "`temperature`, `top_p`, and `top_k` are deprecated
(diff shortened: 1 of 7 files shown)


saeedkolivand:

PR #947, opened 2026-08-05
fix: bound every http body read, validate probe base urls, and stop reporting failed imports as success

saeedkolivand:
Four deferred items from the queue, in one branch, plus the findings from a security pass over the batch itself.

## What was wrong

**Unbounded body reads, fleet-wide.** `read_text_capped` existed but lived in `scraping/http`, so every other subsystem read response bodies with no ceiling. Moved to `net/http` (ADR-003's centralized layer) alongside new `read_bytes_capped` / `read_json_capped`, and converted every non-streaming read: geocoding, LinkedIn profile import, the six `scrape_url` resolvers, the LinkedIn client, and 26 calls across the four AI adapters. SSE and the Ollama pull-progress loop are untouched — already incremental. Scraping's own `MAX_BYTES` / `FetchOptions::max_bytes` behaviour is byte-for-byte unchanged.

**Renderer-supplied `base_url` reaching the network unvalidated.** `ai_test_provider_key`, `ai_list_provider_models` and `ai_model_capabilities` handed it straight through, while the setter path has always dropped it for non-`openai-compatible` providers and run it past `ssrf::validate_provider_base_url`. Both rules now live in `resolve_by_name`, the sole entry point for all three. The parameter stays on the IPC surface — Settings deliberately tests an *unsaved* base URL so an LM Studio endpoint can be verified before saving, and local loopback/LAN endpoints keep working.

**`posting_vectors` couldn't detect a stale format.** The read synthesised `version: EMBEDDING_VECTOR_VERSION`, so `EmbeddingConfig::matches` was structurally incapable of rejecting a row. Freshness rested on a hand-maintained invariant: every future format bump had to remember its own eviction migration. Now persisted, same lesson `upsert_vector_with_conn` already carries a few lines below.

**A failed document import reported success.** `documents_import` returns errors in-band, and `use-import-with-ocr` handed every non-`scanned_pdf` error back untouched. Settings and onboarding toasted "Uploaded"; both `ResumeInputCard` paths guarded on `result.id` with no else, so failure produced no feedback at all. All four call sites already wrap `importFile` in a try/catch, so the hook throws — one change, four sites. Same shape as the LinkedIn import bug in #940.

## What the security review changed

`tauri-security-reviewer` approved with no HIGH/CRITICAL, but three MEDIUMs were worth acting on and one falsified a claim in this PR's own commit message:

- **The cap didn't cap.** The LinkedIn client disables reqwest decompression and gunzips by hand, so bounding the read bounded only the *compressed* bytes — 8 MB in, up to ~8 GB resident. Both decode sites now bound the decompressed side too.
- **`resolve_by_name` wasn't the single entry point.** `ai_set_embedding_config` persisted a `base_url` unvalidated, and it reaches egress via `embed_text` carrying the provider key and résumé text.
- **The sweep missed a read** in a file it had already converted — a discarded `.text()` on an error branch.

Plus: `read_json_capped` rebuilt on `from_slice` so it stops charset-sniffing (JSON is UTF-8 per RFC 8259; a gateway mislabelling it as `iso-8859-1` was turning valid bodies into schema errors), its parse-failure log now emits position rather than serde's message, and the `.await??` rewrite that had silently flipped an IPC error code from `PROVIDER` to `PARSE` is restored and pinned by a test.

Two test fixes worth naming: the cap test drove one oversized chunk, so it passed identically against a per-chunk guard — now three small chunks through the shared accumulator, verified by temporarily breaking the guard. And the migration idempotency test rolls `user_version` back, because `run_migrations` skips on `current >= version`, so the guard it claimed to re-exercise never ran twice.

## Also here

The macOS `xattr` command was already on the download page with a copy button, but nothing said the code block goes into the Terminal app.

`ResumeStep` took the new document id from the pre-import cached list rather than the import result, so it could record the wrong default document — or none at all on a first-ever import.

## Migration note

`add_version_to_posting_vectors` is appended at the **end** of the array. Migrations are position-indexed via `PRAGMA user_version`, so inserting mid-array would make every already-migrated install skip it forever. It defaults to `2`, not `0` — every surviving row post-dates the v1→v2 eviction wipe and is provably current, so defaulting to 0 would have re-embedded the whole posting cache for no correctness gain. The literal must not track future bumps.

## Verification

`cargo fmt --all --check` clean · `cargo clippy --all-targets --all-features --locked -- -D warnings` 0 warnings · `cargo test --lib` 3133 passed · `cargo test --test architecture` 11/11 · `pnpm typecheck` 13/13 · `pnpm lint:strict` clean · `pnpm gen:ipc:check` no drift · `pnpm test` 4864 passed.

Not in scope, flagged by the review as advisory: `read_json_capped`'s call sites all deserialize into `Value` today, so the position-only log is precaution rather than a fix.


saeedkolivand:

PR #949, opened 2026-08-06
fix: make clicking an OS notification open the app, on the right screen

saeedkolivand:
Two reports, one cause: clicking a Windows notification did nothing, and there was no way to jump from a notification to the thing it was about.

## Why it never worked

`tauri-plugin-notification`'s desktop `show()` is notify-rust fire-and-forget, and the `onAction` JS helper its docs point at for clicks resolves to `plugin:notification|register_listener` — a command the plugin registers **only on mobile**. Its desktop `invoke_handler` is exactly `notify`, `request_permission`, `is_permission_granted`.

So the renderer's subscription failed with `notification.registerListener not allowed. Command not found` on **every desktop startup**, and no banner click ever reached the app. `notifications_clicked` — which already focuses the window and opens the inbox — had no caller other than the tray. The capability was granted; the command simply doesn't exist off mobile.

That startup error had been happening the whole time. It only became visible after the renderer console bridge landed in #948, which is what surfaced this.

## The fix

Show the banner natively, per platform — only Windows is callback-based:

| Platform | Mechanism | Blocking? |
| --- | --- | --- |
| Windows | `tauri-winrt-notification`'s `on_activated` | No — WinRT event handler |
| macOS / Linux | notify-rust's `wait_for_action` | **Yes** → runs on `spawn_blocking` |

Linux additionally needs a declared `default` action for a body click to be delivered at all. Both crates were already in the tree transitively and are pinned to the locked versions, so there's no duplicate copy.

## Deep-linking

The click now carries the clicked notification's **own** route, so "Autopilot X found 3 jobs" opens *that* autopilot instead of a list to search through.

This needed almost no new machinery: every notification source already attaches a route (autopilot → `/autopilot` with its id in `search`), and the bell's rows already followed it. The OS banner was the one surface that didn't. It goes through the same `resolveNotificationRoute` validation the bell uses, so an unknown backend route falls back to `/` rather than breaking navigation. A tray click carries no route and still just opens the inbox.

The dead `onOsBannerClick` seam is removed — contract, client namespace, mock and hook — which also silences the startup error.

## CI: macOS was compiling nowhere

Worth flagging on its own. Every Rust job runs on `ubuntu-latest`, and `macos-` appeared in no workflow except `release.yml` — so macOS-only code compiled **nowhere until a release was cut**. `Cargo.toml` already carries a comment about being burned by exactly this shape. Added a `cargo check` on `macos-latest` for PRs (check only, no bundling/tests, to stay cheap on the 10x-billed runner).

## Verification — read this before approving

- **Windows** — the failing path was reproduced first: the exact `registerListener` error was observed in the log file during a real `pnpm dev` run, which is what identified the cause. Compile-checked locally.
- **Linux** — compile-checked by CI (`ubuntu-latest`).
- **macOS** — compile-checked by the new job, and **not behaviourally verified** — I have no macOS machine. The API shape was read from the vendored `notify-rust` source rather than assumed, but whether a body click actually yields `NotificationResponse::Click` in our packaging is unconfirmed.
- Route navigation is unit-tested (exact target + search preserved, unknown route falls back, no-route opens the inbox instead), and the `notifications:open` wire shape is pinned against the renderer contract on the Rust side.

Gates: clippy `--all-features` 0 warnings · `cargo test --lib` 3157 · architecture 18/18 · `pnpm test` 4896 · typecheck 13/13 · `lint:strict` clean · no IPC drift.

## Summary by CodeRabbit

* **New Features**
  * Clicking desktop notifications now opens the relevant in-app destination directly, including preserved search context.
  * Notifications without a destination continue opening the inbox.
  * Added clickable notification support across Windows, macOS, and Linux.

* **Bug Fixes**
  * Improved handling of unknown or route-less notification clicks.

* **Tests**
  * Added coverage for routed notifications, inbox fallback behavior, and notification event handling across platforms.


saeedkolivand:
@coderabbitai review

saeedkolivand:
## Review round 1 — all four fixed (`ec30bb7f`)

**🟠 Blocking-pool exhaustion.** Real, and the worst of the four. `wait_for_action` parks its thread until the user clicks or the banner closes — and a notification nobody touches (sitting in macOS Notification Center) parks it *indefinitely*. `OsBanner::Always` sources like autopilot emit bursts, so a run of ignored notifications would hold blocking workers permanently and starve every other `spawn_blocking` caller.

Bounded to 8 concurrent listeners via an RAII slot, released on `Drop` so a panic inside the wait can't leak the budget. Over budget the banner is **still delivered** — verified in the vendored source that Linux's `show()` has already sent it and macOS's handle `Drop` sends it — it just gets no click handler. Losing one banner's click beats starving the pool.

I checked the async alternative before reaching for a bound: `wait_for_action_async` exists **only on Linux**, and behind a feature that swaps the dbus backend — not a uniform fix.

**MEDIUM — Windows compiles nowhere either.** Correct, and it's the sharper version of the gap I opened the macOS job for: no PR job compiles `cfg(windows)`, so the code that actually fixes the reported bug was built only on my machine. Added `cargo-check-windows`.

**🟡 — my comment overstated the gap.** Also correct. `cfg(not(windows))` **is** compiled by the ubuntu jobs; only platform-exclusive predicates (`cfg(target_os = "macos")`, `cfg(windows)`) were uncovered. Comment rewritten to say that precisely.

**🔵 — tests should use the shared type.** Right for the reason given: those tests exist to guard the `notifications:open` contract, so a local structural `Payload` alias couldn't detect drift in it. They now import `NotificationOpen` from `@ajh/shared`.

Gates: clippy `--all-features` 0 warnings · `cargo test --lib` 3157 · architecture 18/18 · `pnpm test` 4896 · `lint:strict` clean · pre-push AI review clean.

⚠️ Unchanged from the description: the macOS path is compile-checked only. There's no Mac available to verify behaviour on, so that stays unproven at merge — a deliberate, accepted call, not an oversight.

saeedkolivand:

PR #950, opened 2026-08-06
fix: map local Ollama's stop reason, and stop presuming an OpenAI embedding model

saeedkolivand:
Two follow-ups deferred from #948. I verified both against the code before touching anything — and the second one turned out **not** to be the defect I'd recorded.

## 1. Local Ollama never reported a stop reason

`parse_ollama_frames` set `usage` on the final sentinel but skipped `done_reason`, which rides on that **same object**. So `stream::finish` couldn't tell *"the model burned its whole output budget reasoning and never answered"* from *"the provider silently returned nothing"* — the exact ambiguity that made the empty generation in the macOS report unexplainable.

The non-streaming agent path (`parse_ollama_turn`) already read the field. Rather than copy the mapping, it's extracted to `ollama_done_reason` and shared — a second hand-written copy is the duplicated-heuristic defect this repo keeps re-learning. Callers keep their own precedence on top (a turn still lets `Length` outrank a tool call, since truncated tool-call JSON is worse than a truncated answer).

Also corrected the `stop_reason` doc in `stream.rs`, which now claims one fewer unmapped provider.

## 2. `text-embedding-3-small` was offered to providers that never heard of it

This is the part that wasn't what I expected. I had this recorded as *"Ollama Cloud over-reports `supports_embeddings`"*. Tracing it, the actual defect is narrower, more concrete, and **wider in blast radius**:

`default_embedding_model()` returned `text-embedding-3-small` — an **OpenAI** model id — from `OpenAiClient`. Every OpenAI-compatible client is built on it:

- **Ollama Cloud** delegates the method straight through
- **`OpenAiCompatible`** covers LM Studio, vLLM, OpenRouter

So embedding with any of them and no explicit model posted an OpenAI model id to a gateway that doesn't host it. The model-not-found came back reading as "embeddings are broken" instead of "choose an embedding model". Now gated on `id == OpenAi`, exactly the way `supports_web_search` already is.

### Why I did *not* flip `supports_embeddings`

Despite that being how I'd recorded the task:

- The boolean is **provider-level**, not per-model — the error text says so, and Anthropic is the only `false`.
- A gateway's `/v1/embeddings` endpoint may genuinely exist, and an explicit model the user knows works must still go through.
- Making it per-model needs a model-name allowlist, which breaks the zero-code-change rule for new models — and Ollama's embedding models (`all-minilm`, `bge-m3`) don't even share a name pattern, so a heuristic would produce **false negatives that block working setups**. A false negative here is strictly worse than the current false positive.

Only the presumed *default* was ever wrong. Fixing that is what actually removes the failure.

## Also

The two embedding errors were byte-identical strings for different problems. Split them: *"No default embedding model for X — choose one in Settings"* is fixable by picking a model; *"X does not support embeddings"* is not.

## Verification

Both new test groups were proven able to fail — I removed the `stop_reason` mapping and watched `parse_ollama_frames_done_carries_the_stop_reason` fail on the assertion, then restored it.

- `done_reason: length` → `Length`; `stop` → `End` (the differential — a normal completion must not claim truncation); absent → `None` rather than an invented clean end.
- `default_embedding_model` is `Some` only for native OpenAI, and `None` for both `OpenAiCompatible` and `OllamaCloud` (the latter pinned specifically because it delegates, so the gate has to hold at the inner client).

Gates: clippy `--all-features` 0 warnings · `cargo test --lib` 3161 · architecture 18/18 · typecheck 13/13 · `lint:strict` clean · no IPC drift.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved embedding setup guidance when no default model is available; users are prompted to select a model in AI Embeddings settings.
  * Corrected handling of completion stop reasons for Ollama, including streamed responses and length limits.
  * Added a clearer truncation message for local Ollama responses.
  * Preserved the default embedding model for native OpenAI while requiring explicit selection for compatible providers.


saeedkolivand:
## AI review finding — fixed (`83793ae3`)

Good catch, and the sharp part is that **this branch invalidated my own reasoning from #948.**

There I deliberately *removed* the "raise max output tokens in Settings" pointer from `EMPTY_ANSWER_LENGTH_MESSAGE`, on the grounds that the only such control (`LocalModelLimits`) renders solely for the **local** `ollama` provider — and local Ollama was the one provider that could never reach the length branch, since it reported no stop reason at all. Sending an OpenAI user to a field they can't see is worse than no advice.

Mapping `done_reason` in *this* PR is precisely what makes local Ollama reach that branch. So the doc comment became false, and the only users who can actually act on the advice were the ones no longer getting it.

Split the message rather than rewording the shared one:

- **local `ollama`** → names the real control: *"Raise \"Max output tokens\" under Generation limits in Settings → AI"*, matching the UI's own label (asserted in the test, so a UI rename can't silently orphan it).
- **everyone else** → unchanged generic wording.

Pinned with the differential — `ollama-cloud`, `openai`, `openai-compatible`, `anthropic` and `gemini` must all still get the generic message. Without that arm the test would pass on a version that pointed *everyone* at the local control.

Also dropped the CLI-agent call sites' use of this helper: `emit_done`'s own doc says CLI agents have no `finish_reason`, so the call could only ever return the generic constant — it now uses the constant directly instead of passing a provider that doesn't inform the result.

Gates: clippy `--all-features` 0 warnings · `cargo test --lib` 3162 · architecture 18/18 · typecheck 13/13 · `lint:strict` clean.

_(Reposted — the previous comment was mangled by shell expansion before it reached GitHub.)_

saeedkolivand:
@coderabbitai review

saeedkolivand:

PR #951, opened 2026-08-06
feat: index documents automatically, and stop the index strip reading as a chore

saeedkolivand:
Reported as *"after uploading a résumé I have to go to AI settings and press the indexing button"*.

## The premise was wrong, which changed the fix

`match_resume.rs:129-141` already embeds the résumé lazily when it finds no usable vector, and caches the result. **Indexing was never required** — the Settings button is a pre-warm.

And the vector has exactly one consumer: semantic match scoring. I checked every production reader. That matters, because semantic scoring is **off by default**, so for a default-config user an indexed résumé is read by nothing at all.

So what actually cost you something was (a) a slow first match — a synchronous embed inside scoring — and (b) a status strip that read like a chore list.

## 1. Say what's true

The strip said, in amber:

> N documents need re-indexing for the active model.

Nothing *needs* it. Reworded to state that matching indexes on first use, dropped the warning styling (nothing is wrong), and added a distinct line for when auto-indexing is on.

## 2. Auto-index — opt in

New `autoIndexOnUpload` preference, and one hook mounted at the root. All three moments that create stale documents surface identically as a non-zero `documents.stale`, so a single effect covers them:

- a résumé was just imported
- the embedding provider/model changed
- either of those was left over from last session

**Backed by a new command, deliberately not the existing one.** `ai_reembed_all` re-embeds *every* document unconditionally — reusing it would re-bill a cloud embedding provider for already-indexed documents every time one new file is added. `ai_index_stale_documents` embeds only what's missing and returns a null job id when nothing is, so the common case is one cheap query and no work. Both share one job body so they can't drift in error handling, cancellation or progress reporting.

## 3. Ask during setup

New onboarding step immediately before the crash-reporting consent. Both ask "may this run on your behalf", and this one calls a provider that may bill per token, so the two spend/privacy decisions sit together instead of scattered.

**Defaults to OFF**, and the step's switch is seeded from the stored value rather than forced on. Two reasons, both deliberate:

- an existing install must never start calling a paid embedding provider because of an update
- stepping back through the wizard must not silently discard a "no" — my first draft had `stored || true`, which is unconditionally true and would have done exactly that

Leaving it off costs nothing, which is the point of part 1: matching still indexes lazily.

## Tests

The hook's tests are weighted toward **not** running, since that's where the money is:

- preference off + stale documents → never called
- nothing stale → never called
- **no embedding model configured yet** → waits instead of failing once per refetch (this is the normal first-run path — the résumé step is #2, the AI step is #3)
- repeated refetches → exactly one job, not a paid duplicate per refetch
- a failing index → no unhandled rejection, and the raw provider message never reaches the log

Plus the wizard's step-count/navigation tests updated for the new step.

## Not covered

The strip's new wording and the onboarding step have no visual verification here — worth a look in `pnpm dev` before merging, since the whole point of part 1 is how it reads.

Gates: clippy `--all-features` 0 warnings · `cargo test --lib` 3162 · architecture 18/18 · `pnpm test` 4902 · typecheck 13/13 · `lint:strict` clean · no IPC drift.

### DIFF
--- a/apps/desktop/src-tauri/src/commands/ai.rs
+++ b/apps/desktop/src-tauri/src/commands/ai.rs
@@ -819,25 +819,38 @@ fn reembed_run_failed(done: u32, failed: u32) -> bool {
     done == 0 && failed > 0
 }
 
-/// Re-embed every document with the active embedding config, rebuilding the
-/// vector index in the active space. Emits `jobs:event` progress and returns a
-/// job id. Clears the live posting embedding cache so stale-space entries go too.
-#[tauri::command]
-pub async fn ai_reembed_all(app: AppHandle) -> Value {
-    let job_id = new_job_id();
-    crate::commands::jobs::job_start(&app, &job_id, "ai.reembed");
+/// Documents with no usable vector in the ACTIVE embedding space — i.e. never
+/// indexed, or indexed under a different provider/model/format.
+///
+/// The same `EmbeddingConfig::matches` predicate `match_resume` uses to decide
+/// whether it can reuse a stored vector, so "stale" means exactly the same thing
+/// to the indexer and to the consumer.
+fn stale_documents(app: &AppHandle) -> Vec<crate::documents::DocumentRecord> {
+    let store = app.state::<DocumentStore>();
+    let cfg = store.embedding_config();
+    store
+        .list()
+        .into_iter()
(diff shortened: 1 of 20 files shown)


saeedkolivand:

PR #952, opened 2026-08-06
fix: stop squashing the embeddings checkbox, and unnest the provider row button

saeedkolivand:
Three UI defects in Settings → AI. Two reported off a screenshot of the shipped UI, one surfaced by the log bridge, one found while writing the tests.

## The two embeddings toggles looked like different controls

They aren't. **One cause explains both symptoms** ("the checkbox isn't the same" and "the text is too close to it"):

The row is `flex items-center justify-between` with **no gap**, and the `<input>` had **no `shrink-0`**.

- Semantic scoring's one-line description leaves slack in the row → renders correctly.
- The longer auto-index description fills the row → the copy butts against the checkbox, **and** flexbox compresses the input below its own `h-4 w-4`.

So it's not a different checkbox — it's a squashed one. Fixed on **both** rows, so a future copy change to either can't bring it back.

This is precisely the gap I flagged when #951 merged: the copy was never looked at. I reused the semantic-scoring markup and changed the text without checking the layout still held.

## `ProviderRow` nested a `<button>` inside a `<button>`

`ProviderRow` wrapped its header in `<Button variant="unstyled" onClick={onToggleExpand}>` and nested the "test key" `<Button variant="glass">` inside it.

Beyond invalid DOM: **clicking "test key" also fired the row's expand/collapse**, and assistive tech got a nested-control tree it can't express.

React had been reporting this all along — into devtools, where nobody was looking. It only became visible once the renderer console bridge started writing to the log file:

```
[2026-08-06][09:47:56][webview][ERROR] <%s> cannot contain a nested %s. ... button <button>
  <AISettingsTab> → <ProviderRow provider="ollama-cloud"> →
    <_c variant="unstyled" onClick={onToggleExpand}> → <button>
      → <_c variant="glass" onClick={onTestKey}> → <button>
```

Converted to the `role="button"` row pattern the repo already uses — `NotificationBell` carries a comment describing this exact trade-off, so the fix was to follow the existing precedent rather than invent one. Enter/Space activation preserved.

## Found while writing the tests

The semantic-scoring checkbox had **no `aria-label`**. Added.

## On the tests

Asserted on **classes, not geometry**. jsdom reports zero width/height (a documented gotcha in this repo), so a computed-size assertion would have passed happily against the broken markup — it would have been a test that could never fail.

Both layout tests were verified to fail against the shipped version before being kept: reverting `shrink-0` and `gap-3` makes them fail, restoring makes them pass.

Gates: `pnpm test` 4907 · typecheck 13/13 · `lint:strict` clean (incl. the a11y rules, which the `role="button"` conversion has to satisfy).

## Note

The remaining copy from #951 — the stale-documents line — now reads as plain information rather than the old amber warning, but I still haven't seen it rendered. Worth a glance while reviewing this.

### DIFF
--- a/apps/desktop/src/renderer/features/settings/components/ai-settings/EmbeddingsSettings/EmbeddingsSettings.layout.test.tsx
+++ b/apps/desktop/src/renderer/features/settings/components/ai-settings/EmbeddingsSettings/EmbeddingsSettings.layout.test.tsx
@@ -1,0 +1,87 @@
+/**
+ * EmbeddingsSettings — toggle-row layout.
+ *
+ * Both symptoms reported against the shipped UI had ONE cause: the row is
+ * `justify-between` with no gap, and the `<input>` had no `shrink-0`. With the
+ * short semantic-scoring description there is slack and it looks right; with the
+ * longer auto-index description the text block fills the row, so the copy butts
+ * against the checkbox AND flexbox compresses the input below its `h-4 w-4` —
+ * making it look like a different control rather than a squashed one.
+ *
+ * Asserted on the classes because that is where the bug lives: jsdom reports
+ * zero geometry (see the repo's jsdom notes), so a computed-width assertion
+ * would pass against the broken markup.
+ */
+import { describe, expect, it, vi } from 'vitest';
+import { render, screen } from '@testing-library/react';
+
+import type * as AjhUi from '@ajh/ui';
+
+vi.mock('@ajh/translations', () => ({
+  useTranslation: () => ({ t: (k: string) => k }),
+}));
+
+vi.mock('@/services', () => ({
+  useEmbeddingStatus: () => ({ data: undefined, refetch: vi.fn() }),
+  useSetEmbeddingConfig: () => ({ mutateAsync: vi.fn(), isPending: false }),
+  useReembedAll: () => ({ mutateAsync: vi.fn(), isPending: false }),
+  useJobEvents: (_handler: unknown) => undefined,
+}));
+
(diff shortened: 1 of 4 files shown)


saeedkolivand:
## 🟡 fixed (`3f037166`)

Right, and it caught something I'd **claimed in the PR body but hadn't actually done**. I wrote that the old markup meant "clicking test-key also fired the row's expand" and presented unnesting as the fix. Unnesting removed the invalid DOM; the click still bubbles, so the behaviour was untouched.

Enter/Space was worse than described: a focused `<button>` fires a native click too, so the keydown reached the row's handler — whose `preventDefault()` could suppress the test-key button's own activation.

Guarded with `stopPropagation` on both `onClick` and `onKeyDown`, matching the pattern `ApplicationRow` already uses for its note chip and status dropdown (its comment explains the same Enter/Space reasoning).

**The tests now pin behaviour, not markup** — which is the actual lesson here, since a markup assertion would have passed the whole time:

- clicking test-key does **not** expand the row
- the row itself **still** expands — the differential; a guard that swallowed everything would satisfy the first assertion while breaking the row
- the row stays keyboard-activatable

Verified to fail with the guard removed, then restored.

Also corrected my own nesting assertion: it used `getAllByRole('button')`, which matches the `role="button"` row as well, so it flagged the legitimate real-button-inside-role-button-row pattern that `NotificationBell` uses deliberately. It now checks actual `<button>` elements, which is what React's rule is about.

Gates: `pnpm test` 4911 · typecheck 13/13 · `lint:strict` clean · pre-push gate clean.

saeedkolivand:

PR #954, opened 2026-08-06
docs: stop claiming zero telemetry now that crash reporting ships

saeedkolivand:
## Why

We shipped Sentry crash reporting in #927 (ADR-0020: desktop-only, default ON, opt-out) — which reversed a published "no telemetry" promise. The **prose** in `README.md` and `SECURITY.md` was rewritten at the time. The **summary claims** were not, so the repo has been contradicting itself since.

| Where | Said |
| --- | --- |
| `README.md:166` | "local SQLite database; **zero telemetry**" |
| `Features.tsx:175` (landing home) | "OS keychain, local SQLite, zero telemetry." |
| `branding/marketing/build.mjs` | "Zero telemetry" — sub-headline **and** a pill |

The README one is the sharpest: it sits ~70 lines below `README.md:93`, which correctly says the app sends crash reports to Sentry, on by default. One file, two opposite claims.

The marketing slide is worse than the phrase alone — that panel pairs "Zero telemetry" with **"run 100% offline"**, so the two claims reinforce each other.

## What changed

All three now say **"no analytics"** and name the **opt-out crash report**, matching the `/privacy` page and the "Crash reporting" vocabulary entry in `docs/CONTEXT.md` (which reserves "telemetry"/"analytics" for behavioural and usage events — none collected — and keeps them distinct from failure reports).

`branding/marketing/04-private-ai.png` is regenerated, since the committed PNG is the asset that actually gets published. The other three slides re-rendered **byte-identical**, so no font or renderer drift snuck in, and the result was checked visually rather than assumed.

**Deliberately left alone** — the two extension-scoped claims (`apps/extension/README.md:152`, `privacy/sections/Extension.tsx:97`). The extension is excluded from Sentry and declares `data_collection_permissions: { required: ['none'] }` to Firefox AMO, so "no telemetry" is true there and load-bearing for that filed declaration. Also left: "no tracking" in `llms.txt` and `DownloadBody.tsx` — tracking means cross-session behavioural profiling, which genuinely doesn't happen.

## The guard

There was no mechanical guard on this claim, which is why it drifted unnoticed. Adds **check 6** to `scripts/check-landing-drift.mjs`: a denylist for `zero/no telemetry` across `README.md`, `SECURITY.md`, `branding/` and the landing source.

- Scans by **root**, not an explicit file list, so a new marketing surface is covered the day it lands rather than only the three spots we already know about.
- Allowlists the extension privacy section, and the tech-radar Sentry entry whose rationale *narrates* the reversal instead of claiming it.
- Fails loud if a scan root moves, rather than silently voiding itself.
- **Verified non-vacuous:** reintroducing the phrase into `README.md` fails the check.

It also caught my own explanatory comment in `build.mjs` on the first run, which is a decent sign it bites.

## ADR correction

ADR-0020's Consequences section claimed two things that weren't true, both now corrected in place:

1. README, SECURITY.md, ARCHITECTURE.md and CONTEXT.md "were rewritten in the same change" — the rewrite was incomplete, and README's own bullet was one of the misses.
2. `check-parity.mjs` and `check-landing-drift.mjs` "fail the build if code and copy disagree" — neither mentioned Sentry or telemetry. `check-parity.mjs` is a copy-*preservation* gate for the Next port: it asserts phrases survive, it cannot tell a surviving phrase from a true one.

## Verification

`pnpm typecheck` 13/13 · `pnpm lint:strict` clean · `pnpm test` 4912/4912 · prettier clean · pre-push gate clean including `check:parity` (27 signature phrases intact) and `check:tech-radar`.


saeedkolivand:

PR #962, opened 2026-08-09
fix: link orphaned landing routes from home and add sitemap lastmod

saeedkolivand:
## Why

Search Console (Aug 8, 2026) flagged `aijobhunter.app`: 4 indexed / 18 not indexed. Most of the 18 are benign — http→https and www→apex redirects, `?ref=` variants covered by canonical, and Cloudflare's `/cdn-cgi/l/email-protection` 404. The real signal is five pages Google will not serve:

| URL | State |
| --- | --- |
| `/accessibility`, `/agent-system`, `/architecture-map`, `/how-it-works` | Discovered – currently not indexed, **last crawled `N/A`** |
| `/world` | Crawled, not indexed (WebGL page, near-zero text — expected) |
| `/tech-radar` | Not in any Search Console bucket yet |

Nothing was blocking them: all return 200, none carry `noindex`, `robots.txt` is `Allow: /`, and the sitemap reads Success (10 URLs, last read Aug 8).

## What

**1. Home footer had drifted from the doc-page footer.** `SiteFooter` links `/agent-system`, `/architecture-map`, `/tech-radar` and `/storybook/`; the home `Finale` foot-nav did not. Those routes were reachable only from pages Google has not fetched yet, so they sat one hop behind an uncrawled page — and `/` is the only URL on this domain with real crawl equity. Same orphan class #934 fixed for `/how-it-works`.

**2. `HomeBody.test.tsx` asserts the invariant**: every sitemap route (except `/`) is linked from home. Mutation-checked — removing any one link fails it.

**3. Sitemap entries carry `lastmod`**, hand-kept from each route's last content-bearing commit rather than a build timestamp. A fresh date on every deploy is exactly the churn Google learns to distrust; too-old is the safe failure direction.

## Verification

- `pnpm --filter @ajh/landing test` — 338 passed (37 files)
- `pnpm lint:strict` clean · `tsc --noEmit` clean
- Built `out/sitemap.xml` carries `<lastmod>`; built `out/index.html` links all 10 routes
- Pre-push gate: parity OK, agent-system in sync, AI review clean

Indexing is still gated on crawl budget for a young domain — the remaining lever is **Request indexing** per URL in Search Console, which is a manual step on the account.


saeedkolivand:

PR #984, opened 2026-08-12
chore(deps): sentry 0.49 with tauri-plugin-sentry 0.6, gated at the wire

saeedkolivand:
The coordinated Sentry upgrade dependabot #972 could not do alone: `sentry` 0.42 → 0.49.1 and `tauri-plugin-sentry` 0.5 → 0.6 move together (the repo's 0.42 pin existed because the plugin builds against a matched sentry — a solo bump duplicated sentry/sentry-core/sentry-types in the lock and failed 6 checks).

## What's in here

- **The bump**: sentry 0.49.1 + plugin 0.6.0 + sentry-rust-minidump 0.17 — single copy of every sentry crate in the lock, and the `reqwest 0.12` duplicate the old pin dragged along is gone (one fewer HTTP+TLS stack). `ClientOptions` went `non_exhaustive` in 0.49, so `crash_reporting` moved to the builder, field-for-field.
- **A latent no-op removed**: the `rustls-native-certs` reqwest feature we listed was never referenced by reqwest 0.13.1's code and no longer exists in 0.13.2+ — 0.49's resolver made it a hard error. OS trust roots are unaffected (`rustls-platform-verifier` + `native-tls` carry them).
- **The wire gate** (`crash_reporting/transport.rs`, new): security review found plugin 0.6 forwards renderer envelopes that fail to parse as **raw bytes** via `send_envelope`, which never runs `before_send` — an unredacted egress path 0.5 simply dropped. The fix wraps the default (ureq) transport: raw envelopes are dropped and counted, and **consent is enforced per envelope at the wire** — which also closes a pre-existing gap where mid-session opt-out didn't actually stop the plugin's egress (thread-local hub + the plugin's own `Client` clone both bypassed `bind_client(None)`).
- **0.49's new defaults pinned off**: `enable_logs(false)` / `enable_metrics(false)` — upstream flipped both to `true`; consumers are feature-gated today, but feature unification by any future dependency would have turned on new egress invisibly.
- **deny.toml**: `failspot/enabled` (a fault-injection crate arriving via minidump-writer) is now banned, with the ban mechanism mutation-verified live.
- Falsified comments corrected: the module docs now state the transport-level guarantee truthfully instead of claiming `before_send` covers everything.

## Review

Full tauri-security-reviewer pass (BLOCK → fix → delta re-verify → APPROVE). The raw-detection design (`Items::Raw` ⇒ empty `items()` iterator) was verified airtight in 0.49.1 in both directions — including the `envelope_sender::with_item` mutation path between client and transport — and is version-coupled, so it's pinned on the bump checklist in `Cargo.toml` and by a loud test. Known follow-up (recorded, pre-existing, currently PII-free by construction): parsed non-event envelope items (sessions/client reports) pass consent but not redaction; escalates to a blocker only if a browser-SDK integration that emits user-visible non-event items is ever enabled.

## Tests

4004 Rust lib (+7) · 18 architecture · clippy `--all-targets --all-features --locked -D warnings` · fmt · `cargo deny check` (advisories/bans/licenses/sources ok) · `cargo machete` — all green. 12/13 wrapper mutations caught; the survivor (`send_default_pii(false)` deletion equals the default) is disclosed in the test doc rather than overclaimed. macOS compile is CI-verified.

## Summary by CodeRabbit

* **Privacy**
  * Crash reports are now transmitted only when crash-reporting consent is enabled.
  * Consent changes and factory resets take effect immediately.
  * Personal information and server details receive stronger protection during crash reporting.

* **Bug Fixes**
  * Improved handling of malformed or unsupported crash-reporting data to prevent unintended transmission.
  * Enhanced reliability when enabling or disabling crash reporting.

### DIFF
--- a/apps/desktop/src-tauri/Cargo.lock
+++ b/apps/desktop/src-tauri/Cargo.lock
@@ -4978,7 +4978,7 @@ dependencies = [
  "scroll",
  "serde",
  "serde_json",
- "thiserror 2.0.19",
+ "thiserror 2.0.20",
 ]
 
 [[package]]
@@ -4995,7 +4995,7 @@ dependencies = [
  "parking_lot",
  "polling",
  "scroll",
- "thiserror 2.0.19",
+ "thiserror 2.0.20",
  "uds",
 ]
 
@@ -5007,7 +5007,7 @@ checksum = "1dd1d5c43819abcc73ced5a1287c58fdc715c9f993b7cc9debc6ae7cb8421994"
 dependencies = [
  "crash-handler",
  "minidumper",
- "thiserror 2.0.19",
+ "thiserror 2.0.20",
  "uuid",
 ]
 
@@ -8609,7 +8609,7 @@ version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b5520fe8aa00d12fbdb0c290e15f9629f33aaaa1440b1703a01b1cfab44cdf1c"
 dependencies = [
- "base64 0.23.0",
+ "base64 0.23.1",
  "schemars 0.8.22",
  "sentry",
  "sentry-rust-minidump",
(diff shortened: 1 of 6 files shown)


saeedkolivand:

PR #987, opened 2026-08-13
fix(autopilot): stop the double scroll when returning from an application

saeedkolivand:
Fixes the owner-reported jump on the Autopilot page: with several autopilots and the bottom one expanded, applying to a job and then clicking **Back to autopilot** produced two visible motions — the router's global scroll-restoration pass fired first (reset/restore over the whole list), then the page's own focus effect suddenly centered the previously-expanded autopilot.

## What changed

- **One clean scroll.** The application page's `back()` now passes `resetScroll: false` to `navigate()` for the return-to-autopilot trip, so TanStack Router's restore-and-reset pass (verified against the installed `@tanstack/router-core@1.171.21` source — the flag gates the whole block, restoration included) is skipped exactly where the Autopilot page's existing pending-focus effect takes over and performs the single centered scroll it always did.
- **The skip is gated structurally, not by origin alone.** `from=autopilot` is a URL search param that outlives the one-shot in-memory pending-focus state (native forward-nav can replay it), so the gate reads the session store at call time: `resetScroll: !(from === 'autopilot' && hasPendingAutopilotFocus)`. A stale `?from=autopilot` with no pending focus falls back to the router's default reset instead of silently dropping all motion.
- Other back origins (`jobs`, `applications`, none) keep the default reset-to-top — they have no compensating scroll effect.

## Review provenance

frontend-reviewer APPROVE: it read the router package source to confirm the `resetScroll` semantics claim, live-mutated the gate to confirm the seam tests go red, and surfaced the origin-vs-state correlation gap that the second commit closes (plus a comment documenting the trust boundary where backend-supplied search params reach `navigate()`).

## Tests

Seam-level (jsdom zeroes rects, so the tests pin the `navigate()` call shape, not pixels): `from='autopilot'` with pending focus → `resetScroll: false`; without → `true`; `jobs`/no-origin → `true`. Mutation-checked in both rounds — reverting either the flag or the state gate turns exactly the guarding tests red. Touched file 69/69; applications + autopilot suites 497/497; typecheck + eslint clean.

### DIFF
--- a/apps/desktop/src/renderer/features/applications/components/ApplicationDetailPage/ApplicationDetailPage.test.tsx
+++ b/apps/desktop/src/renderer/features/applications/components/ApplicationDetailPage/ApplicationDetailPage.test.tsx
@@ -78,6 +78,9 @@ const mockSessionState: {
     applyForId: string | null;
   };
   setApplicationApply: typeof mockSetApplicationApply;
+  // Only `lastAppliedId` is read by the component (the resetScroll gate); kept
+  // minimal rather than mirroring the full autopilot slice.
+  autopilot: { lastAppliedId: string | null };
 } = {
   applicationApply: {
     applyWizardStep: 0,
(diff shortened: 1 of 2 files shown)


saeedkolivand:

PR #992, opened 2026-08-14
feat(apply): the apply flow runs the staged 4-step pipeline on both surfaces

saeedkolivand:
The apply flow now runs the staged quality pipeline instead of the one-shot fast path, and shows it as four visible steps — **Analyze → Generate → Validate → Remove AI signs** — identically from the Jobs page and from an Autopilot found-job apply. Third of five PRs in the multi-stage pipeline programme (after #990 projects normalization and #991 cover-letter + humanize stages).

## Why the apply flow couldn't run staged before

The run request could only resolve a job ad server-side out of the Jobs-page postings cache, so anything pasted or autopilot-sourced had no `jobId` and fell back to the one-shot path. The staged pipeline was reachable only from a jobs-only side panel.

## What changed

**IPC — a run can now start from text (`packages/shared`, `commands/resume_pipeline/`)**
`resumeId`/`jobId` default to empty; `resumeText`, `jobAdText`, `jobTitle`, `companyName`, `board` are accepted alongside them, with zod refinements requiring one of each pair. Resolution is an **id-wins rule**: a non-empty id means a store lookup and a hard error if it's missing — never a silent fallback to the text. `resolve_resume`/`resolve_job` are pure functions taking an id plus an injected lookup closure, so the text isn't even in scope on that path and a miss can only error. `sourceResumeId` is written only on the id path, which is what keeps #990's projects normalization from running against text of unknown provenance. Job-ad text reaches prompts exclusively through the existing `fenced("job_posting", …)` slot, so it inherits ADR-010 injection protection with no new primitive. `jobAdText` is now persisted into the record's `job_ad` (previously left deliberately empty on the text path).

**Retention** — a URL-less pasted job gets its own history slot via a hashed `unlinked_run_key` used only for `RunRow.job_url`, instead of every pasted job sharing one bucket.

**Résumé provenance (`ResumeInputCard`)** — the card surfaces which saved doc backs the current text, so the flow sends `resumeId` when the text is doc-backed and untouched, and `resumeText` once it's hand-edited. Pasted text is never silently imported. The card is seeded from a host-owned `docId` because it remounts on every wizard-step revisit; without the seed a post-remount edit compared against a local `null` and the host kept sending a stale doc id for text the user had since changed.

**TailorFlow runs the staged pipeline** — new `useTailorPipeline` wrapping `useResumePipelineSession`, reconnecting through a persisted run key. Stage events fold into four steps via a pure mapper (`analyze_job|match_evidence|strategy → Analyze`, `draft|cover_letter → Generate`, `validate|repair → Validate`, `humanize → Remove AI signs`; an unknown stage keeps the last). Numbered rows become checkmarks as stages finish, each with a one-line description, the thinking bubble, and the streaming pane. The depth picker is gone from the wizard. The fast-path tailor engine and its store are deleted, dead after the cutover.

## Review findings resolved in-branch

Four Opus reviews ran (rust-backend-architect, tauri-security-reviewer, frontend-reviewer, ui-ux-expert). Two rounds of blocking findings, all fixed and mutation-verified — each guard was reverted, watched go red, and restored:

- **Infinite update loop** — the run-persist effect had an unstable callback in its deps and no already-persisted guard, so persist → store update → new callback identity → refire. Fixed with a keyed guard plus a ref-held callback. The keying is itself asserted, so a legitimate second run still persists.
- **Cross-application run leak** — the persisted run key wasn't tied to an application, so opening application B could reconnect to A's run. Replaced with a self-describing `applyRun: { forId, runId, jobId }` read only when `forId` matches, correct on first render rather than via a reset effect. The reviewer confirmed the router does not remount on a param change, so the pre-existing `key={id}` on the page is a load-bearing part of this — noted for future readers.
- **A raw i18n key in the timed-out banner** — `pipeline.status.*` is `running/completed/needsReview/failed/cancelled` but the panel's own states are `done/needsReview/cancelled/error`, so a timed-out-but-saved run rendered the literal dotted key next to its stop reason, in both locales. The existing test could not catch it: a key-echoing `t` mock is structurally blind to a missing key. Fixed with an exhaustive `Record<TailorRunState, …>` (a new state fails to type-check until mapped) plus a test that asserts every reachable dynamic key against the **real** locale JSON.
- **Focus escaping an open dialog** — the stage-change focus effect fired against an open modal (the focus trap only intercepts Tab, not a programmatic `.focus()`), reachable via "start a run → open Interview questions → run finishes". Now bails when focus is inside `[aria-modal="true"]`, and no longer fires on mount.
- **Contrast** — the "all steps completed" confirmation and the streaming-target header were `text-foreground/45` (4.07:1 dark, below AA and below the `<14px` floor documented in `docs/DESIGN_SYSTEM.md`). Both at `/70` now. Every changed colour was independently recomputed in both themes, accounting for the light-scheme opacity lifts.
- Also: the 4th step now actually marks as done (including on the zero-cost humanize path), the panel fits the 900×600 floor with the thinking bubble collapsed by default, an elapsed timer keeps a long stage from looking frozen, a timed-out-but-saved run no longer reads as a clean success, a needsReview run with zero claims has a way forward, screen-reader users get per-row state plus a polite announcer that now also announces completion, and cancelling with no output is acknowledged.

Two items were deliberately left out of scope and are **not** regressions from this PR: the thinking-transcript contrast (3.75:1 light, pre-existing — worth its own ticket) and the elapsed-clock resetting on reconnect. One reviewer claim was refuted rather than acted on: the Cancel button was recomputed at 4.63:1 and passes AA as shipped.

## Testing

`pnpm typecheck` (13/13) · `pnpm test` (488 files / 7118 tests) · `pnpm lint:strict --max-warnings 0` · en/de key parity, zero orphans either direction · full Rust suite green via the pre-push hook.

Known gap, stated rather than papered over: the newly wired completion announcement can't fire in production today, because the panel unmounts before the step counter passes the last step. The durable completion signal is the results panel's own screen-reader line. The branch is wired correctly for when that changes and is covered directly by test.

## Summary by CodeRabbit

* **New Features**
  * Added a four-step tailoring progress view for analysis, generation, validation, and humanization.
  * Added run status banners, cancellation/error handling, review indicators, and run history.
  * Résumés and job postings can now be provided as saved items or pasted text.
  * Added separate live résumé, cover-letter, and reasoning output streams.
  * Added saved résumé selection persistence and improved editing behavior.

* **Bug Fixes**
  * Prevented runs from appearing under the wrong application.
  * Improved notification labels by normalizing spacing and escaping markup.
  * Strengthened protection against prompt-content injection.
  * Improved handling of missing or invalid saved inputs.


saeedkolivand:
@coderabbitai review

saeedkolivand:
All 14 findings addressed in `bc84dee6`…`5cda944c`. None were refuted — every one reproduced. Each fix was mutation-tested (revert the production behavior → confirm the test goes red → restore → confirm green), which mattered here because four of the findings were *tests that could not fail*.

Three fixes deliberately deviate from the suggested diff:

**`handleSelectSaved` (useResumeInput.ts)** — took "don't select when there is no text" over "fetch the text like `handleFileChange`". The list response already carries the full text from the same source `getText` reads, so the re-fetch would be redundant. Also audited every other `selectDoc` call site for the same shape (id set without the matching text landing); no further instance.

**`pipeline.status.done`** — used an exhaustive `Record<TailorRunState, …>` rather than extending the ternary, so a future `TailorRunState` member fails to type-check until it is mapped. The accompanying test asserts against the real locale JSON rather than the mocked `t` — as noted, a key-echoing mock is structurally blind to a missing key, which is why the original test passed while rendering the raw key.

**The wiring-check assertions (`resume_pipeline/test.rs`)** — neither normalized nor deleted. Normalizing would leave the argument spelling pinned; deleting would drop a wiring guarantee nothing else covers (that `execute` routes through the pure resolvers instead of re-implementing the branch inline). Narrowed to name + open-paren only, then verified both directions: a rustfmt-style re-wrap that took the old literal to 0 matches leaves the narrowed one at 1, and replacing the call with an inline branch still fails it.

Two notes on the remaining fixes: the i18n sweep now iterates `STOPPED_SUFFIX` + `UNKNOWN_STOPPED_SUFFIX` rather than a hand-listed set, since that is the one key path with an open input set — the sweep's header claim has been narrowed to what it actually checks. And `ResumePipelineRunRequest` intersects two separate two-member unions per the warning about nesting a single `RequireOneOf`, guarded by `@ts-expect-error` on `{}`, résumé-only, and job-only, verified by loosening the type and watching the directives go unused.

Gates: `cargo test --workspace` 4232 + 18 architecture + 3 eval · `cargo clippy --all-targets --all-features --locked -D warnings` clean · `pnpm typecheck` 13/13 · `pnpm test` 488 files / 7145 tests · `pnpm lint:strict --max-warnings 0` · en/de parity, zero orphans.

@coderabbitai review

saeedkolivand:
Both fixed in `3dfc755d`/`6ed2bea8`, each mutation-tested.

**Live region** — added the missing clear on the null branch, so a later clean run no longer exposes stale cancelled/needs-review text. On the second sub-point (an identical consecutive announcement not re-firing because React bails a same-value setState): no extra machinery added, and the reasoning is now a code comment — `key` cannot reach the same terminal value twice without passing through `null`, because starting a new run moves `stage` off `configuring`/`done` first. The new clear therefore always lands between two occurrences, making the second a real `''`→text transition.

**i18n test** — switched to the `@ajh/translations` entrypoint. One correction to the finding: the five sibling `*.i18n.test.ts` files use `i18n.exists(key, { lng, fallbackLng: false })` + `getFixedT`, not `getResourceBundle`; the latter belongs to the separate global parity test that their header comments merely reference. Matched the actual sibling pattern rather than the literal suggestion. Verified before rewriting that the entrypoint exposes the real shipped catalogs (`packages/translations/src/index.ts` statically imports the locale JSON into `i18next.init`) so the test still reads reality, and `fallbackLng: false` preserves per-locale gap detection.

Also strengthened while in there: every case is now a `(locale, key)` cross-product instead of checking one key against both locales in a single case, so a gap in one locale fails its own case rather than hiding behind the other's pass — 62 cases, up from 31. Mutation-tested by deleting `pipeline.runs.title` from `de` only: exactly the `de` case went red while the `en` case for the same key stayed green.

@coderabbitai review

saeedkolivand:

PR #994, opened 2026-08-15
build: pin the rust toolchain across local, ci, and release

saeedkolivand:
Pins the Rust toolchain so local builds, PR CI, and release builds stop being free to disagree.

## The problem

There was no `rust-toolchain.toml` and no `rust-version` in the crate manifest, and CI selected `stable` by ref name — `dtolnay/rust-toolchain@<sha> # stable`. So each surface got whatever `stable` meant that day.

Not hypothetical: a bare `cargo fmt --all` currently mass-reformats the repo, because some committed files predate the installed rustfmt, while the canonical files *do* pass `cargo fmt --check`. That is the signature of an unpinned formatter.

## The part that would have made a config-only fix useless

`dtolnay/rust-toolchain` **does not read `rust-toolchain.toml`.** Verified against the action's own source: the version comes entirely from which `@rev` you check out — `@stable` delegates to `rustup toolchain install stable` at runtime. Adding the config file alone would have pinned this machine while CI kept floating, and looked complete.

So every Rust job had to move onto a version-specific action revision. All 12 workflows were enumerated: 9 jobs go through `.github/actions/setup-rust`, and `release.yml` has its own inline setup step that does not use the composite action and needed the same change separately. The other 8 workflows touch no Rust at all.

## What it pins to, and why 1.97.1

`1.97.1` — what CI has actually been validating against (`stable` resolves there today; released 2026-07-16). An earlier revision of this branch pinned `1.96.1`, the version already installed locally, which read as a freeze from the dev machine but was silently a **downgrade for CI**. Pinning to what CI already ran preserves the tested configuration instead of reverting it; `rustup` fetches it locally from the toolchain file.

`rust-version = "1.95"` (MSRV) was derived from the dependency graph rather than picked: scanned `rust_version` across all resolved packages (744 of 1075 declare one). Highest floor is `sysinfo` at `1.95`; next is `serial_test` at `1.93.1`, a dev-dependency that `cargo test`/`clippy --all-targets` still compile. Cannot shift resolution — edition 2021 means resolver v2, and every CI invocation passes `--locked`.

## Deliberately not included

No `components` list in the toolchain file. Verified empirically that `rustup` *does* honour one — but also that the default profile already installs `rustfmt` and `clippy`, and the one specialist component (`llvm-tools-preview`, for coverage) is needed by a single job that already requests it. Listing them centrally just made every Rust job fetch things it does not use.

No `msrv` field in `clippy.toml`. It would surface `clippy::incompatible_msrv` findings unrelated to this change, and this PR is a pin, not an upgrade — flagged rather than silently skipped.

`release.yml`'s inline setup was left inline rather than refactored onto the shared composite action; it customises the cache key per macOS arch, which the composite action does not expose. Smallest diff wins.

## Testing

`cargo fmt --all -- --check` clean — no reformat was triggered, so there is nothing to land separately. `cargo test --workspace --locked` 4175 passed, identical to the run under the previous toolchain. `cargo clippy --all-targets --all-features --locked -D warnings` clean. `pnpm gen:workflows:check` no drift. `pnpm typecheck` 13/13. `rustup show` confirms the pin resolves via the toolchain file.

Split out from the agentic-deletion PR deliberately: this is the one change that alters how everything compiles, so keeping it alone means a red run on either PR points somewhere unambiguous.


saeedkolivand:

PR #995, opened 2026-08-15
feat: delete the agentic surfaces and extract the prompt fencing

saeedkolivand:
Deletes the agentic surfaces — the "Prep this application" flow and `improve_resume` — after first extracting the prompt-injection fencing that lived inside them. **98 files, +1,113 / −19,013.** Last of five PRs in the multi-stage pipeline programme, after #990, #991, #992 and #993.

## Why

The owner ruled the agentic machinery was spending tokens without producing anything worth keeping. The **browser extension's answer-assist stays** — that is the one agent-adjacent capability deliberately kept.

## Extraction first, deletion second

`agent/` could not simply be deleted: a large amount of ADR-010 fencing machinery lived inside it and is used by code that survives — the staged résumé pipeline (`pipeline/resume/prompts.rs`), which fences scraped job postings and prior-stage model output, and the extension's answer-assist.

So step 1 moved `fenced`, the caps, `neutralize_transcript_boundaries`, `FENCE_TAG_PATTERNS` and its neutralization functions into a new dependency-free `prompt_fence` module (L0), and re-pointed every consumer. The security review confirmed the move was faithful at the byte level: a comment-stripped diff of the moved region yields exactly one hunk (the three items deliberately left behind), and all 16 moved tests are byte-identical — including the two subtlest bits, the attribute-form pattern and the inner-`<` breaker.

Steps 2–4 then deleted `agent/`, `commands/agent.rs`, the renderer surfaces (`PrepApplicationPanel`, `AgentConfirm`, `ImproveResumeRun`, `useAgentRunSession`, `services/use-agent/`, `agent-run.machine`, and `TailoredResumePanel`), and the IPC contract layer.

## The finding that justifies the review cost

`FENCE_TAG_PATTERNS` is an allowlist where a **missing entry is a prompt-injection hole**. The security reviewer removed each of the 24 entries individually and ran the suite after each: **15 could be deleted with all 3,968 tests green — 10 of them tags surviving code still emits.** Only 9 were guarded at all.

That is not theoretical. A docs agent trimmed three entries from this list earlier the same day while "fixing a comment"; it was caught by a human, not by CI.

The subtlety that made it invisible: `every_untrusted_block_is_fenced_and_forgery_resistant` *does* assert on `</job_analysis>`, but its hostile payload only ever forges `</job_posting>` — so that entry was inert to the assertion. Moving a test set verbatim proves the tests did not weaken; it proves nothing about entries no test forges.

Now closed by a table-driven test. Note it does **not** simply iterate the live registry — that cannot catch its own deletions, since removing an entry just means the loop tests one fewer thing and passes. It pairs the loop with an independent hardcoded list and a set-equality check, so a deletion fails the equality check and a weakening fails the loop.

## Also from review

- **Extension fence coverage** went from 1 of 6 untrusted blocks to all 6, including the two with the strongest attacker story (`company_research`, `web_search_notes` — both web-sourced). Behaviour was already correct; the coverage was not. Breaking `fenced` now reddens the pipeline and the extension **separately**, rather than both relying on one distant test.
- **Three more orphans**, found by a better method than delete-and-compile-one-at-a-time: temporarily make every module private, compile, and diff the unused-item set against the same run at the merge base. `report::for_saved_resume`, `report::agent_save_pipeline`, and `PipelineRunStore::update_status_if_present` — the last of which delete-and-compile missed because its own tests still referenced it.
- **`prompt_fence` is now genuinely L0.** `RESUME_CAP` was imported from generated code, forcing an architecture exception spanning L0→L3 — and that constant had no TS consumer left. It is a local literal now, which cleared the exception, `packages/shared/src/agent-caps.ts`, and its codegen target.
- **The error-echo clamp** in `resolve.rs` lost its only guard when `commands/agent.rs` was deleted; replacing its body with a no-op left all tests green. Now asserted.

## Verified, not assumed

The architecture review confirmed the deletion changed no surviving behaviour: `AUTOPILOT_NOTE_SYSTEM` relocated byte-identical, `limits::AGENT_RUN_*` values unchanged (now single-consumer), the `StoppedReason` wire contract intact, no migration or data-store schema touched, and en/de translations symmetric at −164 each. `Budget::RESUME_QUALITY` compared field-by-field against base: identical.

It also confirmed the provider tool-calling surface (`AiProvider::chat_with_tools`, `ToolSpec`, `AgentTurn` and five implementations) is now genuinely unreachable — **flagged, not deleted**, since it belongs to the AI-provider domain. Important for whoever acts on it: `ChatMsg`, `flatten_messages`, `Usage` and `StopReason` are still live, and the removal is code *plus* tests.

## Rider

`fix/reset-stale-refs` is folded in — a one-file renderer fix, file-disjoint from the deletion. `reset()` cleared state but left three refs stale until React committed, so a late event for the just-reset run could overwrite the new session's stage and error text. The state machine already protected the coarse status field, so the symptom was a reset screen flashing the previous run's progress. Mutation-verified.

## Testing

`cargo test --workspace` 3,970 passed · `cargo clippy --all-targets --all-features --locked -D warnings` clean · `cargo fmt --check` · `gen:ipc:check` + `gen:prompts:check` · `pnpm typecheck` 13/13 · `pnpm test` 478 files / 6,855 tests · `pnpm lint:strict --max-warnings 0`.

## Summary by CodeRabbit

* **New Features**
  * Added centralized prompt-safety handling to neutralize forged instructions and tool-result markers in generated content.
  * Added regression coverage for prompt-boundary protection and session reset race conditions.

* **Removed Features**
  * Removed application preparation, résumé tailoring, résumé improvement, confirmation, and related progress interfaces.
  * Removed the legacy agent-run commands and associated event/API contracts.

* **Bug Fixes**
  * Prevented stale pipeline events from affecting a newly reset session.
  * Limited echoed résumé and job identifiers in resolution errors to 64 characters.


saeedkolivand:
All seven addressed. Two are worth more than their severity labels suggest.

**The self-closing fence gap was real, and the suggested fix was not sufficient.** `(?:\s[^>]*|/)?` closes `<job_posting/>` but still misses `<job_posting attrs>` — attributes immediately after the slash, no whitespace. Implemented `((?:\s|/)[^>]*)?` instead, which opens on whitespace *or* `/` and reads to the next `>`, closing that case and duplicated slashes (`<tag//>`) too. Each spelling is pinned by a new test row.

The coverage hole that let it through is closed at the class level, not the instance: `every_registered_fence_tag_is_load_bearing` now probes **every** registered tag with self-closing and plain-opening forgeries, not only the closing form. Mutation-verified — reverting the pattern reddens exactly that test plus `fenced_neutralizes_whitespace_and_case_variants`, nothing else.

Worth recording that this is the third distinct gap found in this one allowlist in a day: three entries silently trimmed by an automated docs pass, 15 of 24 entries with no guard at all, and now a spelling the pattern could never match. The list is small; the failure modes are not.

**The reset regression test could pass without running.** `bus.stage?.(...)` skipped silently when the listener was absent. Now asserts both listeners exist before invoking, and emits a stale `job.failed` in the same synchronous scope asserting `state` stays `idle` and `error` stays `null`.

**On the request to cover all three refs — partially achieved, and reporting it honestly rather than claiming otherwise.** Mutation-testing each `reset()` assignment independently:

- `runIdRef` — reverting alone **reddens**. Independently covered.
- `jobIdRef` — reverting alone stays green.
- `busyRef` — reverting alone stays green; reverting both together also stays green.

The reason is structural rather than a test gap. `handleJobEvent` guards with two short-circuited checks (`!jobIdRef.current || mismatch`, then `!busyRef.current`), so either one alone blocks the event regardless of the other. And in the only state where `RESET` is a valid transition, `busy` is already `false` from the ordinary per-render sync before `reset()` runs — making that particular line redundant for the reachable scenario. The two are validated **jointly**, not independently. The added probe still pins real behaviour worth having: a stale `job.failed` for a just-reset run must never resurrect the error state.

**Also fixed:** dangling rustdoc intra-doc links to the deleted budget constants (swept the crate — every other agent-named link resolves to a live provider symbol); `every_shipped_budget_is_internally_consistent` now driven from a `SHIPPED_BUDGETS` slice so a future budget is covered by construction rather than silently not; and the stale comment claiming those refs are written only during render, which described the pre-fix bug rather than the fix.

The evidence-module nitpick was verified real and reworded within its one-line scope — `strategy` is the only caller of `extract_evidence` and reads only `.roles`. No sorting behaviour changed.

Gates: `cargo test --workspace` 3,970 passed · `cargo clippy --all-targets --all-features --locked -D warnings` · `cargo fmt --check` · `cargo doc --no-deps` (confirms the link fix) · `gen:ipc:check` + `gen:prompts:check` · `pnpm typecheck` 13/13 · `pnpm test` 478 files / 6,856 tests · `pnpm lint:strict --max-warnings 0`.

saeedkolivand:

PR #999, opened 2026-08-16
docs: make adr-034's placeholder note a thin pointer

saeedkolivand:
Follow-up to #998, addressing the one review finding that was still open when that PR merged.

ADR-034's Consequence #2 update restated the placeholder predicate's matching rules (bracket syntax, en/de tokens, case-insensitivity, punctuation tolerance) and the validation call path, against the knowledge base's own rule that docs are thin pointers into the owning symbol with no copied literals.

It had already begun to drift, in the same PR that wrote it: the text still named `locale::letter::is_template_placeholder` after #998 moved the symbol to the module root.

## What was cut

The defect narrative, the predicate's matching rules, and the parser mechanics. Those belong with `locale::is_template_placeholder`, `parse_cover_letter`, and `validate::content::letter`.

## What was kept, deliberately

Two things that are contract rather than mechanics, and are not derivable from reading a symbol:

- **`validate::content` is never called from `export/`.** The export path calls `validate::validate_and_fix`, a structural PDF/DOCX round-trip checker. So `letter.template_placeholder` does **not** block an export — the parser filter is what stops bad bytes reaching rendered output. An earlier draft of this ADR claimed the opposite and had to be corrected once already.
- **Only the placeholder shape of drift is closed.** Other unanticipated furniture shapes are still caught only by human review.

## Note on style

The symbol references are bare backtick paths rather than markdown links. All 34 sibling ADRs use that form and this file's own Related section already did, so linking would have made the file internally inconsistent — fixing a consistency finding by introducing a different inconsistency.

Docs only; no source files touched.


saeedkolivand:
🎉 This is included in version 0.137.0.

PR #1000, opened 2026-08-16
docs: fold the second adr tree into the knowledge base

saeedkolivand:
The repo had **two ADR directories with colliding numbers**. `docs/adr/0013` is email watching; `adr-013` is the resume builder. Both get cited as "ADR-13", and it has already caused a wrong citation.

## Which one is canonical

`docs/knowledge/decision-records/` — created 2026-05-31 with the knowledge base, indexed in its README, owned by `project-steward`, and the only one the `docs-standards` skill points at.

`docs/adr/` appeared **twelve days later** (2026-06-12), carried no index and no owner, and nothing in the repo's rules referenced it. It grew to 23 records because nothing stopped it.

## Moved, deliberately not renumbered

An ADR is a dated record, and its number is an identifier other things point at — commit messages, merged PR bodies, code comments, and the published tech radar. `apps/landing/public/benchmarks/data.js` and `metrics/*.json` are **generated captures of commit text** containing strings like `ADR-0021` and `ADR 0005`.

Renumbering would falsify all of it, so the `NNNN-` series is folded in as-is and marked **closed** in the index: cite existing ones by their current numbers, give every new ADR the next `adr-NNN`. This is standard ADR practice — supersede, never renumber.

Git tracked all 23 as renames, so `git log --follow` still works.

## What follows the move

- 24 live files updated, including the tech radar's `ADR_DIR` and its **public GitHub blob links** (those would otherwise 404)
- Relative links *inside* the moved files gained a directory level; each was verified to resolve
- `docs/knowledge/README.md` now indexes all **57** ADRs, with the closed series and the reasoning recorded
- Generated snapshots under `apps/landing/public/` deliberately left untouched as the historical records they are

## Reviewer note — a user-visible consequence

This changes **published URLs**. Anyone holding a bookmarked `docs/adr/...` link will 404, and GitHub does not redirect moved paths. They would have broken on any move; calling it out rather than burying it.

## Verification

`check:tech-radar` (40 entries, every `adrSlug` resolves) · `check:agent-system` (which explicitly validates the ADR index) · `check:landing-drift` · typecheck · lint · prettier · full TS suite **478 files / 6898 tests**.

The pre-push gate caught one real breakage: `check-tech-radar.test.mjs` writes its fixture ADR via `join(repoDir, 'docs', 'adr', …)`. A path assembled from segments is invisible to a `docs/adr/` grep, so the reference sweep missed it. Fixture repointed; assertions unchanged.

## Summary by CodeRabbit

* **Documentation**
  * Updated decision-record links throughout the project to use their current location.
  * Added an index and guidance for the complete decision-record series.
  * Corrected several relative links and audit references.
  * Added decision records covering recent architecture, privacy, AI, landing-page, and workflow decisions.
* **Validation**
  * Updated tech-radar checks and tests to validate decision records from the new location.


saeedkolivand:
🎉 This is included in version 0.137.0.

PR #1007, opened 2026-08-17
fix(resume): localize every section heading, not the first four

saeedkolivand:
## The bug

A German résumé rendered **Projekte** and **Ausbildung & Sprachen** as ordinary
body text instead of section headings — owner-reported, screenshot-confirmed.
Profil, Kenntnisse, Berufserfahrung and Ausbildung looked right; everything
below did not.

Two independent halves, both required:

**Producer.** `ResumeConventions.headers` covered exactly 4 of the 9 ordered
`SectionId`s. The other five — Projects, Certifications, Languages, Awards,
Publications — were handed to the model as their raw English `SectionId` debug
word inside the "Sections run in this order" list, while the same prompt told it
to write German. The model did the reasonable thing and invented its own German
words. `headers` is now a **total record** over all nine ids in every locale, so
the TS typecheck refuses to build a locale that leaves one unnamed.

**Recogniser.** Nothing in `export::parser`'s `SECTION_NAMES` knew "Projekte",
so on the re-parse it was a paragraph, not a heading. 27 localized headings
added, plus an `"X & Y"` arm for the merged shape models emit often enough that
the draft prompt already has a rule against it.

Italy came along for the ride because `pick_locale` forwards the id verbatim and
`section_order_for` matches the literal `"it"`: without a real `LocaleProfile`,
an Italian user silently got the US section order. `IT_ORDER` runs Languages
before Skills — the one position it deliberately does not mirror the German
order, following Europass. Flagged as a reviewable call, not copied.

## Review round

A pre-PR pass found three guards on this branch that could not fail for the
reason they were written:

- the producer/recogniser guard hardcoded **both** axes its doc said it
  enumerated, so a locale added on the TS side was never visited;
- `header()` falls back to the English word rather than panicking, and nothing
  tied the TS id union to the ids `section_order_for` emits — adding
  `SectionId::Volunteer` to an order const put the literal word "Volunteer" into
  a prompt demanding Italian with every gate green;
- the collision sweep named its cases as literals and missed three of the four
  German words this branch adds.

All three now enumerate generated data. The rewrite found a **live pre-existing
defect**: Portuguese `Competências` classified as `Other`, so a pt résumé
generated by this app did not recognise its own skills section — the `ê` blocks
the `competenc` stem exactly as `z` blocked it for Italian.

It also found that `isKnownSectionName` never gained the `" & "` arm Rust has.
The shared section-names fixture compares two name *lists* and is structurally
blind to a rule that combines two entries, so the TS mirror answered `false` for
every merged heading with all tests green. Mirrored, with a new shared
`section-name-joins.json` pinning the *predicate* on both sides.

Two guard drafts were wrong on first run and are worth naming, since both are
the shape this repo keeps shipping:

- `header(id) != id` reads as a fallback for French **Certifications**, which is
  correctly spelled the same as English. Membership of the key is the property
  that actually separates a translation from a miss. The inequality form failed
  on `fr` the first time it ran.
- the collision sweep initially allowed `want || Other`, which made it unable to
  fail — a heading the recogniser does not know reads identically to one it
  knows correctly. Measured: every bucketed id already lands on its own bucket
  in all 7 locales, so the slack bought nothing.

Every guard here was mutation-checked — green at baseline, red with the feature
deleted, restored.

## Verification

`cargo test --all-features --locked`: 4191 lib + 18 architecture + 20 egress + 3
eval. `pnpm test`: 478 files / 6922 tests. clippy `--all-features --all-targets`
clean, `cargo fmt --check` clean, `gen:prompts:check` up to date, typecheck 13/13,
`lint:strict --max-warnings 0` clean, `check:agent-system` in sync.

## Deliberately out of scope

- `SectionId::from_header` stays English-only. A localized heading becomes
  `Custom` and keeps its position, so there is no silent reorder — note that
  this does **not** hold for English joins: `"Skills & Certifications"` maps to
  `Skills`, which predates this branch and is not made worse by it.
- Spain and Portugal have the same missing-`LocaleProfile` gap Italy had.
- A newly-recognized heading with **no body under it** is now dropped rather
  than kept as body text (`push_nonempty_section`). The lost line is a bare
  label the prompt itself calls worse than no heading at all.

## Summary by CodeRabbit

- **New Features**
  - Added Italian résumé and job-ad localization, including market-specific section ordering and formatting.
  - Expanded localized résumé headings across German, French, Spanish, Italian, Dutch, and Portuguese.
  - Added support for Projects, Certifications, Languages, Awards, and Publications headings.
- **Improvements**
  - Improved recognition of multilingual section headings, including valid headings joined with “&”.
  - Résumé generation now uses localized section ordering and headings more consistently.
- **Bug Fixes**
  - Improved classification of localized skills and other résumé sections while avoiding partial or ambiguous matches.


saeedkolivand:
@coderabbitai review

PR #1014, opened 2026-08-18
fix(scraping): register jobicy in the shared board list and make the drift impossible

saeedkolivand:
## A board the app offers but the shared contract has never heard of

`SCRAPERS` (`scraping/boards/mod.rs`) holds **24** boards. The renderer's `BOARD_IDS` (`packages/shared/src/schemas/index.ts`) held **23**.

The missing one is **`jobicy`** — and it is not a stub. It is a fully registered scraper with `id() == "jobicy"`, a live-verified HTTP implementation, `listed()` defaulting to `true` so the catalog offers it, and `boards.jobicy` labels already written in **both** `en` and `de`. Everything except the shared list knew about it.

For contrast, the one board that is *deliberately* absent from the picker is `comeet` — it overrides `listed()` to `false`, which is why `engine/test.rs` asserts "23 of the 24 registered boards are listed". And `comeet` **is** in `BOARD_IDS`. So the omission is not a policy, it is an oversight, and neither list could see it because nothing compared them.

## Why it has not bitten yet — and why that is not reassuring

The `z.enum(BOARD_IDS)` looks like it constrains the wire. It does not: `pnpm gen:ipc` lowers enums to strings, so the generated Rust is

```rust
pub struct ScrapeBoardsRequest {
    pub boards: Vec<String>,
    ...
```

and the id that actually crosses IPC is validated by the registry lookup (`boards::get`), which knows `jobicy` perfectly well. The schema's own comment claimed the opposite — that each entry *"is already constrained to a known `BoardId`"* server-side — which is precisely why the gap read as safe to anyone checking. That comment is corrected here rather than left to mislead the next reader.

What is genuinely broken today is the **`BoardId` type**: the renderer programs against a union that omits a board the backend ships. The day anything parses `ScrapeBoardsRequestSchema` at runtime — its stated purpose — `jobicy` breaks.

## The fix

`pnpm gen:ipc` now emits the shared list to `ipc_contracts/board_ids.rs`, exactly the shape as the existing `DATE_FILTER_OPTIONS` / `AI_GENERATE_INTENTS` emitters, so this rides the `gen:ipc:check` CI gate that already exists. A Rust test then compares it against `SCRAPERS` **in both directions**:

- one catches a board added to the registry and forgotten in TS — the `jobicy` case;
- the other catches a board removed from the registry while the TS list still promises it to `BoardId`.

Plus a vacuity guard, because both assertions are "this difference is empty", which is trivially true of two empty lists.

## Mutation-checked

Removing `jobicy` from the shared list again and regenerating:

```
registered here but absent from packages/shared BOARD_IDS: ["jobicy"]
  — add them there and run `pnpm gen:ipc`
the two lists must have the same length, not merely overlap
```

## One thing deliberately NOT done

The audit item that prompted this also asked to derive `assert_eq!(health.scrapers.len(), 24)` from `SCRAPERS.len()`. I left the literal alone: driving that assertion off the very list it guards makes it **vacuous** — it is the only check that notices a scraper being deleted, and `SCRAPERS.len() == SCRAPERS.len()` notices nothing. The parity test removes the duplication that actually drifts; the literal keeps guarding the thing a self-referential count cannot.


saeedkolivand:
@coderabbitai review

PR #1016, opened 2026-08-18
fix(scripts): stop a doc comment from promoting a hook to a subscriber

saeedkolivand:
Follow-up to #1015. The AI review posted a **HIGH** after the last push and the check went red; #1015 was merged before it was addressed, so this fixes it on `main`.

## The finding was right, and the cause is worse than a stale entry

It reported `features/monitoring/components/MonitoringPage/index.tsx` as declared-but-not-subscribing. Verified: that file imports `useJobQueue`, which is a plain `useQuery` hook with no subscription anywhere in it.

So why did discovery think it was one? Because a declaration slice runs to the *next* `export`, which means **the doc comment belonging to the hook below lands inside the preceding hook's body**. And `use-jobs.ts` documents `useJob` like this:

```
 * HARD REQUIREMENT: a `useJobEvents()` subscription MUST be mounted somewhere in
 * the tree, or this query never updates.
```

That sentence sits in `useJobQueue`'s slice. The transitive closure I added in #1015 saw `useJobEvents(` in the body, promoted the query hook to a subscription hook, and dragged every file calling it into the inventory. **Prose about a hook counted as a call to it.**

This is the same structural fact already recorded in this repo from the other direction — a doc comment sits *above* its item, so anything anchored on the item *before* swallows it.

## The fix

Blank comments and string/template literals before matching. A hook name inside a string is a mention, not a call; a commented-out subscription is not one either.

It is a **scanner, not a regex**, deliberately: `//.*$` would eat the rest of a line after a URL inside a string literal, silently **dropping a real call**. For a guard, a false negative is the one failure direction that cannot be tolerated — it reports green. Characters are replaced one-for-one with spaces so every `match.index` still points at the right place.

## Result

`useJobQueue` correctly drops out — **15 hooks, not 16** — and with it the `MonitoringPage` entry, which is removed.

The other four wrappers were checked by hand and are genuine, so the closure itself was sound:

| hook | why it counts |
| --- | --- |
| `useWorkerActivity` | calls `useJobEvents()` |
| `useAutoIndex` | calls `useJobEvents()` |
| `useMenuIntents` | registers `api.menu.onNavigate` / `onAction` |
| `useUpdater` | subscribes directly |

## Tests

35, up from 29. The new ones pin the shipped bug (the next hook's doc comment), a commented-out call, a listener appearing only inside a string, and that stripping preserves length and line structure so offsets stay valid.

### DIFF
--- a/scripts/check-event-subscriptions.mjs
+++ b/scripts/check-event-subscriptions.mjs
@@ -159,18 +159,12 @@ const SUBSCRIBERS = {
       'per-run state to preserve and nothing is lost by only listening while the monitoring ' +
       'page is open.',
   },
-  'features/monitoring/components/MonitoringPage/index.tsx': {
-    mount: 'route-scoped',
-    note:
-      'ACCEPTED, not debt. Subscribes via useWorkerActivity to render what is running RIGHT ' +
-      'NOW; the reading is re-derived from the job registry on mount, so there is no history ' +
-      'a missed event could have cost it.',
-  },
   'features/dashboard/components/AISystemStatus/index.tsx': {
     mount: 'route-scoped',
     note:
-      'ACCEPTED, not debt. Same shape as the monitoring page — useWorkerActivity feeding a ' +
-      'live "what is busy" readout, re-derived from the job registry on mount.',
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@coderabbitai review

PR #1017, opened 2026-08-18
test(extension): pin the mv3 permission surface against the readme's promises

saeedkolivand:
Track **D2** of the architecture audit — and the last open piece of Track D (D1 shipped as the egress inventory).

## Why

`manifest.ts` is the source of truth for claims this project makes **outside this repo**: to Chrome Web Store and AMO reviewers, in the extension README's "Permissions — minimal & justified" table, and on the landing privacy page.

Nothing asserted any of it. Widening a permission is a two-line edit, and the widened build would reach store review — and the user — with the old justification still printed beside it.

Store review is also the wrong place to find out. A rejection costs a release cycle; an **accepted** over-broad permission costs the claim itself.

## What is pinned, and why in this shape

- **The permission list, written out by hand.** A test that loops over the thing it guards catches additions only — delete an entry and the assertion deletes itself. The literal is the half that notices a *removal*.
- **A denylist** of permissions the extension must never hold. This is the half that survives a careless update, and it's the one that matters most.
- **Host permissions checked as a property**, not by string equality: every entry must *parse* to loopback.
- **Both targets compared to each other**, so a per-target delta cannot smuggle a permission into one build.
- **MV3, no CSP loosening, no static `content_scripts`** — Scan mode injects on a click bounded by `activeTab`; a static block would run unbidden on every matching page, which is a different product.
- **Firefox's `data_collection_permissions: ['none']`** — the assertion to AMO that nothing leaves the device. If that ever stops being true, this test failing is the intended prompt to declare the real categories rather than ship a false one.

## Mutation-checked

| mutation | result |
| --- | --- |
| add `tabs` to the manifest **and to the allowed literal** — the careless-update case | ✅ denylist fails it anyway, both targets |
| swap a host for `http://127.0.0.1.evil.com/*` — a different origin that merely *starts* the same way | ✅ 5 tests fail, including the loopback property |

The second is exactly why the host check parses instead of comparing strings.

## README parity, both directions

The table is the justification a store reviewer and a user actually read, so:

- a permission present in the build and absent from the table is an **undocumented capability**;
- deleting the "we do **not** request `<all_urls>`, `tabs`, `webRequest`" paragraph now fails too — otherwise the denylist would still be enforced while corresponding to nothing published, and the next reader would have no way to know the guard was protecting a promise.

## No new tooling

Test files are excluded from the extension tsconfig, and `rewrite-preset-parity.test.ts` already reads with `node:fs`, so the README check rides an existing precedent.

### DIFF
--- a/apps/extension/src/manifest.test.ts
+++ b/apps/extension/src/manifest.test.ts
@@ -1,0 +1,185 @@
+/**
+ * The MV3 permission surface is pinned, and the README's published table agrees
+ * with it.
+ *
+ * ## Why this exists
+ *
+ * `manifest.ts` is the source of truth for a set of claims the project makes
+ * OUTSIDE this repo: to Chrome Web Store and AMO reviewers, in the extension
+ * README's "Permissions — minimal & justified" table, and on the landing site's
+ * privacy page. Widening a permission is a two-line edit that changes what the
+ * extension is allowed to do to every page a user visits — and nothing asserted
+ * any of it, so the widened build would reach store review, and the user, with
+ * the old justification still printed beside it.
+ *
+ * Store review is also the wrong place to find out. A rejection costs a release
+ * cycle; an ACCEPTED over-broad permission costs the claim itself.
+ *
+ * ## What is pinned, and why in this shape
+ *
+ * The permission list is written out **by hand** rather than derived from the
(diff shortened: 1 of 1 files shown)


saeedkolivand:
@coderabbitai review

PR #1018, opened 2026-08-18
fix(autopilot): jitter scheduled runs so default installs stop herding

saeedkolivand:
Track **B3** of the architecture audit.

## Why this one is not theoretical

Schedules are clock-anchored, the tick is 60 s wide, and every record created before the run-time picker defaults to **09:00**. So every default-schedule install in a time zone hits the same third-party APIs inside the same minute.

The desktop log already shows scheduled runs failing this way:

```
[autopilot] board 'aggregator' failed (error='adzuna: HTTP 503; ...')
```

## The design decision that matters

**The offset is deterministic per autopilot, derived from its id — not random.**

A random delay would have meant sleeping inside the tick, and that breaks the two properties everything else in this scheduler is built on: catch-up after a missed occurrence, and no-double-run once `lastRunAt` is stamped at or after the occurrence. Both come from the occurrence being a *single well-defined instant*.

Shifting the **occurrence** instead of delaying the **run** keeps that instant well-defined. Catch-up and no-double-run still hold unchanged, and the tests stay reproducible.

Two smaller calls:

- **FNV-1a, not `DefaultHasher`** — the standard hasher is explicitly not guaranteed stable across Rust releases, and an offset that moved on a toolchain bump would silently shift every user's schedule. Ids are UUIDs, so the low bits are already well distributed.
- **Ten minutes** — wide enough to spread a herd across ten tick windows, far enough under the shortest interval (hourly) that a shifted occurrence can never overtake the next one, and small enough that someone who picked 09:00 isn't surprised by what they see.

That upper bound is a `const _: () = assert!(...)` beside the constant rather than a test. A test comparing two constants is a tautology (clippy rejects it, correctly), and this makes a bad value **unbuildable** rather than merely reported.

## Tests

Beyond the offset being stable and in range:

- **the occurrence moves by exactly the offset** — not "roughly", not "is later";
- **the behaviour a user would notice**: between the nominal time and the offset, today's occurrence has *not* been reached, so the most recent one is still yesterday's. A careless implementation gets this backwards and fires early;
- **a full-day, minute-by-minute sweep** asserting exactly one transition — one run per day, no double-run, no skipped day. This is the property the whole feature must not break.

Every pre-existing occurrence test now passes an explicit zero offset rather than inheriting a default, so they still describe the pure clock arithmetic.

## Mutation-checked

Making the offset always zero fails `jitter_actually_spreads_a_herd`:

```
200 autopilots landed in only 1 distinct minutes — the herd is not spread
```

Without that assertion the whole change could be an elaborate no-op and every other test would still pass.

## One thing to note

During verification a single test failed once in a full `cargo test --all-features --locked` run and did not reproduce across two subsequent full runs (4218 passing). The scheduler's `is_due` tests read the live clock, so a minute-boundary flake is the plausible explanation — but I could not reproduce it, so I am flagging it rather than claiming it didn't happen.

### DIFF
--- a/apps/desktop/src-tauri/src/autopilot_scheduler.rs
+++ b/apps/desktop/src-tauri/src/autopilot_scheduler.rs
@@ -66,6 +66,50 @@ const RETRY_BACKOFF: Duration = Duration::from_secs(12 * 60);
 const DEFAULT_HOUR: u32 = 9;
 const DEFAULT_MINUTE: u32 = 0;
 
+/// Width of the per-autopilot schedule jitter window.
+///
+/// Every install created before the run-time picker defaults to 09:00
+/// ([`DEFAULT_HOUR`]/[`DEFAULT_MINUTE`]), and the tick is 60 s wide — so without
+/// this, every default-schedule install in a time zone hits the same third-party
+/// APIs inside the same minute. That is a thundering herd against hosts we do
+/// not own, and this repo's own logs already show `adzuna: HTTP 503` on
+/// scheduled runs.
+///
+/// Ten minutes, deliberately: wide enough to spread a herd across ten tick
+/// windows, far enough under the shortest interval (hourly) that a shifted
+/// occurrence can never overtake the next one, and small enough that a user who
+/// chose 09:00 is not surprised by what they see.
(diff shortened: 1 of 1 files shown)


saeedkolivand:
@coderabbitai review

PR #1019, opened 2026-08-18
fix(jobs): keep the scrape state alive across a route change

saeedkolivand:
The worst instance of the defect class #1015 now guards: state describing long-running Rust work kept in a route-scoped React component. Three defects, one of them **data loss**.

## 1. "Show more" after a route change wiped the persisted posting list

`startScrape` compares the current search signature against the last one; a difference latches `replace: true`, which makes the backend **clear the postings cache** the instant the first new item streams.

Both halves of that comparison were per-mount. Run a search → go to Settings → come back → press **Show more**, and the entire result set is destroyed by a button whose whole contract is *append*.

My original trace said the search signature alone was at fault. **That was wrong, and the fix is bigger:** `scrapeForm` was also `useState` in `JobsPage`, so a route change reset the *criteria* too. Even with the signature restored, the form reads back as `{boards:[aggregator], query:'', amount:25}` and the comparison legitimately says "different search" → same wipe. Mutation **B** below is that proof: signature stored, form left local, and the end-to-end test is still red. The signature is *derived* from the form, and two derived-from lifetimes that differ desync — the same argument that forced the replace latch to move with it.

"Show more" renders unconditionally whenever there are rows, so the repro needs nothing unusual.

## 2. The in-flight scrape became an uncancellable orphan

`scrapeJobId` died with the component. On remount the command bar showed no progress and no Cancel — while the always-mounted status bar still showed it running — and `startScrape`'s "cancel the previous scrape first" step read a now-null id, so **the next search did not cancel the orphan either**: two scrapes writing into one postings cache.

## 3. Per-board diagnostics were lost

`scrapeOutcome` and the summaries hold the "aggregator: 429 rate limited" strip — the one thing that explains an empty result.

**Storing them was not sufficient**, and this is the second thing the trace missed: `useJobEvents` is page-scoped, so `job.completed` is emitted to nobody while unmounted. The job tracker keeps the identical payload, so the watchdog now recovers `result.boards` under `finishScrape`'s existing jobId guard. Five lines, and it's what actually closes defect 3.

## What moved, and what deliberately did not

**Into `JobsSlice`** (which already holds `listScrollTop`, documented as "restored on remount"): the search signature, the replace latch, `scrapeJobId`, the scrape outcome, the summaries, the failure note, and `scrapeForm`.

**Left local, on purpose:** `livePostings` (transient buffer — the backend cache plus the existing throttled invalidation rehydrates it), `pendingFinishRef` (cannot outlive the call that needs it), and **`scraping`**, which stays `useState` seeded from the stored job id. The watchdog remains the reconciler: it keys on `scrapeJobId`, polls the authoritative terminal state, and now settles a remount-onto-a-finished-job on the first tick — tested explicitly, rather than "eventually".

## Mutation checks

Each deletes the *feature*, not a constant:

| | mutation | red tests |
| --- | --- | --- |
| A | signature back to a per-mount ref | 2 |
| B | `scrapeForm` back to `useState` | 1 (the end-to-end one) |
| C | `scrapeJobId` back to `useState` | 10 |
| D | summaries/failure note back to `useState` | 3 |
| E | drop only the watchdog boards recovery | 2 |
| F | `patchScrapeForm` spreads the captured form | 1 |
| G | drop the `job.stream` identity guard | 1 |

The "Show more" test asserts the **wire payload** — `replace === undefined` on the second call, with `replace === true` on the first as an absolute baseline, so a regression that deleted `replace` everywhere cannot pass. Three existing assertions that poked at an internal boolean were replaced with wire-payload assertions for the same reason.

## Test-quality notes

Three sibling `JobsPage` tests mocked the session store with a plain object, which cannot model `getState()` or propagate `setJobs`; they now use the real store seeded per test, and two spy assertions became state assertions. `session-store.test.ts` compared the whole slice to a literal; it now spreads the defaults **plus** six absolute assertions, so a default silently flipping cannot ride along.

## Note

Identity checks inside the event handler read `useSessionStore.getState()` synchronously — the handler runs on backend events, not on React's schedule, so a subscription-captured value would reintroduce exactly the staleness the old ref existed to prevent. The `isActiveJob`-before-`noteScrapeFinished` ordering is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Scraping progress and results now persist when navigating away from and returning to the Jobs page.
  * Active jobs remain cancellable, while completed jobs and per-board summaries are restored automatically.
  * Search changes correctly append or replace results as appropriate.
  * Scrape failures now display clearer, board-specific diagnostics.

* **Bug Fixes**
  * Prevented stale or unrelated job updates from overwriting current results or error details.
  * Improved recovery and reconciliation of jobs after navigation or remounting.

### DIFF
--- a/apps/desktop/src/renderer/features/jobs/components/JobsCommandBar/index.tsx
+++ b/apps/desktop/src/renderer/features/jobs/components/JobsCommandBar/index.tsx
@@ -70,8 +70,13 @@ export function JobsCommandBar({
   scrapeButtonRef,
 }: JobsCommandBarProps) {
   const { t } = useTranslation();
-  const { jobs, setJobs } = useSessionStore();
-  const { filter, sortBy, viewMode, hideAgency } = jobs;
+  // One selector per field (see JobsPage): an unselected `useSessionStore()`
+  // re-renders this bar on every mutation of every other slice.
+  const setJobs = useSessionStore((s) => s.setJobs);
+  const filter = useSessionStore((s) => s.jobs.filter);
+  const sortBy = useSessionStore((s) => s.jobs.sortBy);
+  const viewMode = useSessionStore((s) => s.jobs.viewMode);
+  const hideAgency = useSessionStore((s) => s.jobs.hideAgency);
 
   const trimmedFilter = filter.trim();
   const hasActiveFilter = trimmedFilter.length > 0;
--- a/apps/desktop/src/renderer/features/jobs/components/JobsPage/JobsPage.cluster-filter.test.tsx
+++ b/apps/desktop/src/renderer/features/jobs/components/JobsPage/JobsPage.cluster-filter.test.tsx
@@ -6,11 +6,12 @@
  *   - clusterCanonical === false rows are hidden (collapsed into the canonical).
  *   - Unannotated rows (no cluster fields) always show.
  *   - `hideAgency` (session state) removes isAgency rows.
- *   - The hide-agency toggle writes hideAgency via setJobs.
+ *   - The hide-agency toggle writes hideAgency into the session store.
  *
- * The session-store mock is a mutable container so hideAgency can be toggled
- * per test. All heavy dependencies are module-mocked. `mergePostings` is real —
- * fixtures use distinct urls/ids so nothing collapses at the stage-1 dedup.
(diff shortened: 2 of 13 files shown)


saeedkolivand:
@coderabbitai review

PR #1021, opened 2026-08-18
chore(deps): bump h2 to 0.4.16 for rustsec-2026-0258

saeedkolivand:
**Unblocks pushes on every branch.** A new advisory landed in the RUSTSEC database today and started failing `cargo deny check advisories` in the pre-push gate. Nothing in this repo changed — the advisory did.

## RUSTSEC-2026-0258

> The h2 crate, used internally by hyper, had a flaw that would accept and queue empty DATA frames without limit. If streams were not actively drained, this could lead to unbounded memory usage, or a panic if the length overflows.

Low severity. We reach it transitively through **both** `hyper 1.10.1` and `reqwest 0.13.4`, not directly.

## The change

Lockfile only. `0.4.15` → `0.4.16` is the patched release the advisory names, and `cargo update -p h2` locks it with 173 other dependencies unchanged.

```
cargo deny check advisories
  before: advisories FAILED
  after:  advisories ok
```

The pre-existing `quick-xml` finding under the typst family is untouched and still red by the owner's standing decision — this is a separate, newly-published advisory with an available patch.

### DIFF
--- a/apps/desktop/src-tauri/Cargo.lock
+++ b/apps/desktop/src-tauri/Cargo.lock
@@ -2182,7 +2182,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


saeedkolivand:
@coderabbitai review

PR #1027, opened 2026-08-18
feat(scraping): remember which boards actually work between runs

saeedkolivand:
Track B1 of the architecture audit. A board scrape currently forgets everything
the moment it ends: a source that has failed the last nine runs looks exactly
like one that has never been tried. This adds a small per-board reliability
store (`board_health.db`), derives a verdict from it, and surfaces that verdict
on the board chips and the autopilot card.

Five states: `Healthy`, `Failing`, `Flaky`, `Stale`, `Unknown`. `Flaky` is the
one that earns its keep, because it catches the pattern a consecutive-failure
counter structurally cannot see: a board that works often enough to never build
a streak while failing a meaningful share of the runs that reach it.

## What review changed, since that is most of the value here

Two independent architecture passes and a sibling audit, each verified by
executed mutation rather than by reading. Everything below was found, not
designed in.

**Two HIGHs, both closed and both re-proven by mutation:**

- Unbounded renderer-driven row creation. The health store is keyed on board id,
  and the id came from the renderer, so an unrecognised id could mint a row.
  Now filtered on the resolver (not the registry, which fails the fake-resolver
  tests), and the guard asserts `tracked_boards()` — a `COUNT(*)` — because the
  predecessor's `health_for(id).is_none()` shape could not see a row created
  under an id nobody thought of.
- The recovery path left `failing_since` and `last_error` on disk. A
  `COALESCE(board_health.failing_since, excluded.failing_since)` mutation, and
  four variants of it, all die against
  `a_recovery_clears_the_streak_columns_on_disk_not_just_in_memory`.

**Three MEDIUMs, measured rather than argued:**

- **The `Flaky` verdict used lifetime counters**, so it had inertia in both
  directions. A board that had a 10-run outage and then worked *perfectly* kept
  rendering `unreliable · failed 10 of N runs` for **31 flawless runs**; a board
  with 100 clean runs that started failing **every other run** stayed unbadged
  for **100 more**. Now a decayed rolling window: 16 runs and 14 runs/7 failures
  respectively.
- **The positional zip in the engine was correct but untestable.** Replacing it
  with `enumerate()` left the entire suite green, because the guard test put its
  only valid board at index 0. The discriminating shape is a rejected id ahead
  of two *failing* boards, and that test now exists.
- **`neverWorkedSince` rendered "has never worked here · failing for 2 days
  ago"**, because `timeAgo` returns a relative phrase. The existing chips test
  mocks `t()` down to `key:opts` and could never see it, so the new test calls
  the real unmocked `i18n.t`.

**One correctness bug in the fix for the first MEDIUM.** The decay rescales
`failed` proportionally rather than halving it independently, because
independent halving rounds `failed` down less than an odd `verified` and can
*manufacture* a Flaky verdict out of pure rounding (17/4 = 23.5% halves to
8/2 = 25.0%, on a plain success run). But the rescale was done in `u32`:
`failed.saturating_mul(next_verified)` overflows past ~4.29e9 and clamps
**before** the divide, so a board that failed 150,000 of 200,000 runs decayed to
2-in-12, read 16%, and came out Healthy. Fixed in `u64`.

The existing oversized-tally test could not catch that, because it asserts only
that the cap is respected — and the cap *is* respected while the ratio is being
destroyed. The replacement anchors on an absolute floor instead: a board failing
three runs in four must not decay to something better than a coin flip.

## Known and recorded, not fixed

- **`FLAKY_WINDOW_CAP` bounds runs, not time.** At the daily autopilot cadence
  that is one to two weeks, close to `STALE_AFTER_MS`. For someone scraping
  manually several times an afternoon, the same 16 runs span hours and the
  verdict forgets the morning by evening. Documented at the constant.
- **`keys.boards.health` is invalidated on every job event**, which during a
  watched-company scrape means once per company. It rides the invalidation
  cadence `keys.jobs.all` already had, and each one is a small local SQLite
  read, so this is an accepted cost rather than an oversight.
- **`updated_at` is written and never read.** Every question it could answer is
  answered more precisely by `last_success_at`, `last_verified_at` or
  `failing_since`, all of which `derive_status` already reads. It is redundant
  rather than unused; dropping it is a one-line migration and is not done here.

## Not verified

No visual check of the chips or the autopilot card.

## Summary by CodeRabbit

* **New Features**
  * Added cross-run board health indicators showing failing, stale, flaky, healthy, and unknown statuses.
  * Board summaries now display reliability details, failure counts, timing, and sanitized diagnostic tooltips.
  * Added live board-health updates after job activity.
  * Added factory-reset support to clear stored board health history.

* **Privacy**
  * Machine-specific health data is excluded from saved runs and exports.
  * Diagnostic messages now consistently redact sensitive URLs, paths, credentials, hosts, and email addresses.

* **Localization**
  * Added English and German translations for board health statuses.


saeedkolivand:
@coderabbitai review

PR #1029, opened 2026-08-18
fix(resume-pipeline): resolve a fabrication remove by its anchored line

saeedkolivand:
The renderer half of the Major finding CodeRabbit raised on #1024. The Rust half
shipped with that PR; this closes the TS side, which merged into `main` still
wrong.

## The defect

`fabrications.ts` was internally inconsistent about how it identifies an entry.

`removeEvidenceLines` is anchored on `entry.line`, by exact whitespace-trimmed
line equality, and its doc comment says outright **"never by searching for
`evidence`"** — with the reasoning spelled out: a short span matches lines it
was never about, so a substring search deletes the wrong thing.

But `evidencePresent`, which `isFabricationResolved`, `presentFabrication` and
`unresolvedCount` all decide on, did exactly what that doc forbids:

```ts
return !!span && documentText.includes(span);
```

So the app **deleted** by line anchor and **decided resolution** by bare
substring. The user removes the flagged line, the removal succeeds, and if the
span occurs anywhere else in the document the entry never reads resolved — the
integrity chip stays off green and the run stays `needsReview` over work that is
genuinely done. Short spans make that routine: a certification acronym that also
appears in a skills list, or a two-character language code sitting inside
ordinary words.

## The rule

A `remove` verdict now prefers the anchored `line` — resolved once that exact
trimmed line is no longer anywhere in the document — falling back to the
`evidence` substring only when the entry carries no `line`. That absence is a
**supported state**, not a bug: reports persisted before the field existed have
none.

Same field priority as Rust's `entry_resolved`, deliberately, because a
heuristic duplicated across Rust and TS that then drifts is its own recurring
defect class in this repo.

`evidencePresent` is unchanged and still drives the `pending`/`orphaned` split
for undecided entries — that is a rendering question ("can we show the flagged
span"), not a resolution question.

## The alternative, and why it was rejected

The obvious-looking stronger rule is to require **both** signals: the line is
gone AND the evidence is gone. That was mutation-tested rather than argued
about, and it flips the regression test back to red — ANDing reintroduces the
exact bug, because the bare span legitimately survives elsewhere, so
"evidence gone" is false forever and the entry sticks unresolved.

The residual gap is real and worth stating: an entry whose line is **reworded in
place** while still containing the fabricated span reads as resolved. There is
no honest anchor that survives an in-place edit without relocating the line by
searching for the bare span, which is precisely the operation
`removeEvidenceLines` forbids. Rust's shipped fix takes the same cost.

## Verification

Two mutations executed by hand, each watched red then green: reverting
`fabrications.ts` to its pre-fix `main` content (both new regression tests fail),
and the AND variant above.

`tsc --noEmit` clean, ESLint `--max-warnings 0` clean with no disables added,
and 951/951 across `lib/generate`, `components/generation` and `TailorFlow`.

Every caller checked for dependence on the old semantics:
`FabricationReview.tsx`, `QualityBadge.tsx`, `useTailorPipeline.ts`, and
`QualityReportPanel.pipeline.test.tsx` — none of whose fixtures exercised the
"span recurs elsewhere" case, which is why none of them caught this.

### DIFF
--- a/apps/desktop/src/renderer/lib/generate/fabrications.test.ts
+++ b/apps/desktop/src/renderer/lib/generate/fabrications.test.ts
@@ -109,6 +109,31 @@ describe('presentFabrication', () => {
   it('treats empty evidence as orphaned, not as matching everything', () => {
     expect(presentFabrication({ ...VALID, evidence: '   ' }, document)).toBe('orphaned');
   });
+
+  // Coherence with `isFabricationResolved`: for a `remove` verdict, this must
+  // return 'resolved' in EXACTLY the cases the other function calls resolved
+  // — same line-anchored bug scenario, checked from the presentation side.
+  describe('line-anchored Remove — coherent with isFabricationResolved', () => {
+    const anchored: Fabrication = {
+      issueKey: 'factual.unsourced_certification#0',
+      code: 'factual.unsourced_certification',
+      evidence: 'CKA',
+      line: 'Certified Kubernetes Administrator (CKA).',
+      decision: 'remove',
+    };
+
+    it('reads resolved once the anchored line is gone, span or no span elsewhere', () => {
+      const edited = 'PROFESSIONAL SUMMARY\nA payments engineer.\n\nPROJECTS\n- CKA prep tool.\n';
+      expect(presentFabrication(anchored, edited)).toBe('resolved');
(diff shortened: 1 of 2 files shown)


saeedkolivand:
@coderabbitai review

PR #1031, opened 2026-08-18
refactor(landing): serve the /world clips from a release, not from the repo

saeedkolivand:
Moves 33 files / 62 MB of `/world` background video out of the tree and onto the
[`world-assets-v1`](https://github.com/saeedkolivand/ai-job-hunter-app/releases/tag/world-assets-v1)
release. Two separate problems, one move.

## Problem 1: git weight, and specifically the *rate*

Video does not delta-compress, so a re-encode writes every byte again rather
than a diff. One re-encode round had already put **~38 MB of unreclaimable
history** in the pack. The 62 MB level was survivable; the growth rate was not.

## Problem 2: a Pages bandwidth ceiling nobody had noticed

This is the one that actually justifies the change. Sitting in
`apps/landing/public/`, the clips were built into the published Pages site — so
every view drew on GitHub Pages' **100 GB/month soft bandwidth limit**. At this
route's 20 MB device-set cap that is **roughly 5,000 visits a month** before
HTTP 429s.

Release assets are a different pool. From GitHub's docs, verbatim: *"There is no
limit on the total size of a release, nor bandwidth usage."*

To be precise about what changed: GitHub still serves the bytes. They just no
longer count against the metered Pages pool.

## What stays

The **poster stills stay in the repo and on Pages** — ~2 MB total, and they must
paint before any video is fetched. Only the clips move.

## Shape of the change

Every clip URL now goes through a single `vid()` helper pinned to a release
**tag**. Pinning is deliberate: publishing `world-assets-v2` and bumping one line
is atomic, whereas mutating an existing release's assets in place would leave
cached clients disagreeing with fresh ones.

`NEXT_PUBLIC_WORLD_VIDEO_BASE=/world/vid` restores local files for offline work,
and the directory is gitignored so a local copy cannot be re-committed by
accident.

## Tests

The two tests asserting every asset path starts with `/world/` **split in two**,
because that invariant now holds for the posters and not for the clips.

Both halves assert against **literals**, not against the config's own base — so a
regression cannot move the expectation along with the value it is checking. A
fourth test pins every clip to a single tag, so a partial bump cannot ship.

**Mutation-checked by execution:** pointing the base back at `/world/vid` turns
two of them red; restoring turns them green. 34/34.

## Verified

- `next build` clean; the built output contains the release URL and **zero
  remaining `/world/vid/` references**
- published site **~74 MB → 12 MB**
- `tsc --noEmit`, prettier, and `check:landing-drift` all green
- a real `curl` of a release asset returns 302 → 200

## Not verified

`/world` has not been opened in a browser against the release URLs. The build
output and the asset fetch are both checked, but the scroll route itself is not
visually confirmed.


saeedkolivand:
@coderabbitai review

Issue #1033, opened 2026-08-18
Rewrite history to drop the /world video blobs (147 MB clone → ~47 MB)


saeedkolivand:
Follow-up to #1030, whose precondition is now met.

## The numbers, measured today

| | |
|---|---|
| Video files at HEAD | **0** — removed by #1031 |
| Video blob entries still in history | 55 |
| Packed size of those blobs | **99.9 MB** |
| Total pack | **147.04 MiB** |
| Pack after a rewrite | **~47 MB** |

**A 68% cut.** Worth doing, unlike file-deletion cleanups that reclaim a
fraction of a percent — removing a 1.5 MB audit report earlier reclaimed 0.4% of
blob weight and changed clone size by exactly nothing.

Critically, the assets are out of HEAD and now served from R2, so **nothing will
put them back.** A rewrite done before that would have been undone by the next
re-encode.

## The command

```bash
pip install git-filter-repo
git filter-repo --path apps/landing/public/world/vid --invert-paths
git reflog expire --expire=now --all && git gc --prune=now --aggressive
git push --force --all && git push --force --tags
```

## Costs, all real

- **Every commit sha from the assets' introduction onward changes.** Every sha
  cited in ADRs, PR bodies, issue comments and commit messages goes stale.
- **Open PRs break** and have to be rebuilt. Do this with zero open PRs.
- **Everyone re-clones.** No exceptions; a `git pull` onto rewritten history
  will not do the right thing.
- **GitHub keeps the old objects until GitHub Support runs `gc` for you.** Until
  that ticket is filed and actioned, the space is not actually reclaimed
  server-side and the old shas remain reachable. The rewrite alone does not
  remove the data from GitHub.
- 241 tags need force-pushing.

## Do it when

1. **No open PRs.** This is the real scheduler.
2. Announce first if anyone else has a clone.
3. File the GitHub Support request immediately after, or the headline number
   never materialises for anyone cloning fresh.

## Worth stating: this is optional

CI already uses `fetch-depth: 1` everywhere, so it never fetches history.
Anyone who finds a full clone slow can use `git clone --filter=blob:none <url>`
today with no coordination and no rewrite.

The case for doing it is that **147 MB is a bounce factor for a public repo** —
a first-time visitor judges the clone size before they judge the code. That is a
real reason, but it is a presentation reason, not a correctness one, and it
should be weighed against invalidating every sha reference in the project's
written history.

saeedkolivand:
[...]

PR #1034, opened 2026-08-18
docs: repoint the audit-report sha after the history rewrite

saeedkolivand:
Cleanup after the history rewrite in #1033.

The rewrite renamed every commit, so the `git show <sha>:AUDIT_REPORT.md`
pointer that ADR-0005 and ADR-0006 carry went dead the moment it landed.

Translated through filter-repo's `commit-map`: **`65ea4188c` → `cd1219ef7`**,
verified to still hold all 1,588,368 bytes.

Both citations now also record that the sha **has already moved once** and name
PR #1000 as the durable anchor. A bare sha is exactly the kind of reference a
rewrite breaks; a reader who hits a dead one shouldn't have to guess what to do.

## Known and deliberately not fixed

`CHANGELOG.md` carries a commit link for every release — **all of them now point
at pre-rewrite shas.** That file is generated by semantic-release across 241
tags, so hand-editing it is neither feasible nor durable.

Those links keep resolving today because GitHub doesn't garbage collect on its
own and this repo has forks sharing the object store. They break if that ever
changes. Accepted: the alternative is rewriting generated release history to
chase link cosmetics.


saeedkolivand:

Issue #1026, opened 2026-08-18
Thanks for #1008 — your notes led to real fixes on the freehire side (docs, User-Agent, rate limits)


strelov1:
Hi — I maintain [freehire](https://freehire.me) (the keyless tier added in #1008).

I read that PR carefully, and I want to say thank you properly: the "Known limits" section did more for us than a bug report would have. Three of the things listed there as constraints of our API were actually **defects in our published spec**, and I would not have found them from the inside. All three are fixed and live now.

---

## 1. City filtering exists. Our spec just never mentioned it.

The PR notes:

> **No city filtering.** The documented search has `countries` and facet parameters but no city or free-text location parameter.

That was a correct reading of `openapi.yaml` and a wrong picture of the API. `cities` has always been a real facet:

```
GET /api/v1/jobs/search?cities=London   →  ~43k postings
```

It simply was not in the schema. Auditing the rest, our published spec described **17 of ~35 query parameters**. Also missing: `posted_within_days` (which is why your `date_filter` argument had nowhere to go), `reality`, `is_tech`, `role`, `collections`, `visa_sponsorship`, and the whole `<facet>_exclude` / `<facet>_mode=and` grammar.

Two parameters we *did* document turned out not to exist at all — `semantic_ratio` was removed with an old index, and `/jobs/{slug}/similar` advertised `max: 50, default: 10` when the real values are 20 and 6.

Everything is now documented and, more importantly, **verified against production before being written down** rather than transcribed from memory.

### One thing worth knowing if you ever do add city support

Geography is a single **OR-group**, which is genuinely surprising:

```
countries=gb                  →  89,211
countries=gb&cities=London    →  89,617   ← wider, not narrower
```

`regions`, `countries` and `cities` name one concept — *where* — so they widen each other instead of intersecting. Naive "add a city to narrow the country" makes results grow. That is now stated in the schema description and again on all three parameters. There is also `GET /api/v1/geo/cities?q=lond` to resolve a name to the canonical facet value, since a near miss silently matches nothing.

## 2. The User-Agent point was fair, and we now ask

You raised this and then declined to decide it, which was the right call — it was never your decision to make alone:

> The request carries the shared default user-agent, so the traffic is not attributable to freehire's own integration. Attribution versus fingerprinting is a product decision, not a defect.

We had never asked anyone to identify themselves, so there was nothing for you to comply with. Now `openapi.yaml`, `llms.txt` and `robots.txt` all request a format:

```
[...]

PR #1035, opened 2026-08-18
fix(scraping): wire the freehire params their spec never documented

saeedkolivand:
Acts on issue #1026, where the freehire maintainer disclosed that our PR #1008
"Known limits" section had found three defects in **their** published spec —
`openapi.yaml` documented 17 of ~35 real query parameters.

## The actual bug

`freehire.rs` took `_date_filter: Option<&str>` — underscore-prefixed, accepted
and thrown away, because the documented spec had no date parameter. It has one:
`posted_within_days`. Our date filter had nowhere to go and silently did
nothing.

## Verified against the live API, not taken on trust

Every claim in the issue was checked by hitting `/api/v1/agent/jobs/search`
directly and re-reading the live `openapi.yaml`:

| Claim | Result |
|---|---|
| `posted_within_days` exists | ✅ integer days, `minimum: 1`; `=1` cut 1,477,440 → 66,802 |
| `reality=fresh` exists | ✅ cut the catalogue to 662,112 (~45%) |
| `meta.ignored_params` shape | ✅ a `country=gb` typo returns **200** + the **full unfiltered set**, with `did_you_mean: "countries"`. A clean request omits the key entirely — not an empty array |
| `X-RateLimit-*` on success | ✅ `300/299/1` on the agent endpoint |

**Not** independently verified: the 600/300 req/min ceilings and the leaky-bucket
behaviour under sustained load. Load-testing someone else's production service to
confirm a courtesy disclosure would be rude. Taken on the maintainer's word —
which is why **no rate-limiter code was written**; `fetch_text`'s existing
`429`/`Retry-After` backoff already covers the requirement.

## A pre-existing latent bug found on the way

`fetch_text` wrote `user-agent: DEFAULT_UA` unconditionally and *then* applied
caller headers via `RequestBuilder::header`, which **appends rather than
replaces**. Any caller passing a `user-agent` would have sent **two header
lines**. Nothing did yet, so it had never fired.

Fixed properly with a `FetchOptions::user_agent` override — `None` leaves every
other board on the unchanged `DEFAULT_UA`, which matters because several boards
depend on looking like an ordinary browser. All ~15 `FetchOptions` construction
sites were checked for `..Default::default()` before adding the field.

## Two judgement calls

**`reality=fresh` is unconditional, with no setting.** Every other aggregator
quality lever here (`sort_by=date`, `dedupe_by_url`) is already always-on with no
user-facing knob; a bespoke toggle for one facet on one last-resort tier would be
the odd one out.

**`meta.ignored_params` is treated as an error, not a warning that still returns
data.** It reuses the existing silent-degradation boundary rather than inventing
a signalling path. This is the finding that worried me most: a mistyped filter
returns the whole catalogue and **looks exactly like a legitimate broad result**.
Silently presenting unfiltered results as filtered is the one outcome that is not
acceptable.

## Mutations executed

Five, each broken → run → confirmed red → restored → re-run green:
the `posted_within_days` floor, the identifying UA, the `fetch_text`
replace-not-append fix, the `ignored_params` guard, and `reality=fresh`.

## Out of scope

**`cities` is deliberately not added.** It is real, but geography is a single
**OR-group**: `countries=gb` → 89,211 while `countries=gb&cities=London` →
**89,617**. Adding a city to narrow a country makes results *grow*. That needs a
location-UI decision, not a parameter.

## Provenance note

These two commits were originally authored before today's history rewrite, lost
when I pruned unpushed local branches ahead of it, and recovered from the backup
bundle as patches replayed onto rewritten `main` — so the shas differ from the
originals. Re-verified after replay: 17 freehire tests, 41 http tests, clippy
clean with `-D warnings`.

### DIFF
--- a/apps/desktop/src-tauri/src/scraping/boards/aggregator/freehire.rs
+++ b/apps/desktop/src-tauri/src/scraping/boards/aggregator/freehire.rs
@@ -31,14 +31,35 @@
 //!   MERGED behind the sparse keyed hits rather than replacing them, because
 //!   on a guessed market this tier is location-blind and those hits are not.
 //!
-//! **Degradation:** any non-2xx, timeout, or schema drift resolves to `Ok(empty)`,
-//! NOT `Err`. Every other provider reports its failure because a configured key
-//! means the user asked for that provider specifically. freehire is configured
-//! by nobody — it is always on — so surfacing its outage as a board error would
-//! turn a third party's downtime into an error banner on a search the user never
-//! pointed at them. There is no SLA, no rate-limit header of any kind, and no
-//! documented limit (verified: no `X-RateLimit-*`, no `Retry-After`,
-//! Cloudflare-fronted), so it must be treated as always-possibly-absent.
+//! **Degradation:** any non-2xx, timeout, schema drift, or ignored request
+//! parameter (see `fetch_freehire`'s doc) resolves to `Ok(empty)`, NOT `Err`.
+//! Every other provider reports its failure because a configured key means the
+//! user asked for that provider specifically. freehire is configured by
+//! nobody — it is always on — so surfacing its outage as a board error would
+//! turn a third party's downtime into an error banner on a search the user
+//! never pointed at them.
+//!
+//! **Rate limits (as of the maintainer's issue #1026 disclosure, live-verified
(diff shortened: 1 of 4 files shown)


saeedkolivand:

PR #1036, opened 2026-08-18
fix: make in-flight backend work survive navigation, and stop logs leaking paths

saeedkolivand:
## What this is

An audit of the nine route-scoped entries in `check-event-subscriptions.mjs` found **eight of the nine notes were wrong**. Fixing them turned up six real defects behind them, so this PR carries both: the corrections, the fixes, and two guards so neither class rots again.

## The notes were wrong, and nothing could tell us

Three notes described a loss that had already been fixed, four understated it, one named a cause the code does not have. Only the activity-feed note was accurate.

They rot because nothing coupled them to the code: #1019 fixed the jobs scrape and never touched the note beside it, and #1025 later edited that very file without noticing the note next to it was by then obsolete. The guard enforced that a note **exists**, never that it is **true**.

Each route-scoped entry now carries a content hash of the file it describes, and the check fails when the file drifts, printing the note so re-reading costs seconds. It fired on this PR's own merge for four files — that is the mechanism working.

**Known blind spot, recorded in the file:** the hash covers the declared file only, so a note also goes stale when behaviour changes in something that file merely *calls*. Both misses came from this batch — `JobsPage`'s defect was fixed inside `useScraping.ts`, `update-section`'s inside the updater service — and neither declared file was touched. Hashing the import graph would fire on nearly every commit and be ignored within a week, so those two entries say in their own text that they were fixed elsewhere.

## The six defects

**You get invited to start expensive work that is already running.** Returning to Settings mid-update rendered a live "Check now"; clicking it cleared `downloaded_bytes`, discarding a finished download and forcing a full re-fetch, with no re-entry guard. Returning to onboarding mid-pull re-showed Download; clicking it started a *second* multi-GB pull, because `ai_pull_model` used `job_start` rather than the `job_start_exclusive` that already existed.

**The failure reason never arrived.** A résumé run that failed while you were on another tab left no error at all — you came back to the plain wizard as if you had never pressed Generate, because `error` was only ever written by the live listener. Autopilot had the same shape, plus a `staleTime` hole that showed an idle card with an enabled Run button for a live run.

**Jobs lied about progress.** The default board set is one board, so `scrape:progress` fires once, at the end — after a remount the bar asserted a false 0% for the rest of the run, while the tracker already held the true fraction and the watchdog fetched it and threw it away. A scrape that *failed* off-page came back with no explanation at all.

## Path leaks in logs

145 `log::…("{e}")` sites audited against rule 1 of `AGENTS.md`. **100 fixed**, the rest traced and declared safe with a reason. Not a blanket replacement — 33 of the scraping sites turned out genuinely safe because `scraping::http::fetch_text` already calls `.without_url()`, and replacing them would have discarded useful diagnostics to fix nothing.

The one worth naming: `linkedin/api_client/mod.rs` bypasses that chokepoint with a raw `client.get(url).send()`, so a transport failure logged the full guest-search URL **including the user's free-text keywords and location**.

`scripts/check-log-error-leaks.mjs` now fails in both directions — an undeclared leak, and a declared entry that no longer matches.

## Verification

- `cargo test --all-features --locked` — 4,349 passed
- `pnpm test` — 7,102 passed · `pnpm typecheck` · `pnpm lint:strict --max-warnings 0` — clean
- both new guards plus `check:agent-system` green

Every fix was mutation-checked by execution, not by reasoning: the guard removed, the test watched go red, the guard restored.

Two things I checked rather than assumed. Two Rust tests failed once under full parallelism and passed on a clean re-run; neither file is in this diff, so they are pre-existing flakes, confirmed by re-running rather than by asserting it. And a new leak test dialled the discard port, which Windows leaves hanging until its own timeout — over a minute on every run, now bounded to 0.52s.

## Note for review

`extension_bridge/mod.rs` crossed the R8 LOC cap once its fixes landed, so its token/opt-in persistence split into `persist.rs`, following that file's own existing `autotrack.rs` precedent.

## Summary by CodeRabbit

- **New Features**
  - AI model downloads and software updates now resume reliably after navigation or remounting.
  - Scraping progress, completion status, and failure details are restored more consistently.
  - Autopilot and résumé pipeline errors persist across page changes.
  - Updater status remains synchronized across multiple views.
  - Extension pairing tokens are stored with stronger file protection.

- **Bug Fixes**
  - Prevented duplicate AI and update downloads.
  - Improved refresh behavior for autopilot data and completed scraping jobs.
  - Sensitive details are now redacted from diagnostic logs.

- **Chores**
  - Added automated checks to prevent unsafe error details from entering logs.

### DIFF
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -339,10 +339,22 @@ jobs:
           pnpm check:adr-citations
           echo "::endgroup::"
 
-      - name: ❌ Check for Failures
-        if: steps.eslint.outcome == 'failure' || steps.format.outcome == 'failure' || steps.gen_ipc.outcome == 'failure' || steps.gen_prompts.outcome == 'failure' || steps.landing_drift.outcome == 'failure' || steps.agent_system.outcome == 'failure' || steps.tech_radar.outcome == 'failure' || steps.toolchain_pin.outcome == 'failure' || steps.event_subscriptions.outcome == 'failure' || steps.adr_citations.outcome == 'failure'
+      # A bare `{e}` inside a log::{warn,error,info,debug}! call can leak an
+      # absolute path (rusqlite::Error::InvalidPath, a filesystem
+      # std::io::Error), a credential-bearing URL (reqwest::Error), or a
+      # host:port into the log — see scripts/check-log-error-leaks.mjs.
+      - name: 🔒 Check log error-message leaks
+        id: log_error_leaks
+        continue-on-error: true
         run: |
-          echo "::error file=package.json::Lint, format, IPC codegen, prompts codegen, landing-diagram, agent-system, tech-radar, toolchain-pin, event-subscription, or ADR-citation drift check failed. Run 'pnpm lint:fix', 'pnpm format', 'pnpm gen:ipc', 'pnpm gen:prompts', 'pnpm check:landing-drift', 'pnpm check:agent-system', 'pnpm check:tech-radar', 'pnpm check:toolchain-pin', 'pnpm check:event-subscriptions', and 'pnpm check:adr-citations' locally to fix."
+          echo "::group::Checking for path/URL/credential leaks in log:: error interpolation"
+          pnpm check:log-error-leaks
+          echo "::endgroup::"
+
+      - name: ❌ Check for Failures
+        if: steps.eslint.outcome == 'failure' || steps.format.outcome == 'failure' || steps.gen_ipc.outcome == 'failure' || steps.gen_prompts.outcome == 'failure' || steps.landing_drift.outcome == 'failure' || steps.agent_system.outcome == 'failure' || steps.tech_radar.outcome == 'failure' || steps.toolchain_pin.outcome == 'failure' || steps.event_subscriptions.outcome == 'failure' || steps.adr_citations.outcome == 'failure' || steps.log_error_leaks.outcome == 'failure'
+        run: |
+          echo "::error file=package.json::Lint, format, IPC codegen, prompts codegen, landing-diagram, agent-system, tech-radar, toolchain-pin, event-subscription, ADR-citation, or log-error-leak drift check failed. Run 'pnpm lint:fix', 'pnpm format', 'pnpm gen:ipc', 'pnpm gen:prompts', 'pnpm check:landing-drift', 'pnpm check:agent-system', 'pnpm check:tech-radar', 'pnpm check:toolchain-pin', 'pnpm check:event-subscriptions', 'pnpm check:adr-citations', and 'pnpm check:log-error-leaks' locally to fix."
           echo "::error::Please fix the issues above before merging."
           exit 1
 
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -289,6 +289,19 @@ jobs:
       - name: 🦀 cargo check (all targets)
         working-directory: apps/desktop/src-tauri
         run: cargo check --all-targets --locked
+      # This job is `check` only (see the comment above), so `std::fs::rename`
+      # over an EXISTING file (postings::InteractionStore::save's write-then-
+      # rename) was never actually EXECUTED on Windows — rename-over-existing
+      # has different underlying semantics per platform (MoveFileExW needs
+      # MOVEFILE_REPLACE_EXISTING; POSIX rename(2) replaces unconditionally),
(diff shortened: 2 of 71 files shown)


saeedkolivand:
All 8 findings addressed. Two of them I pushed back on first and was wrong about one, so both are written up honestly below.

**1 — `job_fail` leaks through IPC.** Fixed, and this was the best finding here. The PR sanitized ~100 `log::` sites and left the same data flowing raw through `jobs_get`, `jobs_list` and `job.failed`. Both `job_fail` and `job_fail_with_data` now route through one mutator that sanitizes once and uses the same value for the tracker field and the event payload, so the two cannot drift. `data: Value` is untouched — it is a structured payload, not a message. Verified `timeout_message`'s plain-English output passes through byte-for-byte unchanged.

**2 — `{job_id}` can log a URL.** I initially disputed this: production boards build `id: format!("{}:{}", board, external_id)`, and the only `id: url` in the tree is a test fixture. **That check was incomplete and I was wrong.** `breezy`, `pinpoint` and `themuse` build `id: format!("{BOARD_ID}:{url}")` — the URL is a substring, not the whole id, which is why grepping for `id: url` missed it. The premise is false as literally worded and true in substance for three boards.

Fixed by shape-based redaction rather than hashing, so the ~21 boards with opaque ids stay debuggable and only URL-bearing ones collapse. `match_resume.rs:217` left alone deliberately: it echoes the caller's own `req.job_id` back to the renderer that supplied it, so no information crosses a trust boundary.

**3 — rejected origins.** Fixed, but not with `sanitize_reason`. That redacts credential-*shaped* tokens, and this log line is deliberately actionable — a developer needs the origin verbatim to add it to `extensionDevOrigins`. Control characters stripped and length capped at 128 instead. Worth noting the control-char strip is not exploitable through today's call site, since the value already passed `HeaderValue::to_str()`; it is caller-independent defence. The length cap is the live concern.

**4 — token file permissions.** Fixed with `OpenOptions::mode(0o600)` at creation, so there is no window, plus the trailing `set_permissions` kept to self-heal a pre-existing file, and its failure is now logged and propagated rather than discarded. This predates the PR — #1036 only moved the code into `persist.rs` for the LOC cap.

Honest limitation, recorded in the test module: the shipped tests **cannot** catch a reversion. The old code reaches the identical end state, and a purpose-built repro — 2,000,000 tight-loop `stat` iterations under a fully permissive umask — never once observed the window. It is real (two syscalls) but too short for a black-box test to land inside. The tests guard end-state correctness and a refactor that drops `set_permissions` entirely.

**5 — Windows atomic replace.** Added one narrow step running that single test. Not the whole suite. It covers rename-over-existing, no leftover `.tmp`, and a round-trip through a fresh store; it does not cover other write-then-rename sites or the permission-denied branch.

**6 — reattach reconciliation.** One of the two paths is real and one is not.

The event race is real and now reproduced as a failing test: `useJobQueue` resolves via a real IPC round trip, and the command writes the registry *before* emitting its single terminal event, so an event landing in that window compares against `null` and is dropped forever. Fixed by reconciling the job's status right after adoption.

The stale-`running`-row path is **refuted**: `JobTracker::open()` marks every interrupted job `failed` on startup, with an existing test covering it, so a killed pull can never present as running.

Investigating it also turned up a second half — the id has to be committed to a ref *synchronously*, because a resolved IPC promise unwinds through microtasks before React gives the render body a turn. Each half has its own mutation check.

**7 — hashless entries bypass the tripwire.** Fixed; a route-scoped entry now requires a valid 12-hex hash, checked before staleness runs. This was the guard I added in this PR, and it shipped with a one-missing-key opt-out.

**8 — scanner only matched `{e}`.** Fixed. Rather than guessing which bindings are errors, every one of the crate's ~264 log sites was enumerated: `e` at 45, `err` at 2, and no other captured identifier is ever an error. Now catches captured `{e}`/`{err}`/`{error}` and bare positional arguments; `.code()` and `.kind()` stay excluded. Three newly-surfaced sites, each traced and declared safe rather than blanket-replaced.

One guard added that was not requested. The new `autopilot_helpers` exemption is sound only because every board reaches the HTTP chokepoint, which holds only while no board is browser-mode — a three-hop argument whose last hop nothing enforced. The first browser-mode scraper would have silently re-opened a URL leak already signed off as impossible. That is now a registry assertion, mutation-checked by inverting the predicate.

Suite after all of it: 4,358 Rust, 7,128 JS, all three guards green.

saeedkolivand:

PR #1037, opened 2026-08-18
docs: generate the ipc api reference from the contracts

saeedkolivand:
## Why

`docs/API.md` was 1,095 hand-written lines, and a third of it was wrong.

It documented **three namespaces that do not exist** — `agent`, `search`, `shortcuts` — the `agent` section alone running ~76 lines of methods and events for a namespace with no contract. It **omitted nine real ones**. And its header read *Last updated: 2026-08-17*, which made it worse: the date told readers it was current.

The fiction went deeper than the missing namespaces. `credentials` was documented with four methods when the real contract has one. `dialog` with three, real answer one. `jobs` with nine, real answer five. `documents.import` and `ai.setProviderKey` had the wrong signatures entirely. An event, `ai:pull-progress`, appears nowhere in the repo — pull progress actually rides `job.stream` on the generic job channel.

## What changed

`scripts/gen-api-docs.mjs` emits the whole page from `packages/shared/src/ipc/contracts/`, wired as `gen:api` / `gen:api:check` alongside the three generators that already exist, and gated in CI's `lint-format` job.

Two design choices worth calling out:

**Signatures are copied verbatim from source text**, never re-rendered from a parsed type. Nothing can come out plausibly-wrong, which is the failure this page existed to demonstrate. Multi-line params and their inline field docs survive intact.

**It fails loudly rather than guessing.** An unsupported member kind, a non-literal channel entry, a missing interface — all exit 1 with a message saying to extend the generator. Proven by mutation: adding an index signature produces `unsupported contract member kind IndexSignature … extend the generator rather than guessing at its shape`.

## Prose was migrated, not deleted

The old page held real knowledge no signature carries — that `temperature` overrides `intent` on every adapter, that `embed`'s `model` field is on the wire but ignored, that `save` merge-upserts on `jobUrl`. Thirteen such facts moved into TSDoc on the owning contract method, each re-verified against the implementation first rather than copied on faith.

Deliberately dropped: the three fictional namespaces, and a hand-maintained 23-item board list that had already drifted — it was missing `jobicy` **while sitting next to a pointer at the canonical constant**. The pointer stayed.

107 of 174 methods now carry TSDoc. The other 67 render as signature-only entries. That is honest: the gaps are visible where a developer can fix them, instead of covered by prose that may or may not still be true.

## The gate provably fails

Adding a method to a contract and running `gen:api:check`:

```
Generated docs/API.md — 31 namespaces, 175 methods, 108 with TSDoc (67 without).
 | [`geocode`](#geocode) | 1 → 2
+#### `geocode.gateMutationProbe`
[ELIFECYCLE] Command failed with exit code 1.
```

Reverted, exit 0. Verified twice, independently.

Also worth recording: the generator's first version rendered an empty `*_CHANNELS` as "this namespace exposes no invoke channels" — **false** for `notifications`, which has nine live commands invoking their Tauri names directly. That is exactly the plausible-but-wrong statement this work exists to eliminate, so the last commit replaces the inference with the structural fact.

## Verification

`pnpm test` 7,136 passed · `pnpm typecheck` 13/13 · `pnpm lint:strict --max-warnings 0` clean · `prettier --check docs/API.md` clean (the generator formats its own output through prettier's API, so the check can never fight the formatter) · `check:adr-citations` · `check:agent-system`.

`AGENTS.md` rule 14 now sends a new capability to `pnpm gen:api` and says prose belongs in contract TSDoc, so the next person doesn't hand-edit a generated file and lose it.

## Summary by CodeRabbit

* **Documentation**
  * Expanded IPC API documentation with clearer behavior, limitations, privacy requirements, scoring options, data handling, and integration details.
  * API documentation is now generated consistently from documented contracts, reducing manual maintenance and improving accuracy.

* **Chores**
  * Added commands to generate API documentation and check for outdated generated content.
  * Continuous integration now detects documentation drift and reports required updates.


saeedkolivand:
All six real findings fixed. Two of the eight were the two halves of one finding, and one had a false premise with a correct remedy — details below, since the distinctions matter.

**Duplicate exported names.** This was the important one: `decls` and `consts` were keyed by bare name across every contract file, so two files exporting the same name meant the last one won and another file's interface could render under the wrong namespace. That is exactly the plausibly-but-wrongly output this generator's header claims to prevent, sitting inside the code that claims it. Now fails loudly rather than resolving silently — a real name clash is something a human should see, not something a generator should paper over.

**The staleness gate.** The reported wording was off by one step, and the correction is worth recording: a *staged stale* file already failed the old gate, because regenerating overwrites the worktree and worktree ≠ index. The actual hole is **stale at HEAD, fresh in the index** — regenerate, `git add`, do not commit. Proved by execution: old gate exit 0 while HEAD was stale, new gate exit 1.

`gen:workflows:check` carried the identical hole — that is where the shape was copied from — and got the same one-word fix. Leaving a known-vacuous gate beside a fixed one is not defensible.

**Absolute paths on failure.** Every `fail()` escaped to Node's default handler and printed a stack trace with drive-letter paths, violating `AGENTS.md`'s first rule in a repo that spent the previous day sweeping ~100 such leaks out of logs. Node's own filesystem errors carry an absolute path in the message too, not just the trace, so the repo root is stripped from those as well. Verified: `ENOENT … scandir '.\packages\shared\src\ipc\contracts'` — relative, no drive letter.

**Escaping.** CodeQL and the review flagged the same line independently, which is worth weighting. `\` is now escaped before `|`; a probe TSDoc containing `\|` previously split into four table cells and now renders as three, with the cell reading back byte-identical to the source prose.

**Extensionless specifiers.** `linkedin.ts` imports `./boards`, which emitted a link to a file that does not exist. Fixed, and default/namespace imports now fail loudly instead of being skipped.

**The channel arithmetic sentence** now only fires when channels < methods. It fired on any inequality, so a registry also carrying event names published a count larger than the total. Noting the irony rather than hiding it: that sentence was itself added to replace an inference that was plausible and wrong.

**On `geocode.ts`** — the finding says the TSDoc points at a directory that does not exist. That is not correct: `apps/desktop/src-tauri/src/commands/geocoding/` exists and holds `geonames.rs` and `test.rs`. The remedy is still right for a different reason — `to_city_country` is defined in the sibling *file* `geocoding.rs:65` — so the pointer now names the file that contains the function.

Since that prose came from a migration undertaken to remove fiction, every path introduced by it was re-verified against the symbol claimed, not just against the file: `sampling_profile` at `ai_provider/mod.rs:714`, `merge_application` at `ai_generations/mod.rs:926`, `SCRAPERS` at `scraping/boards/mod.rs:63`, `semanticScoringEnabled` at `schemas/index.ts:280`, and the four cited ADRs. `match_resume_batch` has zero occurrences outside the sentence saying it was removed.

**Deliberately not fixed**, with reasons: side-effect imports (`import './x'`) still skip silently — they carry no types to render, so failing on them would be a new rule rather than closing a leak. And #6 uses the count comparison that was asked for rather than a set difference, which would be strictly more truthful since equal counts do not prove the keys *are* the method names.

Verification: `pnpm test` 7,136 passed · `typecheck` 13/13 · `lint:strict --max-warnings 0` clean · `gen:api:check` and `gen:workflows:check` both exit 0 · `prettier --check docs/API.md` clean. Each guard change proved by execution against the pre-fix generator, so none of them passes vacuously.

saeedkolivand:

PR #1038, opened 2026-08-19
fix: stop a false language critical and read today as a date column

saeedkolivand:
Two defects in résumé validation, both of which made the app wrong about a
document while reporting nothing.

## 1. A date column spelled `Today` merged two employers into one

`is_date_only` matched present-tense markers by exact token equality against
`PRESENT_MARKERS`, which never carried English `Today` or its Romance
equivalents. An entry line ending in one failed to open its own role, so the
employer header **and all of its bullets** were absorbed into the entry above
it — one job's work credited to a different company.

Measured before the fix:

```
is_date_only("2020 - Present")     = true
is_date_only("2020 - Today")       = false   ← defect
is_date_only("2015 - Actualidad")  = false
is_date_only("2019 - Aujourd'hui") = false
```

Widening it surfaced eight more live spellings, including two where an
already-passing form sat one letter away from a failing one: `Present` passed
but `Presente` did not, `current` passed but `Currently` did not. English is
the default locale, so `2020 – Currently` was a defect on the primary path,
not only an i18n edge. Review found a further one — the list carried
Portuguese `atualmente` but not Spanish `actualmente`.

The spellings live in a **separate** `DATE_ONLY_MARKERS` list rather than
being added to `PRESENT_MARKERS`, and that separation is load-bearing.
`PRESENT_MARKERS` is matched as a single word anywhere in free text by
`is_open_ended` and by `validate::content::factual::unsupported_date_issues`;
putting `today` or `currently` there would turn an ordinary bullet into a date
context. `is_date_only` cannot make that mistake because it requires every
token on the line to be a digit, month or marker. A test now enforces the two
lists stay disjoint, so the invariant no longer depends on someone reading a
doc comment.

## 2. A truthful English résumé was flagged as the wrong language

`whatlang`'s `confidence()` is a **top-1 vs top-2 margin, not a probability**.
A noun-phrase-heavy block — a skills line, a terse CV — starves its n-gram
model of the function words it needs, so a truthful English document reads as
French at confidence `1.0000` with `is_reliable() == true`. That raised a
`content.language_mismatch` Critical, which blanks `keywordCoverage` and
suppresses every alignment finding on a correct document.

The fix requires positive evidence — function words distinctive to a language
other than the target — before a confident read becomes an accusation. The
rule is deliberately *positive evidence of a different language*, never
*absence of the target's*, because the latter fires on the terse English CV
again by a new route.

## Three designs were tried in this PR and taken back out

Recorded because the reasoning is the useful part. All three were caught by
mutation or by measurement, none by reading, and every one of them looked
correct and argued well.

**A globally-pruned evidence pool.** Pooling all seven word lists and dropping
any token appearing twice looked strictly safer — it can only make the guard
quieter. It killed Spanish. Every high-frequency Spanish function word
(`de`, `la`, `un`, `en`, `con`, `por`) collides with another list and was
pruned, leaving 30 survivors, none of which appear in a real CV. Four of five
realistic Spanish documents stopped raising a Critical that `main` raises
today, taking the draft-stage language retry down with them. Replaced with
**pairwise** evidence: a word only excludes when the language it is compared
*against* shares it.

**A short-token denylist and a comparative margin.** Both added as defence in
depth. Mutation showed no behavioural test depended on either — disabling each
failed only its own unit test. A sweep did find one input needing the denylist
(`au` repeated as disconnected French idioms with no connecting English
grammar), judged adversarial rather than representative and weighed against a
measured cost: denylisting `el` sat one hit from Spanish's margin. Deleted.

**A title-case-sandwich exclusion**, which ignored a function word sitting
between two capitalised neighbours on the theory that it was inside a proper
noun. **German capitalises every noun.** The standard German CV register is
nominal — "Aufbau und Betrieb der Zahlungsplattform" — so every connector sits
between two capitalised nouns and was discarded. A genuinely German document
dropped from 10 evidence hits to 1 and went silent, and with the flag false,
the English keyword-density check then ran over German text — the exact false
accusation its own doc says it exists to prevent. Deleted.

The common shape: each one encoded an assumption about what real text looks
like, and each assumption was wrong for some language. "Collisions remove
themselves" was wrong for Spanish. "Short tokens are ambiguous" was
unprovable. "Function words between capitals are inside proper nouns" was
wrong for German. What shipped instead assumes nothing about shape — only
quantity: a floor of 5 distinctive pairwise hits, derived from a corpus sweep
and mutation-checked in both directions (a floor of 4 reopens a false
positive, 7 reopens a false negative).

**This cannot introduce a false positive.** The new predicate is the old one
ANDed with the evidence requirement, so its fire set is a strict subset of
what `main` already reports. Everything below is a pre-existing gap that this
PR narrows, never a new one.

**Known limits, stated rather than papered over.** A count cannot fully
separate an unusually credentialed English CV from a short genuine foreign
paragraph. The boundary is language-dependent, and French is the worst case
rather than German — French official names are function-word dense
("Institut **de** Recherche **en** Informatique **et en** Automatique" is four
hits by itself), so two of them in an EDUCATION section reach the floor, where
German needs six. A realistic English CV with three German employers reads as
English with zero evidence and stays silent.

The floor also leaves a short genuine foreign paragraph silent against a
close-relative target — Portuguese against Spanish, Spanish against French —
where there is not room to accumulate five hits. Both limits are pinned by
tests as accepted misses. Given this file's history of three consecutive false
negatives, the floor is set to protect the true positive.

## Scope

`validate::content` is not called from `export/` (ADR-034). A Critical parks a
run at `needsReview`; it does not block an export. This changes when the
Critical is raised, not what it gates.

## Found here, fixed separately

Two defects confirmed by measurement during this work and deliberately left
out to keep the diff reviewable:

- `is_open_ended` returns true on ordinary prose containing `current`,
  `ongoing` or `actual` at a word boundary, so
  `factual::unsupported_date_issues` treats such a line as a date context.
  Measured: `is_open_ended("Reduced actual costs by 20% in 2023") == true`.
- `scraping/rate_limiter/mod.rs` computes `now - t` on `u64` wall-clock
  milliseconds. Wall-clock is not monotonic; a backwards step underflows —
  debug panics, release wraps and retains the entry forever. Reproduced twice
  under parallel load, not reproducible in isolation.

## Verification

`cargo fmt --all -- --check`, `cargo clippy --all-features --locked
--all-targets -- -D warnings`, and `cargo test --all-features --locked`
(without `--lib`, which skips `tests/*.rs`): 4381 lib + 18 architecture + 20
egress + 3 eval + 3 pipeline_kill_recovery, zero failures.

Every guard added here has a mutation check on record: delete the feature, not
the key, and watch the test go red. That includes the two guards that were
deleted for failing it.

## Summary by CodeRabbit

* **Improvements**
  * Improved recognition of date-only columns, including localized “Today” markers and additional date formats.
  * Preserved correct handling of employer names, dates, and bullet points when processing unparsed entries.
  * Improved language validation by requiring stronger linguistic evidence, reducing false mismatches from names, short phrases, and ambiguous text.
  * Applied consistent language checks across entire documents and individual sections.

### DIFF
--- a/apps/desktop/src-tauri/src/documents/evidence/entry.rs
+++ b/apps/desktop/src-tauri/src/documents/evidence/entry.rs
@@ -511,33 +511,70 @@ const MONTH_TOKENS: &[&str] = &[
     "dezember",
 ];
 
-/// "Today"-family present-tense markers recognised **only** by [`is_date_only`]
-/// — deliberately never added to [`PRESENT_MARKERS`], which [`is_open_ended`]
-/// and `validate::content::factual::unsupported_date_issues` both also consult.
+/// "Today"/"currently"-family present-tense markers recognised **only** by
+/// [`is_date_only`] — deliberately never added to [`PRESENT_MARKERS`], which
+/// [`is_open_ended`] and `validate::content::factual::unsupported_date_issues`
+/// both also consult, and never merged with either by stem/prefix matching.
 ///
-/// Measured, not just cautious: those two consumers match a SINGLE WORD
-/// anywhere in arbitrary text via [`contains_word`]/`contains_phrase` — no
-/// year, no separator, no other date structure required, because
-/// `is_open_ended`'s first branch returns as soon as the word is found.
-/// "today" is ordinary English vocabulary ("the system still runs today",
-/// "shipped the migration today"); putting it in [`PRESENT_MARKERS`] would
-/// turn any such bullet into a date context, and in
-/// `unsupported_date_issues` specifically, a false `factual.unsupported_date`
-/// Critical the moment that bullet also names a year the source résumé
-/// doesn't. (Confirmed the same failure mode is already live for `actual`,
-/// which is in [`PRESENT_MARKERS`] for Spanish `actual` — `is_open_ended`
-/// returns true for "Reduced actual costs by 20% in 2023" on the word alone.)
-///
+/// **Why a separate list at all — measured, not just cautious.** Those two
+/// consumers match a SINGLE WORD anywhere in arbitrary text via
+/// [`contains_word`]/`contains_phrase` — no year, no separator, no other date
+/// structure required, because `is_open_ended`'s first branch returns as soon
+/// as the word is found. Every spelling here is ordinary prose in its
+/// language, not just "today": "currently" ("we are currently migrating"),
+/// "derzeit", "actuellement", "presente" (also an ordinary Spanish/Portuguese/
+/// Italian adjective — "el problema presente"). Putting any of them in
+/// [`PRESENT_MARKERS`] would turn a matching bullet into a date context and,
+/// in `unsupported_date_issues` specifically, risk a false
+/// `factual.unsupported_date` Critical the moment that bullet also names a
+/// year the source résumé doesn't. (Confirmed the same failure mode is
+/// already live for `actual`, which IS in [`PRESENT_MARKERS`] for Spanish
+/// `actual` — `is_open_ended` returns true for "Reduced actual costs by 20%
+/// in 2023" on the word alone; see the handoff for the measurement.)
 /// [`is_date_only`] cannot make that mistake: it already requires EVERY token
 /// on the line to be a digit, a month or a marker, so a prose sentence
 /// carrying any other word is rejected before this list is ever consulted —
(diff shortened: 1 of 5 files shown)


saeedkolivand:

PR #1050, opened 2026-08-19
ci: group dependabot updates by area, and unblock local pushes

saeedkolivand:
Two build-tooling fixes. The second is unrelated but had to land first, because it blocked pushing anything at all from a Windows machine.

## 1. Group dependabot updates by the part of the app they touch

Ten open dependabot PRs at once. The breakdown was diagnostic: 5 GitHub Actions, 5 Cargo, 0 npm. npm already had a catch-all group and collapsed to a single PR; cargo and github-actions only grouped `typst` and `codeql-action`, so everything else opened individually.

Adds the same catch-all to both, named for the area so the PR title says which part of the app moved:

| Ecosystem | Groups |
| --- | --- |
| npm | `javascript` (renamed from `minor-and-patch`) |
| cargo | `typst`, then `desktop-rust` |
| github-actions | `codeql-action`, then `ci-actions` |

Majors stay individual on all three, unchanged.

Both new catch-alls carry `exclude-patterns` for the existing lockstep groups. `exclude-patterns` beats `patterns`, and group matching is first-match-wins by file order — but relying on order alone would let a `typst` bump land inside a 40-crate PR, where a partial bump that splits the exact pins is invisible. The exclusions make that impossible regardless of ordering.

npm stays one entry: workspace members share the root `pnpm-lock.yaml` and cannot be split per-directory.

Expected effect: ~3 PRs/week instead of 10.

## 2. Move the vitest browser server off a Windows-reserved port

Every push failed the pre-push hook with all 6956 tests passing:

```
Test Files  444 passed (444)
Tests       6956 passed (6956)
Error: listen EACCES: permission denied ::1:63315
```

Vitest hardcodes `defaultBrowserPort = 63315` and allocates upward from there. Windows reserves port blocks for Hyper-V/WSL in that neighbourhood (63285-63384 on the affected machine). A reserved port is still offered by the resolver but refused at bind, and vitest counts the unhandled error as a run failure.

Reproduced identically three times, sandboxed and not. Nothing in the repo calls `.listen()`; the binder is the Storybook vitest addon's preview server.

Pins `browser.api.port` to 6317, below the dynamic port range where these exclusions live. The default was already a fixed port, so this is not trading a dynamic allocation for a static one — it is the same shape at an address the OS will actually hand over.

Verified: `vitest run --project storybook` serves on `localhost:6317`, 46 files / 180 tests, exit 0. Full pre-push gate then passed end to end.


saeedkolivand:
🎉 This is included in version 0.138.1.

PR #1051, opened 2026-08-19
fix(ats): stop false date-context criticals, and route translation through the active provider

saeedkolivand:
Five threads folded into one PR, plus the pre-PR gate fixes they turned up.

**Read this first: the reported link bug is NOT fixed here.** A bare domain with no path (`aijobhunter.app`) still renders as plain text. The fix was built, reviewed, and reverted — see section 3. `model/rich.rs` has zero net diff against main.

## 1. A bare present-tense word no longer fakes a date context

`is_open_ended` returned true for ordinary prose containing `current`, `ongoing` or `actual` at a word boundary:

```
is_open_ended("Reduced actual costs by 20% in 2023")   = true
is_open_ended("Led the current platform migration")    = true
is_open_ended("Reduced costs by 20% in 2023")          = false   <- control
```

`factual::unsupported_date_issues` decides `date_context` by matching `PRESENT_MARKERS` the same way, so a truthful bullet — "cut spend 30% below the 2019 baseline while keeping the ongoing migration on schedule" — raised a false `factual.unsupported_date` Critical whenever 2019 was absent from the source.

Word boundaries were already the previous fix for this class (`present` inside *presented*), so the fix is deliberately not "add more boundaries".

## 2. Translation goes through the active provider

The root bug: `translate_if_needed` read `store.embedding_config()` and used the embedding provider as a proxy for the active one. It never asked which provider the user had actually selected.

Consequences measured from diagnostics logs:

- `first_chat_model()` returned whatever Ollama listed first. Nobody chose the 27B model; it just sorted first. **117 seconds for one job-ad translation**, which starved the 15s embed timeout and silently degraded semantic scoring to keyword-only.
- Every embed error mapped to `"Ollama unreachable"`, which misdiagnosed a timeout as a connectivity failure twice.

Now reads `ai_config::active_config()`. The `is_local()` cost gate is untouched — metered API keys still skip translation, because autopilot translates up to 20 postings per scheduled run indefinitely.

A fast Ollama-specific reachability gate was restored during review: dropping `reachable_chat_model` had removed the 3s skip, so an unreachable daemon would block each attempt for the full completion deadline — the same run-starving shape this PR exists to fix, reachable again by another route.

Also logs the embedding model on success and failure, and makes the silent `nomic-embed-text` fallback loud.

## 3. The bare-domain link widening was reverted

Linking bare lowercase domains was implemented, then reverted after review found `urls_in()` serves two opposite purposes:

- in the validator, a match means **this text is a link**
- at `pipeline/resume/source.rs:312`, a match means **delete this token from the user's résumé**

The widened rule silently deleted lowercase technology tokens from source résumés, over-linked lowercase library names in exports, reopened a documented false-Critical regression, and made `urls_in("jane.doe@gmail.com")` return `["gmail.com"]` — a phantom domain the renderer never produces, creating a fresh renderer/validator divergence in the opposite direction.

A context-aware version was investigated and rejected as not reachable in a small diff: `split_urls`/`tokenize_rich` are called generically from `model/adapter.rs`, `export/typst_engine/letter.rs`, `contact_profile/`, and `export/docx_renderer.rs` with no line-kind context threaded through.

Reverting restores the pre-branch 4-arm regex on both sides. The renderer no longer hyperlinks such text either, so nothing renders as a verified-looking link the validator cannot see.

**Still open:** the original report. Bare domains without a path remain plain text.

## 4. Data loss fix found by the gate

`source.rs:312` filtered stack-line tokens with `!item.is_empty() && urls_in(item).is_empty()` and no `names_a_resource` guard, unlike its two siblings at `:253` and `:288`. Any `urls_in` match deleted the token outright. Guard added; covered by a test that fails without it.

## 5. Prompt quality control in AI settings

`promptQuality` was only reachable from the résumé analyzer despite applying to the Ollama branch generally. Adds the control to the AI settings page with search-index entries and en/de translations.

## Gate fixes

- **Leak guard was line-number keyed**, so the routing commit shifted a declared site 1153 -> 1176 and made the branch unpushable. Entries can now opt into content-anchored matching via the log call's own format string; an entry survives its site moving within the same file. The other ~56 entries are a documented follow-up.
- **`try_state` restored** where it had been changed to `state`, which panics — the module contracts it must never error.
- **Two new `tracing::warn!` sites sanitized.** One wraps a `reqwest::Error`, and Gemini authenticates via a `?key=` query param.
- **`documents/mod.rs` was at exactly 1400 LOC**, the R8 hard cap, with comments trimmed to fit. Split into `documents/embedding.rs` (-244/+221) instead.
- **Dead disjunct removed** in `entry.rs`: `|| is_open_ended(s)` could never be true once `is_open_ended` required a year. Two doc comments describing the pre-fix behaviour as current were corrected.

## Verification

`cargo fmt --check`, `cargo clippy --all-features --locked --all-targets -D warnings`, and `cargo test --all-features --locked` (without `--lib`, which skips `tests/*.rs`) all clean: 4395 lib + 18 architecture + 20 egress + 3 eval + 3 kill-recovery.

Every new guard was mutation-checked in both directions — the mutation was applied, the test observed failing, then reverted. A green suite was treated as uninformative here, since all three defects on the parent branch were invisible to 4300+ passing tests.

## Summary by CodeRabbit

* **New Features**
  * Added AI prompt-quality settings with Full, Automatic, and Compact options, searchable from Settings.
  * Added localized English and German text for the new setting.

* **Bug Fixes**
  * Improved translation provider and model selection, including clearer Ollama availability handling.
  * Improved Ollama timeout and connection-error reporting across AI operations.
  * Fixed date recognition for ongoing roles and reduced false date detections.
  * Preserved technology names that resemble domains, such as `crates.io`, when importing resumes.

### DIFF
--- a/apps/desktop/src-tauri/src/commands/ai_provider/mod.rs
+++ b/apps/desktop/src-tauri/src/commands/ai_provider/mod.rs
@@ -1096,24 +1096,6 @@ pub fn resolve_by_name(name: &str, base_url: Option<String>) -> AppResult<Box<dy
     Ok(resolve(provider_id, base_url))
 }
 
-/// Discover a reachable local chat model for `provider_id`.
-///
-/// Returns `Some(model_name)` only when the provider is reachable and has an
-/// available chat model. CLI agents and cloud providers always return `None`.
-pub async fn reachable_chat_model(provider_id: ProviderId) -> Option<String> {
-    match provider_id {
-        ProviderId::Ollama => {
-            let (reachable, model) = ollama::reachable_model().await;
-            if reachable {
-                model
-            } else {
-                None
-            }
-        }
-        _ => None,
-    }
-}
-
 // ── Embeddings ────────────────────────────────────────────────────────────────
 
 /// Vector-FORMAT version — bumped whenever the ALGORITHM that produces a
@@ -1313,10 +1295,19 @@ pub fn friendly_api_error(
     }
 }
 
-/// Map a *transport* failure (the `send()` that raced a completion HTTP call
-/// never got a response at all) to `AppError::Timeout` or `AppError::Network`.
-/// Distinct from [`friendly_api_error`], which maps a response the server DID
-/// send back.
+/// Map a *transport* failure (the `send()` that raced an HTTP call never got a
+/// response at all) to `AppError::Timeout` or `AppError::Network`. Distinct
+/// from [`friendly_api_error`], which maps a response the server DID send
+/// back.
+///
+/// Named for its original completion call sites but equally correct for any
+/// other request against the same pooled client (list-models, model-pull,
+/// embeddings, tool-calling): the classification below depends only on
+/// `is_timeout()`, never on what the request was *for*. Ollama's embed path
+/// (`ollama::embed_with`) once mapped every transport failure — including a
+/// real embed-timeout while the daemon was up and busy — to "unreachable",
+/// which sent two separate investigations to the wrong root cause; it now
+/// routes through here too.
 ///
 /// `is_timeout()` walks `e`'s WHOLE source chain, not just this crate's own
 /// `.timeout()` call: it also matches an inner `hyper::Error::is_timeout()`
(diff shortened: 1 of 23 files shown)


saeedkolivand:
🎉 This is included in version 0.138.1.

PR #1059, opened 2026-08-20
fix(documents): score the saved résumé the score tab promises

saeedkolivand:
Reported from a running dev session: the Score tab added in #1056 still read **"Save a résumé to score"** after #1057 supposedly fixed it — on an application reached through Autopilot → Apply.

## Why #1057 didn't cover it

#1057 seeds `resumeDocId` from the default résumé, but only when the seeded text **is** that document's text. That guard is correct for the field's own consumer: `useTailorPipeline` sends `resumeText: ''` whenever an id is set, so a mismatched id would silently generate from a different résumé than the one on screen.

It is the wrong constraint for the Score tab, which asks a different question.

`AutopilotPage` seeds `applySeedResume: ap.resumeText` — a snapshot taken when the autopilot was created. Re-import your résumé (new document id) and that snapshot no longer equals the document's text, so the guard correctly declines to seed, and the tab has no id to score with. Permanently, for every autopilot-originated application.

Confirmed against the reporter's own data: one document, `is_default = 1`, indexed, 8665 chars — `defaultResumeId` resolved fine the whole time. The id simply never reached `JobAdView`.

## The fix

The tab's own copy already states what it scores:

> Scored against your saved résumé, not the tailored version shown here.

So it wants the saved document even when the editor holds something else. It now prefers an explicitly-picked document and falls back to the default — which is what the Jobs page has always scored via `useDefaultResumeId`, and what the comment on that very line already *claimed* to do while reading the form field instead:

```
// Threaded to JobAdView's Score tab so it reads the same résumé the Jobs page
// would score.                                    ← the intent, not the behaviour
```

The form field keeps its strict meaning for the pipeline. Only the id handed to `JobAdView` changes. With no saved résumé at all there is still no id — never a fabricated one.

## Verification

Three new tests on the fallback (default used, explicit pick wins, nothing fabricated when there are no documents). Mutation-checked: dropping the fallback fails the first one.

`pnpm typecheck` clean · `eslint --max-warnings 0` clean · full renderer suite **7219 passed**.

Two existing `@/services` mocks needed `useDocuments` added, since `TailorFlow` now resolves the default résumé.

## Follow-up, not in this PR

With the tab finally rendering, it immediately exposed that the score itself is not trustworthy for non-English postings: a German ad yields *"Keyword coverage 7% across 291 job keywords"* with `abgeschlossenes`, `abgestimmt`, `abseits` — and a postcode, `13385` — listed as missing keywords. `documents::keywords::STOPWORDS` is English-only and the token filter admits any token longer than three characters, so the denominator fills with German function words and numbers. The codebase already documents this (`evidence/mod.rs`), and `STOPWORDS` is formula-version-pinned, so it needs its own change.

### DIFF
--- a/apps/desktop/src/renderer/features/documents/components/TailorFlow/TailorFlow.generation-error.test.tsx
+++ b/apps/desktop/src/renderer/features/documents/components/TailorFlow/TailorFlow.generation-error.test.tsx
@@ -73,6 +73,10 @@ vi.mock('@/components/ui/ModelSelector', () => ({
 }));
 
 vi.mock('@/services', () => ({
+  // `TailorFlow` resolves the DEFAULT résumé for the Score tab's fallback id
+  // (`useDefaultResumeId` reads this) — empty list = no default, which keeps
+  // these tests on the pre-existing 'no résumé' behaviour.
+  useDocuments: () => ({ data: [], isLoading: false }),
   useResolveJobUrl: () => ({ data: undefined, isLoading: false }),
   useExtractText: () => ({ mutateAsync: vi.fn(), isPending: false }),
   useActiveModelCapabilities: () => ({ data: { supportsWebSearch: false }, isSuccess: true }),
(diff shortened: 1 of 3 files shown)


saeedkolivand:
🎉 This is included in version 0.139.2.

PR #1070, opened 2026-08-21
docs: repair the citations the history rewrite broke, and stop citing shas

saeedkolivand:
The history rewrite moved every commit sha. Four references in tracked docs pointed at main commits and stopped resolving.

**I reported one, and that was wrong.** My sweep only tested shas that happened to resolve in my local clone at the time, so it undercounted. Re-checked against the pre-rewrite mirror, every one of them resolved before the rewrite and does not resolve in a fresh clone now. (GitHub's API still answers for the old shas, because it holds force-push orphans for a while — which is exactly why "it still works when I try it" was not evidence.)

## One was worse than a stale sha

Two ADRs cite findings from the 2026-08-16 full-history audit and told the reader to run `git show cd1219ef7:AUDIT_REPORT.md`.

`AUDIT_REPORT.md` was a dead path, so the rewrite removed it from **every commit**. Refreshing the sha would not have helped — the file no longer existed anywhere in the repo, and the doc's own fallback ("find the commit for PR #1000 and read it there") was dead for the same reason.

Recovered from the pre-rewrite mirror and published to the parentless `assets` branch. Verified: **1,588,368 bytes**, and all four finding ids the ADRs name (`p2-contra-cross-001`, `renderer-feat-3-001`, `p2-b1-ipc-chain-001`, `p2-b1-ipc-chain-002`) are present in it. That branch is the right home for something read occasionally and revised never — 1.5 MB once, no history growth.

## The rest

| Doc | Was | Now |
| --- | --- | --- |
| `0015-ripbook-notebook-landing` | `66c36c68` | PR #709 (`d72a401652`) |
| `adr-018-…-aurora-…` | `8688eb91` | PR #313 (`2b254fbcac`) |
| `EXPORT_TEMPLATES.md` | `2fb85227` | names the file, not a commit |

That last one is its own small finding: the sha was **already dangling before this rewrite** — an earlier one broke it and nobody noticed — and the caveat was obsolete anyway. It said `cargo test` works "on branches containing the build.rs delay-load fix". `DELAYLOAD` is in `apps/desktop/src-tauri/build.rs` on main, so it is unconditional.

## The actual lesson, now written into each site

A commit sha in a doc is precisely the drift-prone literal rule 17 forbids, and this repo has now had the same citation broken by **two separate rewrites**. The ADRs anchor on PR numbers, which no rewrite can move, and carry the sha only as a convenience explicitly labelled perishable.

## Checked, not assumed

Six shas in `.claude/scratch/` notes name commits that were **never on main** — unmerged branch work, still served through GitHub's `refs/pull/*`, which a rewrite of main cannot touch. Verified against the pre-rewrite mirror rather than left to inference. Untouched.

Final sweep: zero references in `docs/` or `.claude/` fail to resolve in a fresh clone.

### DIFF
--- a/docs/EXPORT_TEMPLATES.md
+++ b/docs/EXPORT_TEMPLATES.md
@@ -357,8 +357,11 @@ confined to the test function; no typst types appear in production code paths.
 
 **Dev note — preview-regen.** The two preview generators are `#[ignore]`
 libtest fns (`generate_templates_showcase_banner` and
-`generate_cover_template_previews` in `typst_engine/test.rs`). Local `cargo test`
-now works on branches containing the build.rs delay-load fix (2fb85227). Preview
+`generate_cover_template_previews` in `typst_engine/test.rs`). Local `cargo test` works
+unconditionally — the delay-load fix that used to be branch-dependent lives in
+`apps/desktop/src-tauri/build.rs` on main (grep `DELAYLOAD`). The sha that
+caveat named stopped resolving in an earlier history rewrite and went unnoticed,
+which is the argument for naming the file rather than the commit. Preview
 assets may still be stale/incomplete: `cadence`, `regent`, `aria`,
 and `saffron` may lack `.svg` (résumé + cover), the `classic` preview
 may still reflect older `classic.typ`, and the résumé showcase banner may show the old nine. **Regenerate in CI or on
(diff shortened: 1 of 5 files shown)


saeedkolivand:
🎉 This is included in version 0.140.0.

PR #1085, opened 2026-08-31
feat: agent-facing CLI so an AI agent can drive the app

### LINKED ISSUE #1084: Expose local API (or browser-accessible UI) for AI agent automation
AI Job Hunter is currently a Tauri desktop app — a native window with no way for an AI agent (browser-automation tools only) to read or interact with it. To let an AI assistant operate it reliably, we need one of two things, in order of preference:

## Option A (preferred): Local HTTP API

Expose a localhost-only HTTP API the app already has the data for, so an agent can query it directly instead of trying to read pixels off a window.

- Bind to `127.0.0.1` only (not `0.0.0.0`) — no external exposure.
- Protect it with a local token (e.g. written to a file in the app's own AppData folder at startup, like `%LOCALAPPDATA%\AI Job Hunter\api-token.txt`) so only local processes with file access can call it.
- Read-only endpoints to start:
  - `GET /api/best-matches` → list of matched jobs (id, title, company, location, match score, date found, short description)
  - `GET /api/jobs/{id}` → full job detail (description, requirements, application URL/route, source)
  - `GET /api/profile` → current CV/profile summary being matched against
  - `GET /api/automations` → list of configured automations and their status
- Serve a machine-readable schema at `GET /api/openapi.json` so an agent can self-discover endpoints instead of hardcoding them.
- **No auto-submit/auto-apply endpoint.** Application submission should stay a manual, human-approved step — the API is for discovery/evaluation data only, not for triggering irreversible actions.

## Option B (fallback): Browser-accessible UI

If the API is a bigger lift right now, an interim option is running the existing frontend as a normal local web page (e.g. `http://127.0.0.1:PORT`), using the same localhost-only + token-gated access. That alone would let an agent navigate it with standard browser automation, no native window access needed.

Either option unblocks reliable automation; Option A is more robust long-term since UI automation is brittle against redesigns, while an API is stable and explicit about what an agent can and can't do.

saeedkolivand:
Closes #1084.

## Why

The issue asked for a localhost-only, token-gated read-only HTTP API plus an `openapi.json`, because a Tauri desktop app is a native window an AI agent cannot read or drive. The owner redirected to a **CLI**: an agent that can run shell commands needs no HTTP listener, which deletes most of the requested security surface rather than securing it — no port, no CORS, no origin allowlist, no token file to protect.

## What

`ajh-tauri agent <verb>` — a **mode of the existing binary**, not a second `[[bin]]` (the release upload globs only read `target/release/bundle/**`, so a second binary would ship to nobody). It is a thin client over the loopback bridge the native-messaging host already relays to, so no new port, listener or credential, and query logic stays in one place.

```
ajh-tauri agent best-matches [--limit N]   ajh-tauri agent job <url>
ajh-tauri agent profile                    ajh-tauri agent automations
ajh-tauri agent schema                     ajh-tauri agent --help
```

`url` is the cross-resource key, not an id — `BestMatchRow.key` is a cluster id, not a stable handle. There is deliberately **no apply/submit verb**: the issue's one hard boundary.

## Design notes

The argv sentinel sits below the native-host short-circuit and above `run()` — load-bearing both ways, since `run()` forks the minidump supervisor first and the single-instance plugin would otherwise hand the CLI's argv to the running GUI and print nothing.

Reading the data directory from a second process was rejected on evidence: `ApplicationStore::open` runs backfills that write to `ai_generations.db` and can create `Application` rows, and two processes racing `run_migrations` can leave the app booted with no store at all.

Every payload is an allowlist projection, not a delegated record — `Autopilot` alone carries `resume_text`, `cover_letter`, `assistant_notes` and full scraped descriptions. Scraped text goes through `prompt_fence`, the same primitive `answer_assist.rs` already applies to the identical string.

## Verified end to end against a running app

All five resources return real data; `automations` renders 4 autopilots in 1.4 KB from a 9.4 MB store. An 8,697-char real posting comes back fenced and capped. `trust` serialises as exactly `{flags, level, score}`. Throttling holds across three separate processes with fresh sockets — proof the `BridgeState` scoping was necessary. A leak sweep over real payloads found none of the nine forbidden keys, and a distinctive snippet of the real stored résumé appears in no payload.

Two defects were found by testing rather than reading: distinct failures collapsing into one misleading sentinel, and a `next_json` timeout re-armed every loop iteration that let a ping flood hang an invocation indefinitely.

## Also in here

Fleet hygiene found while adding the new domain owner: `cleanup` holds `Edit`/`Write` and deletes code but loaded no contract skill at all, because it was filed under "cross-cutting critics". Reclassified as a writer, and the rule now keys on capability rather than table position. The critic contract — loaded by all 13 critics — gained rules on stating a finding (quote the span, take the `UNKNOWN` escape hatch) and holding one (only new evidence moves a severity, never pushback).

## Review status

Security and protocol reviews ran and their findings are fixed (a blocking HIGH on unfenced scraped text, a nested projection passthrough, origin discrimination, inline dispatch stalling the read loop, a UNC `dataDir` that manufactured outbound SMB, and an unreachable timeout sentinel). The pre-PR ensemble was **cancelled at the owner's request** and has not run.

## Summary by CodeRabbit

- **New Features**
  - Added the `ajh-tauri agent` command-line interface for machine-readable JSON output.
  - Added read-only access to best matches, jobs, profiles, automations, and schema information.
  - Added help output and standardized exit codes and error responses.
  - Added Windows console support for CLI launches.

- **Security & Reliability**
  - Restricted agent requests to authenticated CLI connections.
  - Added throttling, bounded responses, protected data projections, and prompt fencing.
  - Added confirmation and safety guidance for future destructive commands.

- **Documentation**
  - Added agent CLI standards, review workflows, and supporting contributor guidance.

### DIFF
--- a/.claude/agents/agent-cli-author.md
+++ b/.claude/agents/agent-cli-author.md
@@ -43,7 +43,8 @@ Your consumer is an LLM with shell access, not a person. That drives everything:
 ## Before you call it done
 
 `cargo fmt` · `cargo clippy --all-features --all-targets -- -D warnings` · `cargo test --lib` ·
-`cargo test --test architecture`. Then **run the real binary against a running app** — a unit test
+`cargo test --test architecture` · `cargo deny check` · `cargo audit` · `cargo machete` (the full
+R10–R13 gate), plus `cargo test --test egress` if you touched an outbound host literal. Then **run the real binary against a running app** — a unit test
 against the state machine does not prove the two halves talk; this surface has already shipped a
 30-second hang and an empty-payload scare that only an end-to-end run exposed. `cargo test --lib`
 flakes ~1 run in 3 on a pre-existing `rate_limiter` overflow in scraping tests; re-run and say so.
--- a/.claude/agents/agent-cli-author.md
+++ b/.claude/agents/agent-cli-author.md
@@ -1,0 +1,52 @@
+---
+name: agent-cli-author
+description: WRITE-access implementer for the agent-facing CLI surface (`ajh-tauri agent …`) — CLI verbs, argv parsing, the agent.query/agent.result bridge resources, allowlist projections, the machine-readable output contract, help/schema generation, and destructive-command ergonomics. Use for changes to extension_bridge/agent_cli.rs and agent_read.rs, or when adding/changing any CLI verb an AI agent can invoke. Implements to spec; never approves its own work — agent-cli-reviewer audits it (tauri-security-reviewer on destructive or data-exposing verbs).
+tools: Read, Grep, Glob, Edit, Write, Bash, mcp__graphify, mcp__codegraph, mcp__mcp-search
+model: sonnet
+---
+
+You implement the agent-facing CLI — the surface an AI agent drives instead of reading pixels off a
+native window. **First `Read` `.claude/skills/author-contract/SKILL.md` +
+`.claude/skills/agent-cli-standards/SKILL.md`** (subagents don't auto-load skills). Add
+`.claude/skills/extension-standards/SKILL.md` when you touch the bridge transport, and
+`.claude/skills/security-checklist/SKILL.md` for any verb that mutates or exposes data.
+
+## Primary paths
+
+- CLI client + argv: `apps/desktop/src-tauri/src/extension_bridge/agent_cli.rs`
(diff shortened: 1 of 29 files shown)


saeedkolivand:
🎉 This is included in version 0.145.0.

PR #1086, opened 2026-08-31
docs: record the agent surface decisions and sharpen its vocabulary

saeedkolivand:
Follow-up to #1085, which merged without the per-PR steward pass — so the architecture it settled had no decision record, while the standards skill it shipped says "settled — do not redesign without an ADR". This closes that, and adds the glossary section a grill against the domain model produced.

## ADR-037 — what shipped

A mode of the existing binary rather than a second `[[bin]]` (release upload globs read only `target/release/bundle/**`, so a second binary ships to nobody); the argv sentinel placement that is load-bearing in both directions; and a thin client over the bridge rather than a second reader of the stores.

That last one was rejected on evidence, not taste: `ApplicationStore::open` runs backfills that write `ai_generations.db` and can **create** `Application` rows, and two processes racing `run_migrations` can leave the app booted with **no** `ApplicationStore` — the whole application tracker silently reads empty.

## ADR-038 — the approved parity approach

- A **164-row policy table** that must match `generate_handler!` exactly, asserted by test. That is ADR-014's static-allowlist invariant applied to *inbound* dispatch — an unclassified command fails CI instead of shipping.
- **Two tiers, visibly different grammars** (`agent <resource>` vs `agent call <ns>:<command>`) — the `aws s3` / `s3api` split, so a caller knows before running whether the reply carries a projection guarantee.
- **An irreversible command demands a proof its own refusal withholds.** There is no derivable dry-run — no simulation path exists anywhere in the app — so safety comes from *where the token lives*: satisfying it requires a second call to a different verb, and therefore requires having actually read the record.
- **`dispatched`, not an overloaded `ok`.** ~47 commands signal failure in-band inside a `Value`, so the dispatcher cannot know whether a call succeeded; it should not claim to.

It also **amends ADR-0005 in the open**: `agent call` returns records raw, résumé text included, into what is by design an LLM context. That was chosen deliberately, and is written down rather than left implicit — a promise that quietly became false is worse than one openly narrowed.

## Vocabulary

`CONTEXT.md` gains six terms the repo had been overloading: **Command** (164, the authoritative registry), **Channel** (marked explicitly non-authoritative — `NOTIFICATIONS_CHANNELS` is literally `{} as const`), **Resource**, **Verb**, **Effect class**, **Confirm proof**.

## Two citation corrections

Found while writing these, and worth knowing: the repo runs **two ADR families**, so `ADR-005` and `ADR-0005` are different documents. The local-data promise is ADR-0005 (`0005-network-egress-privacy-boundary`), not `adr-005`; the bridge handshake is ADR-0010, not `adr-015`. `check:adr-citations` confirms 23 closed plus 38 open, none ambiguous.

Docs only — no source changes.

## Summary by CodeRabbit

* **Documentation**
  * Added a glossary explaining Agent CLI concepts, command behavior, safety classifications, and confirmation requirements.
  * Documented the Agent CLI’s architecture as a thin client that communicates with the running app.
  * Documented full CLI parity with the app through curated and generic command tiers.
  * Updated the knowledge-base index with new decision records and the latest update date.

### DIFF
--- a/docs/CONTEXT.md
+++ b/docs/CONTEXT.md
@@ -502,3 +502,46 @@ Was the shader-driven wobble of ink strokes (`uBoil`) that gave RIPBOOK its hand
 Retired - TERMINAL VELOCITY is a PBR realistic film, no ink boil. The only "on twos" that
 carries over is character animation stepped at ~12 fps against a smooth 60 fps camera.
 _Avoid_: reusing Line boil / `uBoil` for any TERMINAL VELOCITY effect
+
+## Domain — Agent CLI surface
+
+**Command**:
+A `#[tauri::command]` Rust function. There are 164, and the set is exactly 1:1 with the
+`generate_handler!` registration list — diffed both directions, no ghosts, no unregistered
+commands. This is the **authoritative registry** of what the app can do, and the only source a
+parity surface may derive from.
+_Avoid_: "channel" (a different, incomplete set) / "endpoint" (there is no HTTP surface)
+
+**Channel**:
+A TypeScript-side `namespace:verb` string in `IPC_CHANNELS`, used by the renderer to call a
+Command. **Not authoritative and not complete**: `NOTIFICATIONS_CHANNELS` is literally
+`{} as const`, and `AI_CHANNELS` carries 5 entries for a 29-method contract. A surface derived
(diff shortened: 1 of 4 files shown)


saeedkolivand:
🎉 This is included in version 0.145.0.

PR #1088, opened 2026-09-01
fix: make the agent CLI reachable after an install or update

saeedkolivand:
## Why

ADR-037/ADR-038 shipped a 164-command, effect-classified agent CLI — and no user could invoke it. `ajh-tauri agent` returns "command not found" on a machine where the app is installed, because the binary is only reachable through the `~/.ajh-agent/agent.json` pointer file the app writes for *programs*. There was no human path, and the CLI appeared nowhere in the README, nowhere under `docs/` except `CONTEXT.md`, and nowhere on the landing site. The only landing mention is the `.claude/` development fleet, which a reader will conflate with it.

Same failure shape as a feature shipping dead: built, tested, reviewed, unreachable.

## What

Every platform row below was checked against a **downloaded v0.145.0 artifact**, not inferred — an earlier draft of this work contained two guesses, one of which would have shipped a broken symlink.

| Platform | Before | After |
| --- | --- | --- |
| Linux `.deb` / `.rpm` | already `/usr/bin/ajh-tauri` | unchanged — **no code needed** |
| macOS via Homebrew | not on PATH | `binary` stanza in the cask |
| macOS via dmg drag-install | not on PATH | documented; unchanged |
| Windows `.exe` (nsis) | not on PATH | NSIS `POSTINSTALL`/`POSTUNINSTALL` hook |
| Windows `.msi` | not on PATH | **deliberately unchanged** — see below |
| AppImage | n/a by design | documented |

Plus a README section and `docs/knowledge/agent-cli.md`, which states shape and points at `--help` and the owning symbols rather than copying the verb table, the sentinel list or `max 50` (AGENTS.md rule 17).

### Why the MSI is deliberately untouched

It genuinely lacks a PATH feature (`PathEnvVar` occurs 0 times in the shipped MSI while `INSTALLDIR`/`PATH` strings are readable), and WiX would actually be the *safer* implementation — `<Environment>` handles append and uninstall removal natively with no length ceiling. It is still not worth a second installer surface: floor-adjusted downloads across 73 releases are **122 for the `.exe` against 5 for the `.msi`**, the `.exe` is the primary download button, and it is the only one the updater signs. Revisit if MSI usage grows.

## The Windows hook is the risk-bearing part

PATH is a value whose corruption breaks the user's shell, so every branch is defensive: the value's content never enters an NSIS string variable (`RegQueryValueExW`/`RegSetValueExW` against allocated memory, so there is no 1024-char ceiling), whole-segment `;`-padded matching makes a sibling-prefix directory structurally unable to match, the registry value type is preserved, uninstall splices out exactly our segment and the one delimiter it owns, and every failure path leaves PATH untouched rather than aborting the install. A compile-time guard refuses to build under any `installMode` other than `currentUser`, since the hook hardcodes `HKCU`.

## What review caught

Two rounds of security review, both with the findings *measured* rather than reasoned about — none of these were visible by reading:

- **The uninstaller did not compile.** `${StrLoc}` expands to `Call StrLoc`, and NSIS forbids calling a non-`un.` function from an uninstall section. **PR CI structurally cannot catch this** — it runs `cargo check` and never bundles, so it would have surfaced first in the release job.
- **A failed probe destroyed PATH.** `System::Call` writes the literal string `"error"` into its output register on a resolution failure; `${If} $1 != 0` treated that as "no PATH exists" and replaced the whole value with one entry, then reported success.
- **The feature was a no-op on an ordinary machine.** The original 1016-byte budget refused any PATH longer than that; a developer machine with a 988-char PATH was already over it, so the installer would have silently done nothing while the docs promised otherwise.
- A crash class where a length reached `RtlMoveMemory` without a precondition (zero-length registry value, `GlobalAlloc` returning NULL, an unresolvable `StrStrW`). Data-safe in every case — PATH was untouched in 100% of injected failures — but a crashing installer is still a defect.

The final review re-attacked the splice across 22 shapes including 8 KB values, traced allocation discipline (8 frees against 4 allocations is correct multi-exit cleanup — no double-free, no leak on any `Goto`), and re-derived the offset arithmetic independently rather than trusting the clamps.

## Test plan

Verified locally: both macros compile clean under makensis 3.11 with the install macro in a real `Section Install` **and** the uninstall macro in a real `Section Uninstall`; the `installMode` guard aborts the build when flipped to `perMachine`; the hook was executed against a scratch `HKCU\Software\...` key across absent / `REG_SZ` / `REG_EXPAND_SZ` / `REG_DWORD` values, sibling-prefix directories, re-runs, >1024-char values up to 8 KB, and removal from first/middle/last/only position. The real `HKCU\Environment` was never written.

**Not verified:** no Windows release bundle was produced and installed end-to-end. Before release, on a real install:

1. Fresh install, open a **new** shell, run `ajh-tauri agent --help`.
2. Check `HKCU\Environment\Path` — exactly one entry for the install dir.
3. Re-run the installer (the update path) — still exactly one entry, not two.
4. Uninstall — the entry is gone and the rest of PATH is byte-identical.

macOS: `brew install --cask ai-job-hunter` then `ajh-tauri agent --help`.

### DIFF
--- a/Casks/ai-job-hunter.rb
+++ b/Casks/ai-job-hunter.rb
@@ -29,6 +29,15 @@ cask "ai-job-hunter" do
 
   app "AI Job Hunter.app"
 
+  # The agent CLI is the SAME binary in argv mode (ADR-037): `ajh-tauri agent
+  # <verb>` is detected from argv and short-circuits before the GUI (and before
+  # the single-instance plugin) ever starts. Symlinking it onto PATH is what
+  # makes the CLI reachable by a human — the app's own ~/.ajh-agent/agent.json
+  # pointer solves discovery for *agents*, but nothing put the binary on PATH.
+  # Verified against the shipped bundle, not assumed: the executable inside the
+  # bundle is named after the Cargo [[bin]] (`ajh-tauri`), NOT productName.
+  binary "#{appdir}/AI Job Hunter.app/Contents/MacOS/ajh-tauri"
+
   # The app is not notarized; clear the quarantine flag after install so
   # Gatekeeper doesn't refuse to open it.
   postflight do
--- a/README.md
+++ b/README.md
@@ -307,6 +307,43 @@ pnpm dev
 
 </details>
 
+<details>
+<summary><strong>CLI agent — `ajh-tauri agent <verb>` (headless mode)</strong></summary>
+
+Run commands from your shell or an LLM agent, with the desktop app already running:
+
+```bash
+# Find the best matches across all autopilots
+ajh-tauri agent best-matches --limit 10
+
+# Get full details for one job by URL
+ajh-tauri agent job "https://example.com/job/123"
+
+# Export your profile for autofill (requires opt-in consent)
+ajh-tauri agent profile
+
+# List all autopilots and their status
+ajh-tauri agent automations
+
+# Generic command dispatch (advanced: read/reversible commands, irreversible with proof)
+ajh-tauri agent call "namespace:command" --input '{"key":"value"}'
+
+# Show help
+ajh-tauri agent --help
+```
+
+**Platform-specific paths:** The CLI is the same binary as the desktop app, no separate install needed. On **Linux** and **Homebrew/macOS** it is already on `PATH`. On **macOS with dmg drag-install**, add the path to your shell profile:
+
+```bash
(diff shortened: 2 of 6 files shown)


saeedkolivand:
🎉 This is included in version 0.145.1.

PR #1093, opened 2026-09-02
docs: record the live client handshakes in adr-040

saeedkolivand:
## Why

ADR-040 shipped with one recorded UNKNOWN: which `protocolVersion` the real clients send. It also stated, in its context section, that both Claude Code and Codex use the 2025-11-25 handshake — an assumption from reading client source, not an observation.

## What

After #1092 merged, both documented clients were connected to the built binary and their frames captured through a stdio tee shim:

- **Claude Code 2.1.258** (`claude mcp add`): `initialize` with `protocolVersion` 2025-11-25 → `notifications/initialized` → `tools/list` as its health check; a `claude -p` turn issued a real `tools/call` and answered from the result. Requests carry `_meta.claudecode/toolUseId` and a `progressToken`, both ignored correctly.
- **Codex 0.144.6** (`codex mcp add`): `initialize` with `protocolVersion` **2025-06-18**, then a real `tools/call` carrying `_meta["x-codex-turn-metadata"]`, answered from the running app.

So the context sentence was wrong for Codex, and the echo-if-supported version rule is what made it connect. The ADR's verified section now records the observed sequences and what is still unverified: a client keepalive during a long call, macOS and Linux, and an irreversible dispatch with a matching proof.

## Test plan

Docs only. `pnpm check:adr-citations` and prettier pass.


saeedkolivand:
🎉 This is included in version 0.146.0.

PR #1108, opened 2026-09-04
fix: give extension answer tools a side panel; fix the inline rewrite popover

saeedkolivand:
## Why

Two complaints, one about each surface, turned out to be four defects and one design problem.

On the extension, the Answer tools live in the action popup, and a popup closes the moment it loses focus. That moment is exactly when the tool is used: you click into the page to paste the answer you just drafted, and the tool disappears along with the draft. In the desktop app, the inline rewrite was reported as flaky, so it was driven live against a real model for thirty-eight runs across two sessions. It is not flaky. It has five separate defects, and the worst of them is that you cannot click Accept at all.

The extension redesign was grilled against the existing decision records before any code was written; ADR-044 records the outcome and the alternatives that lost.

### The Answer tools get a side panel, and keep the popup

The tools now live in a persistent side panel, and the popup keeps them too. Neither is the home: both are views over one state per tab and origin, held in session storage by the background worker, so a draft started in one is on screen in the other and closing the popup loses nothing. Versions stay for the session and are never persisted.

Page access does not widen. There are still no content scripts, no host permissions beyond loopback, and nothing is ever auto-submitted. The toolbar click grants access to the tab and opens the popup, whose control opens the panel from that same gesture, and a context-menu entry on selected text opens it directly. That gesture model is the same on both browsers, because a declared popup takes priority over Chrome's open-the-panel-on-click behaviour, which the manifest test now pins as absent. After a cross-origin navigation the panel keeps its rows and every control that would write is replaced by a line explaining how to re-grant access, and the refusal is enforced in the background rather than only hidden in the view.

The page's questions are the primary object. Each gets a row and its own composer, with tone and length chips that rewrite the latest version through the existing wire verb, and a Regenerate that redrafts from scratch. The interface says which is which, because a chip never sees your resume or the posting. Copy is the primary action and Accept the quiet one, absent entirely for a field that is not on the page, and the Accept sentence names the question it will overwrite.

The scan reads each field's own length limit and the panel counts against it live. That limit rides on the draft request as one new optional field, shaped only on the wire so an older desktop drops it and a newer client never has a whole draft refused over a number.

### The inline rewrite in the desktop app

Five defects, all measured on the running app rather than inferred.

The popover rendered inside two clipping containers, so on a generation card the finished result and the whole Cancel, Regenerate and Accept row sat up to 225 pixels below the cut. Nothing was broken in the model; the answer arrived and could not be reached. The component already had a portal mode built for this and was never given the anchor, so it gets one, taken from the toolbar row rather than the trigger button, which unmounts as soon as a rewrite starts.

A result identical to the selection was presented as success with Accept enabled. Shorten returned the input verbatim three times out of three on a short span, differing by a single comma, so the comparison normalises whitespace and trailing punctuation but deliberately does not fold case. An unchanged reply is now named as one, with Accept disabled and Regenerate live, and it is not styled as an error.

The popover gave up after sixty seconds while the backend runs to five minutes or more, inverting the invariant the shared timeout module exists to protect. It did fire in testing. The bound is now derived the way every other stream derives it, and a long pass shows a still-working line instead of looking dead.

A numeric limit in the instruction is now counted in code. The model cannot count: asked for under 200 characters it returned 199, 211 and 171, and asked for at most 40 words it missed six times out of six. The result is measured, re-asked exactly once with the overshoot named, and if it is still over, the count is shown next to Accept and the choice is yours. Nothing is silently truncated and nothing loops.

Finally, the rewrite took its language from the document's metadata rather than the selection, so a Dutch span came back in English in twelve of eighteen runs. It now detects the language of the span itself and falls back to the document only when detection is unsure.

One thing deliberately not done: the surface does not ask for a cheap reasoning tier. Measured, low effort made limit-following strictly worse, returning 204, 438 and 547 characters against a 200 limit.

## Verified

Every new guard was mutation-checked by deleting the feature and confirming a named test fails. The manifest amendments were checked the same way: removing the Firefox sidebar key, moving a permission to the wrong target and dropping a README justification row each turn a specific test red.

Repo-wide lint and typecheck are clean, the Rust suite passes, both browser targets build, and the extension suite runs green with new coverage for the row model, the panel control and the gesture handlers.

## Review history

Two internal rounds before this pull request was opened.

The first ran security, frontend, AI provider and visual passes over the branch as it stood after the panel, the shared wire field and the desktop rewrite fixes landed. Security and the AI provider infrastructure came back clean. Frontend approved with three advisory notes, folded in below. The visual pass drove the actual built extension rather than reading the diff, took ten screenshots, and blocked on one critical and four high defects, each reproduced by executing the real code path: the always visible free text question input lost its typed value and keyboard focus on every single streamed token anywhere in the tab, because every state push tore the whole section down and rebuilt it from nothing; a row's own instruction box lost focus the same way during its own stream, without losing its value; Rescan stayed clickable after a navigation and failed with no visible feedback at all; a chip based rewrite that echoed the input back unchanged was presented as a fresh, acceptable version; and every control on a row stayed enabled while that row's own draft was still streaming. A parallel pass over the desktop side found a sixth blocker: the one time reask that fires when a rewrite comes back over its character limit shared the first attempt's timeout instead of getting its own, so a slow first attempt could make the reask abort and throw away a perfectly usable result for a generic error.

All six were fixed and the fixes verified against the source rather than the description of them: focus and typed text now survive a rebuild through a captured and restored selection range keyed to a stable attribute on each input, the reask gets its own timeout window, and the smaller ones follow the same shape.

A second, stricter pass over the whole diff then found two more defects, both traceable to how this branch inherited an unmerged fix rather than to anything new in its own design. This branch was cut from main before a separate, already reviewed but still unmerged branch landed a fix for a real defect in the extension's build: the default minifier folds away the trailing object literal an injected script uses to hand data back to the extension, because the object's only reader is the browser executing the script, which the bundler cannot see. That fix never reached this branch's base, so the new page scanning script this pull request adds inherited the same defect from day one. It built without error and worked correctly under every unit test, because every test calls the script's functions directly rather than running the built file the way the browser actually does, and it returned the wrong value the moment it was really injected into a page, so every scan for this feature failed silently and the feature never worked in a built extension. The other finding was a genuine race: the last step of a finished draft writes two different things into the same shared per tab record with no coordination between the two writes, so depending on timing one of them could be lost, and losing the wrong one could permanently disable every control on the row whose draft had just finished. Both are fixed now: the build fix is ported from the other branch in the same shape, so picking up main later needs no extra work, and writes to one tab's shared record are now queued so two of them can never land out of order.

Every fix in both rounds was mutation checked: revert it, confirm a named test goes red, reapply it.

## Deliberately deferred

Desktop-side verification of the character limit lands in the same functions an open pull request rewrites, so it is deferred rather than written twice and merged by hand. Until it arrives, the extension states the limit in the prompt and the panel counts the returned text itself, which is why the documentation says so plainly instead of claiming enforcement. The in-page floating card, a keyboard shortcut and a word-level diff between versions are all out of scope, with reasons recorded in the decision record.

No visual verification of the desktop popover fix was possible without running the app, so that fix is proven by document assertions rather than by looking at it.

## Summary by CodeRabbit

- **New Features**
  - Added Answer tools with shared state across the popup and persistent side panel/sidebar.
  - Scan application questions, generate drafts, add custom questions, rewrite answers, select versions, accept answers, and restore originals.
  - Added context-menu access and page-change safeguards.
  - Added answer length-limit handling and visible character/word counters.

- **Improvements**
  - Rewrites now use the selected text’s language, detect unchanged results, respect length instructions, and provide clearer progress feedback.
  - Improved overlay placement so rewrite controls remain visible.

- **Documentation**
  - Documented Answer tools, permissions, and usage.


saeedkolivand:
🎉 This is included in version 0.146.1.

PR #1112, opened 2026-09-05
fix: make tray pause/resume-all autopilots actually work

saeedkolivand:
## Summary
- The tray's "Pause all autopilots" item silently pumped every autopilot to `Paused` on every click and never became "Resume all autopilots" — the menu item handle was a local variable never retained past `build()`, so its label could never change.
- First attempt made it a single stateful toggle deriving direction from live store state. Review caught a real regression that introduced: the label only refreshed on its own click, so pausing/resuming an autopilot from Settings (or a privacy reset wiping them all) left the tray label stale — the next click then did the opposite of what it said.
- Final fix: two separate, always-visible, statically-labeled items. "Pause all autopilots" always pauses every currently-`Active` autopilot (no-op otherwise); "Resume all autopilots" always resumes every currently-`Paused` one (no-op otherwise). Neither label ever changes, so neither can say the opposite of what it does — no relabeling, no cold-start computation, no state to go stale. `Archived` autopilots are untouched by either action.

## Test plan
- [x] `cargo test --lib tray::` — 9/9 pass
- [x] `cargo test --lib` (full) — 5230 passed, 0 failed, 27 ignored

## Summary by CodeRabbit

- **New Features**
  - Added separate tray menu options to pause or resume all eligible autopilots.
  - Pausing affects active autopilots, while resuming affects paused autopilots.
  - Archived autopilots remain unchanged.

- **Tests**
  - Added coverage for status filtering, archived autopilots, no-op actions, and pause/resume cycles.

### DIFF
--- a/apps/desktop/src-tauri/src/tray/mod.rs
+++ b/apps/desktop/src-tauri/src/tray/mod.rs
@@ -15,7 +15,7 @@ use tauri::{AppHandle, Emitter, Manager, Wry};
 
 use crate::events::{emit_event, AUTOPILOT_FOCUS, MENU_ACTION, MENU_NAVIGATE};
 
-use crate::autopilot::{Autopilot, AutopilotStatus, AutopilotStore};
+use crate::autopilot::{AutopilotStatus, AutopilotStore};
 use crate::commands::notifications::{push_and_notify, OsBanner};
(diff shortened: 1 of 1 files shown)


saeedkolivand:
🎉 This is included in version 0.148.0.

PR #1113, opened 2026-09-05
feat: bring Import/Check fit/Fill form/Save answers into the side panel

saeedkolivand:
## Summary
- The side panel only ever showed the Answer tools section (ADR-044). The popup additionally has four page-scoped action buttons the panel was missing: Import this job, Check fit, Fill this form, Save my answers.
- Extracted into a new shared `apps/extension/src/job-tools/job-tools.ts` module (mirroring the existing `mountAnswerTools` extraction), mounted by both popup.ts and sidepanel.ts. Same wire calls both surfaces already used — no new permissions, no manifest change.
- **Trust gate (ADR-045, new):** Chrome's `activeTab` is granted only by specific gestures (toolbar click, context-menu item, keyboard shortcut, omnibox) — never by clicking a button already rendered inside an open side panel, which persists across tab switches with no fresh per-tab gesture. Reuses the existing `AnswerState.pageChanged` signal: when untrusted, the panel replaces the four controls with "Click the toolbar icon to grant access to this page, then use these tools." Both context-menu handlers now force-clear the flag on their tab so a context-menu gesture actually unlocks the panel it just opened.
- Deliberately out of scope: "Mark as applied" and the adaptive Import-button relabel stay popup-only — not among the four requested wire verbs, so panel parity is scoped to exactly those four (ADR-045 decision 5).

## Review history
- First implementation had a real HIGH caught by `extension-reviewer`: the panel's `follow(tabId)` fired the fields-probe trigger synchronously right after subscribing to tab state, but that subscription's first delivery is always async — so the probe ran against the *previous* tab's trust flag (or the cold-mount default), never the newly-followed tab's real state, on every mount and tab switch. Fixed by deferring the trigger to fire from inside the subscription's first delivery, after `render()` has updated the trust flag for the actual tab. Added a regression test that drives the real mount → subscribe → async-delivery order (mutation-tested against the original bug).

## Test plan
- [x] `pnpm -F @ajh/extension typecheck` — clean
- [x] `pnpm -F @ajh/extension test` — 623/623 pass
- [x] `pnpm -F @ajh/extension lint` — clean
- [x] `pnpm check:agent-system` — clean
- [x] No manifest/permission diff

## Summary by CodeRabbit

* **New Features**
  * Added job tools to the Answer-tools side panel: Import this job, Check fit, Fill this form, and Save my answers.
  * Job tools are now available consistently in both the popup and side panel.
  * Added protection that keeps page actions unavailable until the page is refreshed through a supported user gesture.
* **Bug Fixes**
  * Context-menu actions now correctly refresh page access status without rescanning existing answer data.
  * Improved side-panel updates when switching between tabs.


saeedkolivand:
@coderabbitai review

saeedkolivand:
@coderabbitai review

saeedkolivand:
Fixed all 6 in 74207908:
- Origin capture race (background.ts): `rearmPageChangedForGesture` now takes `origin` as a parameter sourced from the context-menu event's own `tab.url`, instead of re-querying the active tab after an await (which could read a different tab if the user switched away meanwhile).
- Cross-tab state leak (job-tools.ts): `render()` now also tracks the last-rendered tab id, so switching between two already-trusted tabs correctly resets the Check-fit score / form-group visibility instead of leaving the previous tab's content showing.
- Stale `follow()` callback (sidepanel.ts): added a generation guard mirroring popup.ts's existing `appliedCheckGeneration` pattern, since `subscribeAnswerState`'s unsubscribe doesn't cancel an already-in-flight read.
- Scoped the leaking `tabsQueryMock` default, added the two origin-degrade edge case tests, replaced two silently-optional test calls with explicit guards, and trimmed the extension-domain.md section to a symbol pointer.

All three logic fixes were mutation-tested. 626/626 extension tests pass.

saeedkolivand:
🎉 This is included in version 0.148.0.

PR #1116, opened 2026-09-05
feat: give the extension side panel connection-status awareness

saeedkolivand:
## Summary
- Unlike the popup, the side panel had no connection-status awareness at all — it always rendered Answer tools + Job tools as if the desktop app were connected, so if the app wasn't running or paired, every button just silently failed on click instead of the panel saying so upfront.
- Extracted the popup's pill/retry button + pair/offline/outdated/searching views into a new shared `apps/extension/src/connection-status/connection-status.ts` module (`mountConnectionStatus`), mirroring the extraction pattern already used for Answer tools (ADR-044) and Job tools (ADR-045). Both `popup.ts` and `sidepanel.ts` now mount it.
- Full parity, as requested: the panel now shows the same pairing screen, offline screen, and outdated-app screen the popup does — not a thinner pointer back to the popup.
- Both surfaces get live status pushes via the same `runtime.onMessage` listener (`background.ts`'s existing `broadcastStatus()`), so the panel stays correct across its whole (long) lifetime without polling. New ADR-046.

## Review
- `extension-reviewer` audited this closely given the two prior extractions in this exact feature area both had real bugs found (a stale-tab-origin race, a cross-tab state leak, a subscription-ordering race). No HIGH/CRITICAL found — every callback in this extraction reads current state live rather than a captured snapshot, which is exactly the property that was missing in the prior two bugs.
- One MEDIUM (missing negative test for message-kind cross-talk) fixed: added a test proving a non-`status` broadcast (`answerAssistProgress`) never triggers a re-render.

## Test plan
- [x] `pnpm -F @ajh/extension typecheck` — clean
- [x] `pnpm -F @ajh/extension test` — 646/646 pass
- [x] `pnpm -F @ajh/extension lint` — clean
- [x] `pnpm check:agent-system` — clean
- [x] No manifest/permission diff
- [x] No popup regression (all popup tests adapted, none dropped)

## Summary by CodeRabbit

* **New Features**
  * Added shared connection-status views to the popup and side panel, including pairing, offline, outdated, searching, retry, and status indicators.
  * The side panel now displays connected tools only when the connection is active.
  * Added live connection-status updates and improved pairing, retry, app-download, and recovery flows.

* **Documentation**
  * Added decision-record documentation for consistent connection-status behavior across extension surfaces.


saeedkolivand:
Fixed all 3 in 9e086fa7:
- `refresh()`/`refreshWithTimeout()` now catch a rejected `send()` (MV3 service worker asleep/crashed) and fall back to the offline/Retry view instead of leaving an unhandled rejection with nothing rendered. `refreshUntilSettled()` already had this — confirmed no gap there.
- Trimmed ADR-046 and its README index row down to pointers instead of restating DOM ids/callback behavior.

New tests cover both the `start()` path and the direct `refresh()` path rejecting. 648/648 extension tests pass.

saeedkolivand:
🎉 This is included in version 0.149.0.

Issue #1115, opened 2026-09-05
Agent/MCP cannot enumerate Autopilot.found_jobs — no pagination, all 8 autopilots exceed the 256 KiB MCP result cap


NimaNabi:
## Summary
No Agent/MCP-exposed command can enumerate an autopilot's complete `found_jobs` list. `autopilot_get`/`autopilot_list` return the full, unbounded `Autopilot` record, and every one of the app's 8 current autopilots (459 KB–1.87 MB) exceeds the MCP bridge's own `MCP_RESULT_MAX_BYTES = 256 * 1024` cap (`extension_bridge/agent_cli/mcp.rs:622`), so both commands are refused outright (`result_too_large`) for every autopilot tested. The only Agent/MCP-reachable read path, `autopilot_best_matches`, is a separate cross-autopilot ranked projection hard-capped at `BEST_MATCHES_CAP = 100` (`commands/autopilot/best_matches.rs:32`) — a constant the command's own doc comment calls "a payload guard, not the selection rule," and the command takes **no parameters at all**, so a caller-supplied `limit` is silently ignored.

This is **not** a database limitation and **not** a GUI limitation — see below.

## What we verified (source-referenced)
- `useAutopilots()` → `api.autopilot.list()` (`renderer/services/use-autopilot/use-autopilot.ts:29-36`) loads the full `Autopilot[]`, each with its complete `foundJobs` array, into the renderer.
- `AutopilotCard` receives the whole `Autopilot` object as a prop and renders the full `foundJobs` array (filtered only for cluster-canonical rows) into a plain scrollable `max-h-64` div — `renderer/features/autopilot/components/AutopilotCard/index.tsx:248-251, 715-717`. There is no frontend pagination anywhere in this component.
- Backend `autopilot_list`/`autopilot_get` (`commands/autopilot.rs:101-117`) take no filter, offset, cursor, or board-scope parameter — full record or nothing.
- `autopilot_best_matches` (`commands/autopilot.rs:759-761`, `commands/autopilot/best_matches.rs:29-32`) takes zero arguments; `BEST_MATCHES_CAP = 100` is a hardcoded constant with no request-side override.
- The Agent/MCP bridge's transport cap, `MCP_RESULT_MAX_BYTES = 256 * 1024` (`extension_bridge/agent_cli/mcp.rs:622`), applies to every JSON-RPC response regardless of command.
[...]

Issue #1114, opened 2026-09-05
No automatic post-discovery description enrichment for search-only boards (LinkedIn + 6 others)


NimaNabi:
## Summary
No automatic post-discovery enrichment pass exists for boards whose search results never include a description — LinkedIn's free tier plus six others (TheMuse, Comeet, Breezy, BambooHR, Pinpoint, Rippling, per the maintainer's own comment at `commands/autopilot.rs` — see below). This is a distinct, narrower remaining gap from #1105, which is closed and correctly scoped to *signaling* honesty, not content fetching.

## What's already fixed (not being re-litigated)
#1105 (closed, PR #1109) fixed the score/trust dishonesty around missing content: `score_provisional` now fires for any title-only-scored posting, and a `DescriptionUnavailable` trust flag caps confidence below High. Confirmed still correct and unchanged in current `main`.

## What already exists (pull-based, manual only)
`scraping/scrape_url::resolve` → `try_linkedin` (`scraping/scrape_url/mod.rs:994`) can fetch a real LinkedIn description for a single URL, using an authenticated client + real selectors. It's invoked from exactly two places, both requiring an explicit action: `scrape_url`/`scrape_resolve_url` (`commands/scrape.rs:369,597`) and the browser-extension's "import this job" flow (`extension_bridge/import_flow.rs:233,255,263`). No caller anywhere invokes it automatically after autopilot discovery — confirmed via a repo-wide search for any background/queue/worker/scheduled description-fetch mechanism (zero matches beyond unrelated `backfill_*` functions for geocoding, salary, application linking, etc. — none touch job descriptions).

## The actual gap
`scraping/linkedin/api_client/mod.rs:303` — every LinkedIn search-result posting is built with `description: Some(String::new()), // Will be filled in background`. No such background process exists. The maintainer's own comment on the `no_jd_text` flag (added in the #1109 fix, `commands/autopilot.rs` ~line 845) confirms this precisely:

> "Six other boards never populate a search-result description either (no detail-enrichment pass runs between the initial scrape and this projection for any of them): TheMuse, Comeet, Breezy, BambooHR, Pinpoint, Rippling."

Same statement is echoed in `docs/knowledge/matching-algorithm.md`. So this isn't a LinkedIn-only complaint, and it isn't disputed that the gap exists — it's already documented in the codebase, just not tracked as an actionable item anywhere I could find.

## Why this matters for bulk/automated qualification
[...]

PR #1117, opened 2026-09-06
feat: LinkedIn description enrichment + paginated found-jobs MCP resource

saeedkolivand:
## Summary

Two independent fixes for the Agent/MCP job-discovery surface, bundled into one PR by request:

### 1. LinkedIn post-discovery description enrichment (#1114, LinkedIn slice)
- LinkedIn search results never carry a description (`LinkedInJobsApiClient` always writes `Some(String::new())`), so LinkedIn postings scored title-only forever. The other 6 boards named in #1114 have no per-URL detail resolver at all (confirmed by inspection) and are out of scope here.
- A best-effort background pass, spawned after `autopilot_run`'s `record_run` call, resolves each still-blank LinkedIn posting via the existing `scraping::scrape_url::resolve` single-URL resolver, paced through LinkedIn's existing shared rate limiter, and writes any real description back through the existing `scrape_update_description` mechanism (#1109's write-back fix).
- Capped at 15 fetches/run; a per-URL failure is logged and never fails the run itself.
- **Review history**: `scraping-applier-expert` caught a CRITICAL — `merge_found_jobs` treated LinkedIn's `Some("")` blank sentinel as "known," silently wiping every enriched description back to blank on the next scrape and re-fetching it forever. Fixed with a `description_is_blank`-aware merge guard, mutation-tested. The pre-push architecture gate then caught a real L2→L3 layer violation in the fix (Tauri types/command calls leaking into an `autopilot_helpers` module); fixed by splitting pure logic from Tauri orchestration across the existing layer boundary this crate already uses for `rerank.rs`/`best_matches.rs`.

### 2. Paginated `found-jobs` Agent/MCP resource (#1115)
- No Agent/MCP command could enumerate an autopilot's `found_jobs` — every real autopilot (459 KB–1.87 MB) exceeds the MCP bridge's 256 KiB response cap, and the only alternative (`autopilot_best_matches`) is a hardcoded top-100 across all autopilots combined with no parameters.
- New read-only, paginated `found-jobs` resource (CLI verb, MCP tool, `agent.query` resource): `{autopilotId, limit?, cursor?} → {jobs, nextCursor, total, autopilotId, autopilotName}`.
- **Review history**: `agent-cli-reviewer` and `tauri-security-reviewer` both reviewed in parallel (new data-exposing surface). Two HIGH findings from `agent-cli-reviewer`, converging with the security review's MEDIUM: a numeric `cursor` was silently coerced to "absent," resetting pagination to page 1 without any signal; and the page-size guard bounded characters, not bytes, so the real worst-case page was ~2.5 MB, not the ~168 KB either review first estimated — reachable via ordinary non-ASCII postings, not just adversarial ones. Fixed with a byte-budget trim (drops rows until the serialized page fits, cursor reflects rows actually kept) and correct `Value`-variant cursor parsing.

## Test plan
- [x] `cargo test --lib` — 5265 passed, 0 failed
- [x] `cargo test --test architecture` — 18/18 passed
- [x] `cargo test --test mcp_smoke` — fixed a hardcoded tool-list assertion that hadn't been extended for the new resource
- [x] `cargo deny check` / `cargo machete` — clean

## Summary by CodeRabbit

- **New Features**
  - Added asynchronous enrichment for missing LinkedIn job descriptions after autopilot runs.
  - Added a paginated `found-jobs` resource and tool for viewing an autopilot’s discovered jobs.
  - Added pagination controls, result limits, and safe field filtering for job data.

- **Bug Fixes**
  - Existing job descriptions are no longer overwritten by blank LinkedIn results.
  - Individual enrichment failures no longer interrupt autopilot completion.

- **Documentation**
  - Documented LinkedIn description enrichment and the new `found-jobs` capability.


saeedkolivand:
Fixed all 3 in d934709a + 56bcbfee:
- LinkedIn enrichment (`commands/autopilot/linkedin_enrich.rs`): in-flight coalescing keyed by URL (`claim_unclaimed` / `release_claim` around a process-wide set, same shape as `RUNS_IN_FLIGHT`), so a re-run before the previous pass settles no longer double-fetches the same postings. Regression tests cover the overlapping-pass and release-then-reclaim cases.
- found-jobs (`agent_read/found_jobs.rs`): `autopilotName` is now fenced, and the byte budget measures the full response envelope (a measured `base_cost` for the metadata fields is subtracted before rows are accumulated), not just the `jobs` array. New test proves a maxed-out name plus worst-case CJK rows still stays under `MCP_RESULT_MAX_BYTES`.
- `docs/knowledge/automation-domain.md`: trimmed to a capability line plus symbol pointers.

5271 lib tests, architecture 18/18, clippy and fmt clean.

saeedkolivand:
🎉 This is included in version 0.149.0.

PR #1125, opened 2026-09-07
ci: fix the chrome web store zip path and add per-store re-publish dispatches

saeedkolivand:
## Summary

Release run #569 (v0.149.0) was the first to reach the automated store submissions. Firefox AMO went through; the Chrome Web Store job failed before uploading anything:

```
Error: Zipped extension not found: …/apps/extension/extension-zips/ai-job-hunter-extension-chrome-0.149.0.zip
```

The zip existence check ran at the workspace root and passed, but `pnpm --filter @ajh/extension exec` runs the store CLI from `apps/extension/`, so the relative `--source` path resolved one directory too deep. The CLI now receives the path under `$GITHUB_WORKSPACE`, with a comment on why it must stay absolute.

**Per-store re-publish.** `publish-chrome` and `publish-firefox` are new `action` choices on the workflow dispatch. Each runs `package-extension` plus that one store's job for the chosen tag and nothing else. Both stores reject a version they already have, so recovering one store by re-running `build-installers` would only turn the other store's job red. `docs/DEPLOYMENT.md` § Browser extension store publishing points at the new actions.

Verified with actionlint (clean) and a YAML parse of the resulting job conditions.

## After merge

Dispatch **🚀 Release → `action: publish-chrome`, `version: 0.149.0`** to submit the Chrome version that run #569 could not. Nothing else in #569 needs re-running.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,10 @@ name: 🚀 Release
 #       and commits the bump back to main. Does NOT build installers.
 #   - build-installers (default) → build Tauri apps + browser extension for a chosen
 #       tag, then run the cask/manifest/download-page fan-out.
+#   - publish-chrome / publish-firefox → re-run ONE store submission for a chosen
+#       tag (package-extension + that store's publish job, nothing else). For when
+#       one store's job failed and the other went through: a store rejects a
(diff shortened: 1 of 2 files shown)


saeedkolivand:

Issue #1147, opened 2026-09-07
Dependabot auto-merge landed #1127 with a failing Tests check; require the test jobs before merge


saeedkolivand:
The dependabot PR #1127 (vitest group → 5.x) was merged into `main` although its own `🧪 Tests` check concluded **FAILURE**. Since that merge the root `pnpm test` fails on `main` (21 extension tests in `apps/extension/src/sidepanel/sidepanel.test.ts` and `src/background.test.ts`: listeners registered at module import are no longer captured under vitest 5), and because the pre-push hook runs the root suite, every branch's push is blocked until a hotfix lands.

**Severity**: high (process; blocks all pushes)
**Area**: repository automation — dependabot auto-merge flow and the required-checks set on `main`.

**Observed**

- `gh pr view 1127 --json statusCheckRollup`: `🧪 Tests: FAILURE`, `🤖 AI Review OK: SUCCESS`; PR state merged.
- `gh run list --branch main`: only Format Guard / Deploy Landing Page / Security / Quality ran on the merge commit; no test job gates the merge.

**Expected**

A dependabot PR that fails `🧪 Tests` must not auto-merge. Either add `🧪 Tests` (and the Rust test job) to the required status checks for `main`, or make the auto-merge workflow wait for those checks explicitly.

**Suggested fix**

- Branch protection / rulesets: require `🧪 Tests` (JS) and the Rust test job alongside `🤖 AI Review OK`.
- The auto-merge workflow (see the dependabot auto-merge workflow under `.github/workflows/`) should enable auto-merge only after the required checks are green, not on approval alone.
- A follow-up hotfix PR fixes the two extension test files for vitest 5.

saeedkolivand:
Checked the ruleset and the check runs on #1127: `✅ CI OK` is a required check and already `needs: tests`; on #1127 both `🧪 Tests` and `✅ CI OK` concluded FAILURE at 15:18 UTC, auto-merge was never armed, and the PR was merged manually at 18:50 UTC by the repository owner account, which the ruleset lets bypass every rule (`RepositoryRole` admin, `bypass_mode: always`). So no workflow or required-check change would have stopped this merge. The only mechanical fix is dropping the admin bypass from the ruleset, which also blocks the `gh pr merge --admin` flow this repo uses deliberately. Leaving this open for the owner to decide; no PR planned for it.

saeedkolivand:
[...]

PR #1148, opened 2026-09-07
test: keep import-time mock history under vitest 5

saeedkolivand:
## Why

Since #1127 moved the workspace to vitest 5, the root `pnpm test` fails on `main`: 21 tests in `apps/extension/src/background.test.ts` and `apps/extension/src/sidepanel/sidepanel.test.ts`. Vitest 5 enables `clearMocks` by default, which wipes every mock's recorded calls before each test. Those two files test entry-point modules with no exports: the listeners and mount calls under test are recorded once, at import time, so the history the assertions read back was already cleared. Every other project already runs on the new default without issue.

The pre-push hook runs the root suite, so this has been blocking every branch's push. Process side filed as #1147 (#1127 merged with its Tests check failing).

## What changed

Test-only. Each of the two files opts out for itself with `vi.setConfig({ clearMocks: false })` above its top-level dynamic import, with a comment citing the migration guide. The runner resets the config after each file, so no other project's mock semantics change; both files keep their explicit per-test resets, which is the vitest 4 behaviour they were written against. No assertion weakened, nothing skipped, extension runtime untouched.

Also carried here because the same bump broke it on `main`: the landing tech radar entry named "Vitest 4", which `check:tech-radar` (part of the Lint & Format job) now rejects against the workspace's `vitest ^5.0.0`. The entry is renamed to Vitest 5 with a fresh review date.

## Verification

- The two files: 96 passed. Whole extension suite: 22 files, 701 passed. Typecheck, eslint (`--max-warnings 0`) and prettier clean.
- Repo sweep: no other test file records module-load history at top level; the two `beforeAll` recorders elsewhere assert inside the test body and pass.

### DIFF
--- a/apps/extension/src/background.test.ts
+++ b/apps/extension/src/background.test.ts
@@ -124,6 +124,18 @@ vi.mock('./lib/bridge', () => ({
   }),
 }));
 
+// `background.ts` has no exports: every listener under test — `onMessage`,
+// `onInstalled`, `tabs.onUpdated`, `contextMenus.onClicked` — is recorded on
+// these mocks by the import BELOW, once, before any test body runs. Vitest 5
+// turned `clearMocks` on by default (a `vi.clearAllMocks()` before every test),
+// which wipes exactly that history — the migration guide names module-load
+// recording as the most affected pattern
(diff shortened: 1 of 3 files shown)


saeedkolivand:

PR #1153, opened 2026-09-08
fix(prefs): merge and write under one lock, and clear the country with its location

saeedkolivand:
## Why

The last commit of #1152 was still in the pre-push gate when that PR was merged, so `main` has the preference merge semantics without two review fixes that its resolved threads describe. This PR carries that one commit.

## What changed

- `JobPreferencesStore::update(merge)` reads, merges and writes inside one critical section, so two partial updates from the renderer or an agent can no longer merge from the same snapshot and erase each other's field; `job_preferences_set` runs its pure parse inside it. Concurrency tests: two threads merging different fields 200 times each never lose a write; a second update cannot complete while a merge is in flight; a failing merge leaves the row untouched. Reverting to the two-lock version fails both.
- A body that clears `location` alone now clears `countryCode` with it, matching the contract; a body that names `countryCode` keeps what it sent, and a new location never touches the stored country. Tests cover the lone-null, two-null and new-location shapes.

## Verification

Whole Rust lib suite (5389), `job_preferences` (40), `commands::job_preferences` (12), `architecture`, clippy `-D warnings`, fmt: green.


saeedkolivand:

PR #1174, opened 2026-09-08
ui(settings): say the app must stay open and repeat the transcript note on the agent card

### LINKED ISSUE #1156: Settings agent card should say the app must stay running and repeat the PII note
The Settings → Developer agent card is the copy-the-command surface most users ever see, yet it omits the two facts a first-timer needs: the desktop app must stay open for any agent call to work, and everything the agent reads lands in the AI client's transcript. Both facts are documented elsewhere (README / docs), just not where the user actually starts.

**Severity**: medium

**Area**

- Renderer copy: `apps/desktop/src/renderer/features/settings/components/preferences/AgentCliSection/index.tsx` and the `settings.developer.agentCli.*` keys in `packages/translations` (`en/translation.json`)
- Behaviour being described: agent CLI / MCP surface — `apps/desktop/src-tauri/src/extension_bridge/agent_cli/mcp.rs`, `agent_call.rs`, `agent_read.rs`, `agent_cli/policy.rs`; the underlying app commands under `apps/desktop/src-tauri/src/commands/**`

**Observed**

Reproduced against the installed app v0.149.0 on Windows 11 during an end-to-end run where agents drove the app through real AI clients and directly over stdio. The card renders exactly these strings:

> "This app's binary doubles as a command-line client and as an MCP server for coding agents. It is read-only unless you register it with a write tier."

and, for the default tier:

> "Reads your data and nothing else. This is what the server does with no flag."

Nothing on the card says the desktop app has to be open for any call to succeed — that sentence exists only outside the app. And "Reads your data and nothing else" reads as a privacy reassurance while the README's own warning (that the data read is copied into the AI client's transcript, PII included) is not repeated anywhere on the card. The copied command itself is just `<exePath>` plus flags, so the user pastes it into a client with no further context.

**Expected**

The card is the copy-button surface, so the two facts a first-timer must have before pasting the command belong on it: (a) keep the app open, (b) everything the agent reads is written into the AI client's transcript. Neither needs more than one sentence.

**User impact**

With the app closed, the freshly-pasted server fails on every call and the assistant reports a connection problem the user has no way to attribute — the natural conclusion is that setup failed, not that a window is shut. Separately, the tier line reads as a privacy guarantee: a user who takes "reads your data and nothing else" at face value can hand a hosted assistant their resume text, contact profile and a real application record without realising the content is retained in that client's conversation history.

**Suggested fix**

The card is the only onboarding surface most users see, so put both facts there in one sentence each: add "the app must stay open while the agent is connected" next to the copy button, and add the README's PII/transcript warning under the read tier so "reads your data and nothing else" is not the last word on privacy.

**Verification**

Reproduced by an independent run and judged against the design docs. The card's rendered strings were confirmed against the `t()` call sites in `AgentCliSection/index.tsx` and the `settings.developer.agentCli.*` entries in the English translation file: no privacy sentence, no app-must-be-running sentence.

Note: the closed issues #1128–#1144 were fixed in #1150/#1152; this copy gap is not covered by those and remains open.

saeedkolivand:
Adds two short notes to the Agent CLI settings card: the app must stay open while an agent is connected, and everything the agent reads is written into the AI client's own transcript. Both are gated on a resolved exePath so they never render before there's anything to copy or connect to. A follow-up fix reworded the stay-open note after review found it overclaimed how the CLI/MCP server relates to the app process.

Closes #1156

## What changed

- `apps/desktop/src/renderer/components/preferences/AgentCliSection/index.tsx` — added the stay-open note above the copy actions and the transcript note under the tier description, both gated on `exePath`.
- `packages/translations/src/locales/en/translation.json`, `packages/translations/src/locales/de/translation.json` — new `settings.developer.agentCli.stayOpenNote` / `.transcriptNote` keys; the stay-open note was later reworded in both locales to say calls reach the running app and fail while it's closed (matching the existing support FAQ), instead of the original "runs inside the app" framing (ADR-037, ADR-040 §11).
- `apps/desktop/src/renderer/components/preferences/AgentCliSection/AgentCliSection.test.tsx` — pins both notes' presence/visibility gating and the corrected stay-open wording across tiers and locales.

## Verification

- Pre-push gate (lint, typecheck, unit tests, `check:parity`, `check:workflow-catalog`, `check:agent-system`, `check:log-error-leaks`, `cargo check`, `cargo audit`/`cargo-machete`) — passed, run with `AJH_SKIP_CARGO_TEST=1` (the Rust test-binary rebuild step failed on an unrelated host-level file lock from an already-running local debug build of `ajh-tauri.exe`; this PR touches no Rust code).

## Follow-ups

None noted in the commit bodies.


saeedkolivand:
🎉 This is included in version 0.151.0.

PR #1196, opened 2026-09-10
ci(repo-charts): scope permissions to job level

saeedkolivand:
## Summary
- Moves `repo-charts.yml`'s `contents: write` + `actions: write` from workflow top-level to job-level (the workflow has only one job, so this is a zero-behavior-change scoping fix). Resolves code-scanning alerts #266/#267 (OpenSSF Scorecard Token-Permissions, flagged specifically for top-level write grants).

### DIFF
--- a/.github/workflows/repo-charts.yml
+++ b/.github/workflows/repo-charts.yml
@@ -44,14 +44,6 @@ on:
   release:
     types: [published]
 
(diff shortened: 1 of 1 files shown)


saeedkolivand:
🎉 This is included in version 0.153.0.

Issue #1197, opened 2026-09-10
ci(release): pin msstore-cli to a release tag and verify against an in-repo sha256


saeedkolivand:
## What

`publish-msstore` in `.github/workflows/release.yml` installs the Microsoft Store Developer CLI from `releases/latest/download/MSStoreCLI-linux-x64.tar.gz` and checks it against the `.sha256.txt` fetched from the **same** release. That verifies transport integrity only, never authenticity: a repointed or compromised upstream release replaces both files together, and `latest` is floating, so the binary that runs is not pinned.

Until the settings-dir fix (the PR that closes the `SellerId is not set` failure, upstream microsoft/msstore-cli#181) the job always failed before publishing. Now it runs to completion holding the real Store credentials, so this dormant supply-chain gap is live.

## Fix

- Download from `releases/download/v<x.y.z>/…` (pinned tag).
- Commit the expected sha256 for that tag next to the workflow and compare against the constant instead of the downloaded checksum file.
- Bump deliberately, the same way action SHAs are pinned.

Raised by the security critic on the msstore/snap PR; pre-existing, not introduced there.

saeedkolivand:
🎉 This is included in version 0.154.0.

PR #1198, opened 2026-09-10
fix(release): create the xdg data dir before msstore runs; surface the snap store listing

saeedkolivand:
## Microsoft Store auto-submit: root cause found and fixed

`publish-msstore` failed on every run at the `reconfigure` → `publish` handoff with `SellerId is not set` (filed upstream as microsoft/msstore-cli#181). It is not a CLI file-handling bug. Reproduced locally with the real v0.4.2 Linux binary and no credentials:

1. msstore-cli builds its settings dir from .NET's `Environment.GetFolderPath(LocalApplicationData)`.
2. On Unix that call returns an **empty string** when the resolved directory (`$XDG_DATA_HOME` or `~/.local/share`) does not exist yet (`SpecialFolderOption.None` semantics), so the settings dir becomes the **cwd-relative** `Microsoft/MSStore.CLI/`. That is exactly where earlier runs found `settings.json`.
3. `reconfigure` writes there, then stores the client secret through libsecret. gnome-keyring creates `~/.local/share/keyrings/` on that first store, so `reconfigure` itself brings `~/.local/share` into existence.
4. The next `msstore` process resolves `~/.local/share/Microsoft/MSStore.CLI/`, finds nothing, silently starts from an empty config, and reports `SellerId is not set`.

Fix: create the directory (0700) before the first `msstore` call. Verified in WSL: same binary writes cwd-relative with the dir absent and to `~/.local/share/…` with it present; the keyring dir appears only after the first secret store. An Opus refutation pass could not break the claim and found matching evidence in the earlier runner logs (a valid `./Microsoft/MSStore.CLI/settings.json` present in publish's own cwd, yet `File.Exists` false with no exception logged).

Also in this PR: a guard comment on the unquoted release-notes heredoc, and the copy-pasteable `--edge` command removed from release notes so the manual stable-promotion gate is not inverted by the notes.

## Snap Store listing surfaced

First snap went live today (`ai-job-hunter`, revision 2, `edge`). Linked from the README badge row, the release-notes downloads table (Linux row, store-first like Windows), the landing Linux card (a `store-btn`, not a `dl-btn`, so the positional installer-URL contract stays at 7), and `docs/DEPLOYMENT.md`, which now also lists the three places that must change when the snap is promoted to stable.

## Review

Sonnet author → Opus pr-reviewer (APPROVE, 1 MEDIUM + 1 LOW, both folded) + Opus root-cause refuter (not refuted, confidence 88) → Opus security critic (APPROVE; its advisories folded). Landing typecheck, tests, parity and drift checks pass locally and in the pre-push gate.

Follow-up filed: #1197 (msstore-cli is installed from `releases/latest` with a same-origin checksum; pin it now that the job actually runs with credentials).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1480,7 +1480,9 @@ jobs:
           # "SellerId is not set". Fix: create the dir first so both writes
           # land in the same place. Upstream fix: SpecialFolderOption.Create
           # in ConfigurationManager.cs.
-          mkdir -p "${XDG_DATA_HOME:-$HOME/.local/share}"
+          # 0700: this dir will hold settings.json (tenant/client/seller ids)
+          # and the keyring unlocked with a throwaway password.
+          mkdir -p "${XDG_DATA_HOME:-$HOME/.local/share}" && chmod 700 "${XDG_DATA_HOME:-$HOME/.local/share}"
 
           # reconfigure (writes the secret) and publish (reads it back) both
           # go through libsecret, so they must share the same live D-Bus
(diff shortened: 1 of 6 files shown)


saeedkolivand:

PR #1199, opened 2026-09-10
ci(release): pin msstore-cli to v0.4.2 and verify against a committed sha256

### LINKED ISSUE #1197: ci(release): pin msstore-cli to a release tag and verify against an in-repo sha256

saeedkolivand:
Closes #1197.

`publish-msstore` installed the Microsoft Store Developer CLI from `releases/latest` and checked it against the `.sha256.txt` published next to the tarball. A same-origin checksum only proves the download survived transit; it cannot tell a reviewed release from a repointed one, and `latest` floats. Now that #1198 lets the job run to completion with real Store credentials, that gap is live.

This pins the tag and the tarball's sha256 as two step-level env values, verified against a plain `sha256sum -c` line (the UTF-16LE `.sha256.txt` dance is gone with it). `v0.4.2` is the current latest release, so the running binary does not change. `curl -f` makes a bad tag fail as an HTTP error rather than a misleading checksum mismatch.

Verified locally: the committed digest matches both a fresh download of the pinned URL and Microsoft's published checksum; a wrong digest aborts the step under `set -euo pipefail`. Opus pr-reviewer: APPROVE (its one LOW, the `-f`, is folded). zizmor and the workflow-catalog check pass.

Bumping: change both env values together; the digest is `sha256sum` of that tag's `MSStoreCLI-linux-x64.tar.gz`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1421,7 +1421,7 @@ jobs:
           set -euo pipefail
           echo "::group::Installing msstore CLI $MSSTORE_CLI_VERSION"
           cd "$RUNNER_TEMP"
-          curl -sSLO "https://github.com/microsoft/msstore-cli/releases/download/$MSSTORE_CLI_VERSION/MSStoreCLI-linux-x64.tar.gz"
+          curl -fsSLO "https://github.com/microsoft/msstore-cli/releases/download/$MSSTORE_CLI_VERSION/MSStoreCLI-linux-x64.tar.gz"
           echo "$MSSTORE_CLI_SHA256  MSStoreCLI-linux-x64.tar.gz" | sha256sum -c -
           mkdir -p "$HOME/.local/msstore-cli"
           tar -xzf MSStoreCLI-linux-x64.tar.gz -C "$HOME/.local/msstore-cli"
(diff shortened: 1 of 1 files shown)


saeedkolivand:

PR #1200, opened 2026-09-10
ci(charts): count store installs in the public installs badge

saeedkolivand:
## One public number, labeled "installs"

The nightly `repo-charts` job now adds four store-side counters to the floor-corrected GitHub installer downloads it already computes, and the shields badge in the README is relabeled **installs**:

| Source | Metric | Auth |
|---|---|---|
| Microsoft Store | cumulative acquisitions (Partner Center analytics API) | existing `MSSTORE_*` secrets, client-credentials token |
| Snap Store | installed base on the latest day with a reading (`snapcraft metrics`) | existing `SNAPCRAFT_STORE_CREDENTIALS` |
| Chrome Web Store | listed users (scraped from the public listing, no API exists) | none |
| Firefox AMO | average daily users (public API v5) | none |

Every term is an install by any reading and the sum undercounts if anything, so the label is honest without publishing the per-store breakdown, which is tiny today. The parts are still published as `store-counts.json` on the `badges` branch (nulls preserved when a store is unreachable), and the landing `/download` page shows the same total under its lede, hidden until the file lands. The downloads-over-time chart and its history stay GitHub-only: their seed is GitHub-only and mixing units would draw a fake jump on the day the stores were added.

## Soft-fail contract

A store outage, a missing credential or a markup change can never break the badge: every fetcher catches everything and degrades to null, the MS Store pagination is capped and origin-pinned (a foreign next link returns null before any request, so the bearer token never leaves `manage.devcenter.microsoft.com`), snapcraft runs with a three-variable env and only its exit code is logged, and the snapcraft install step is soft-failing with its own timeout.

## What I could not verify locally

The two credentialed fetchers only run in CI. First nightly run (or a manual `📈 Repo Charts` dispatch) will show in its log whether the exported snap login carries the metrics ACL and whether the Entra app's Manager role is enough for the analytics endpoint. Both degrade to null, not a failure.

## Review

Sonnet author → Opus pr-reviewer + Opus security critic (both APPROVE after one fix round) → advisories folded by a Sonnet pass → scripts vitest (270), landing typecheck/tests (366)/build/parity, landing-drift and workflow-catalog checks all green, plus the pre-push gate.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -109,13 +109,15 @@ jobs:
           # It shipped missing exactly once, because repo-charts.yml builds the
           # badges tree from its own hardcoded file list: adding an output to
           # build-repo-charts.mjs publishes nothing until that list names it too.
-          if git show "FETCH_HEAD:downloads-by-platform.json" > apps/landing/public/downloads-by-platform.json 2>/dev/null \
-            && [ -s apps/landing/public/downloads-by-platform.json ]; then
-            echo "fetched downloads-by-platform.json ($(wc -c < apps/landing/public/downloads-by-platform.json) bytes)"
-          else
-            rm -f apps/landing/public/downloads-by-platform.json
-            echo "::warning::downloads-by-platform.json missing on the badges branch — /download ships without its count pills. Check the publish list in repo-charts.yml."
-          fi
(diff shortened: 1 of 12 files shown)


saeedkolivand:

PR #1241, opened 2026-09-22
fix(deeplink): accept the trailing slash windows adds to job deep links

### LINKED ISSUE #1237: desktop: Open/Generate/Prepare in the app deep links never navigate on Windows (ajh://open/?url= rejected)
## Summary

The extension's **"Open in app"**, **"Generate in the app"** and **"Prepare in the app"** links bring the desktop window to the front but never navigate. The app stays on whatever page it was showing. `ajh://settings/extension` ("Find my token") works on the same machine.

## Environment

Extension 0.154.0 (store build), desktop app 0.154.0 on Windows 11, Chrome. The machine has both the Microsoft Store package and the regular installer build installed. Windows resolves `ajh://` to the Store package, which hands the launch off to the already-running instance.

## Steps to reproduce

1. With the desktop app running, click **Open in app** in the extension's side panel on a tracked posting. Or run `start "" "ajh://open?url=<percent-encoded posting url>"` from a shell.
2. Watch the desktop window.

## Expected

The app navigates to that job's application page. With no match, it falls back to the Jobs list filtered by the URL, per `resolveJobDeepLinkTarget`.

## Actual

The window is shown and focused, and nothing else happens: no navigation, not even the Jobs fallback. This is deterministic, reproduced every time. `ajh://settings/extension` triggered the same way navigates correctly.

## Evidence

The command line the relaunched instance actually receives, read directly from the process:

```
<Store package>\ajh-tauri.exe ajh://open/?url=https%3A%2F%2Fjobs.smartrecruiters.com%2F...
```

Note the `/` inserted after `open`. The extension builds `ajh://open?url=…` (`apps/extension/src/sidepanel/sidepanel.ts`, `documents/documents.ts`, `prep/prep.ts`). Windows protocol activation canonicalises a URL whose path is empty by adding a trailing slash after the authority. `ajh://settings/extension` has a non-empty path, so it arrives unchanged, which is why it works.

## Root cause

1. **The slash-inserted form is rejected.** `apps/desktop/src-tauri/src/deeplink/mod.rs` `parse_job_url_target` does `rest.split_once('?')` and requires the action to be exactly `open` / `generate` / `prep`, so `open/` is rejected. `parse_one` then rejects it too (`rest.contains('?')`), so `parse_focus_target` returns `None`. The single-instance callback in `apps/desktop/src-tauri/src/lib.rs` has already called `tray::show_focus`, and `handle_deep_link(None)` does nothing. `deeplink/test.rs` only exercises the exact `ajh://open?url=` form.
2. **Cold start drops the intent (separate defect, same links).** When the app is *not* running, the cold-start block in `lib.rs` `setup()` calls `handle_deep_link` → `tray::dispatch_menu` before `tray::build` has managed `tray::PendingMenu` (`tray/mod.rs`). `try_state::<PendingMenu>()` returns `None`, the buffered intent is silently dropped, and the early emits fire before the renderer listens. The sibling `PendingFocus` (autopilot link) was already moved above that block for exactly this reason; `PendingMenu` was not. This affects all four `menu:navigate` deep links, including `settings/extension`, on a cold launch.
3. **No diagnostics.** The only log line on this path, `log::info!("[menu] dispatch …")` in `tray::dispatch_menu`, is under the `ajh_tauri::tray` target. That target sits below the global `Warn` level set in `lib.rs`'s log plugin builder, so a rejected link and a successful dispatch both leave nothing in the log.

## Suggested fix direction

- In `parse_job_url_target`, accept an optional single trailing `/` on the action (`open/?url=`, `generate/?url=`, `prep/?url=`). Keep every other rule (one `url=` param, length cap, http(s) with an authority). Add tests for the slash form and for a still-rejected `open//?url=` / `open/x?url=`.
- Manage `tray::PendingMenu` in `setup()` next to `PendingFocus`, before the cold-start deep-link block.
- Log a rejected `ajh://` argv at `warn` (scheme and action only, never the full URL) and raise the dispatch line to a logged level.
- Side note: `tauri_plugin_sentry` is registered before `tauri_plugin_single_instance`. That contradicts the in-code comment "Single-instance must be the FIRST plugin". It isn't the cause here, but worth aligning.

saeedkolivand:
Closes #1237.

## The bug

Windows protocol activation inserts a slash after the action when a deep link's path is empty, so the relaunched app receives `ajh://open/?url=…` while the extension sent `ajh://open?url=…`. `parse_job_url_target` accepted only the exact form, so the link was rejected, the window came to the front and nothing navigated. Captured from the live process on both the Store package and the installer build, so every Windows user hit it on **Open / Generate / Prepare in the app**. `ajh://settings/extension` carries a path, arrives unchanged, and always worked, which is what made this look random.

## What changes

- **Parser** — `parse_job_url_target` strips at most one trailing slash before matching the action allowlist. `open//?url=` and `open/x?url=` are still rejected, as are all non-allowlisted actions.
- **Cold start** — `tray::PendingMenu` is now managed in `setup()` beside `PendingFocus`, before the cold-start deep-link block. It used to be managed inside `tray::build`, which runs much later, so on a cold launch `dispatch_menu` wrote to missing state and every menu-intent deep link was dropped, `settings/extension` included.
- **Diagnosability** — the `ajh_tauri::tray` log target now logs at Info (its `[menu] dispatch` line sat below the global `Warn`, so even a successful dispatch left no trace), and each delivery path warns once on a rejected `ajh://` argv. The warning carries only the action segment, bounded to 32 characters with unsafe runs collapsed to `?`, so an attacker-controlled argv cannot inject control characters or flood the log.

## Tests

- `accepts_one_trailing_slash_after_the_action` — all three actions with the slash, plus the no-slash regression guard.
- `rejects_more_than_one_trailing_slash_or_a_real_path_segment` — `//` and `/x` per action, plus `opened`.
- `sanitize_action_for_log_collapses_control_chars_and_newlines` and `..._caps_an_over_long_action_at_32_chars`.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets --all-features -D warnings` (clean), the full Rust lib suite (5916 passed), the deep-link filter (18 passed), and the IPC codegen drift check. The pre-push gate passed, including cargo-deny and cargo-machete.

Not covered by a test: the cold-start ordering itself, which is a boot-sequence invariant with no harness. It was verified by reading the sequence in `setup()`.


saeedkolivand:

Issue #1166, opened 2026-09-08
Match a job by LinkedIn job id and host variants, not the byte-exact stored URL


saeedkolivand:
The agent `job` resource resolves a posting by comparing the caller's URL to the stored URL as a byte-exact normalized string, so the canonical LinkedIn URL copied out of a browser tab misses a job the app actually holds. The assistant then tells the user "the app has no record of this job" — a confident, wrong answer about the user's own data.

**Severity**: high

## Area

Agent CLI / MCP read surface and the URL identity helpers it leans on:

- `apps/desktop/src-tauri/src/extension_bridge/agent_read.rs` — `resolve_job` (the `==` comparison) and `job_lookup_key` (builds the caller-side key)
- `apps/desktop/src-tauri/src/applications/job_url.rs` — `normalize_job_url`, `decode_unreserved`
- `apps/desktop/src-tauri/src/scraping/scrape_url/mod.rs` — `canonical_job_url`
- `apps/desktop/src-tauri/src/extension_bridge/agent_cli/mcp.rs`, `agent_call.rs`, `agent_cli/policy.rs` — how the verb and its error reach an MCP client
- app-side owners of the stored URL under `apps/desktop/src-tauri/src/commands/**`

## Observed

From an end-to-end MCP run against the installed app v0.149.0 on Windows 11, driving `<exePath> agent mcp` from real AI clients and directly over stdio:

> Only the exact stored string matches. Working: the stored URL, and the stored URL with a trailing slash, and the stored URL with tracking query params (query is correctly stripped). Failing with `{"error":"no job found for this url","ok":false,"resource":"job"}`: `www.linkedin.com` host, the numeric-only `/jobs/view/<id>` form, `http://` instead of `https://`, and a scheme-less paste.

The failure is deterministic — identical output on two harness runs 12 s apart. Mechanically, `resolve_job` is a plain `==` between `normalize_job_url(decode_unreserved(stored))` and the caller key from `job_lookup_key`; `normalize_job_url` strips only a `www.` prefix and keeps the scheme verbatim, and `canonical_job_url` short-circuits to `None` for a LinkedIn URL that is already a `/jobs/view/` page, so no id-level identity is ever computed for the common case.

The refusal carries no `detail` — nothing tells the caller which spelling the app stored or where to obtain it.

## Expected

A posting is identified by the posting, not by the byte string the user happened to paste. Any of these should resolve to the same job:

[...]

PR #1242, opened 2026-09-22
fix(bridge): match a tracked job by posting identity when applied.check misses

### LINKED ISSUE #1214: extension: applied.check misses a tracked LinkedIn posting on a regional host redirect
## Summary

The extension's `applied.check` (and its batch form `applied.check.batch`, used for results-page stamping) looks up a tracked/applied posting by an exact normalized-URL string match. LinkedIn's own host redirects (e.g. a `de.linkedin.com` posting later opened as `www.linkedin.com`) produce two different normalized keys, so a posting that is genuinely already tracked shows up as untracked in the popup and side panel.

## Environment

Extension 0.154.0 (store build == repo main for `apps/extension` and `extension_bridge`), desktop app 0.154.0, Windows 11, Chrome.

## Steps to reproduce

1. Track/apply to a LinkedIn posting whose URL is stored under a regional host, e.g. `de.linkedin.com/jobs/view/<slug>-<id>`.
2. Open the SAME posting again later; LinkedIn's own redirect can land the tab on a different host, e.g. `www.linkedin.com/jobs/view/<slug>-<id>/`.
3. Open the extension popup, and separately the side panel, on that tab.

## Expected

The popup's page-context card and the side panel's stage strip should recognize the posting as already tracked, since it is the same job under a different LinkedIn host spelling.

## Actual

Both popup and side panel show the generic "Import the job on this page…" controls with no page-context card and no stage strip -- identical to a genuinely untracked posting. This is deterministic: it reproduces on 100% of host-spelling mismatches for LinkedIn regional/apex-host variants.

- `apps/desktop/src-tauri/src/extension_bridge/applied_check.rs:72-106` (`resolve_applied_check_url`) resolves `applied.check` via `canonical_job_url` -> `normalize_job_url` -> `ApplicationStore::find_by_job_url`, an exact `WHERE job_url = ?1` match (`apps/desktop/src-tauri/src/applications/mod.rs:485-495`).
- `apps/desktop/src-tauri/src/scraping/scrape_url/mod.rs:140-149` (`canonical_job_url`) returns `None` once the path already contains `/jobs/view/`, so no id-level rewrite happens for the common direct-posting case.
- `apps/desktop/src-tauri/src/applications/job_url.rs:102` (`normalize_job_url`) strips only a leading `www.` from the host, not other LinkedIn regional subdomains, so `de.linkedin.com/...` and `linkedin.com/...` (www-stripped) normalize to two different keys.
- `apps/desktop/src-tauri/src/extension_bridge/applied_check_batch.rs:16-24` reuses the same per-url resolver for `applied.check.batch`, so results-page stamping has the identical gap.
- The identity-aware fallback that already solves this class of problem -- `job_is_applied` / `scrape_url::job_identity` (`apps/desktop/src-tauri/src/extension_bridge/agent_read.rs:469-504`, `apps/desktop/src-tauri/src/scraping/scrape_url/identity.rs:43-64`), which folds any `*.linkedin.com` host and matches on the trailing numeric id -- was built for the agent/MCP `job` resource (closing a near-identical gap) but was never wired into `applied_check.rs`.

## Suggested fix direction

In `resolve_applied_check_url`, fall back to the identity-aware match (`job_identity`) already used by `job_is_applied` when the exact `find_by_job_url` lookup misses -- ideally by sharing one lookup helper across `applied_check.rs`, `applied_check_batch.rs`, and `agent_read.rs` so the three surfaces can never disagree about the same posting again.

saeedkolivand:
Closes #1214.

## The bug

`applied.check` matched a url by exact normalized-key equality, and `normalize_job_url` strips only a leading `www.` from the host. So a posting saved as `de.linkedin.com/jobs/view/<id>` never matched the same posting opened as `www.linkedin.com/jobs/view/<id>/`, and a slugged `/jobs/view/<slug>-<id>` never matched a bare numeric one. The popup's page-context card, the side panel's stage strip and the results-page stamps all showed a genuinely tracked job as untracked. Reproduced 4/4 during end-to-end testing.

## What changes

- The exact lookup stays the fast path and is untouched. On a miss, the url folds onto its `(board, id)` identity via `scrape_url::job_identity`, and the store returns the most recent application whose own url folds onto the same identity.
- That is the matcher `agent_read::job_is_applied` already used to close this same gap for the agent read surface (#1166); `applied_check` never picked it up. Both halves decode unreserved escapes before extracting, mirroring that function.
- A board `job_identity` does not cover behaves exactly as before, and an exact hit is unchanged.
- `applied.check.batch` reuses this resolver per url, so results-page stamping gets the fix for free.

## Performance

The store side scans a light `id, job_url, created_at` projection and loads only the winning row with `get(id)`. It does **not** materialise every application, which would pull job descriptions and parse the answers JSON per row — on a results-page stamp, most urls are untracked, so the miss path is the common case. The doc comments state that cost shape rather than asserting a timing.

## Tests

Six, mirroring `agent_read/tests.rs`'s `job_is_applied_*` family and using the existing store fixtures: regional host match, slugged vs numeric both directions, a different id must NOT match, a url with no identity behaves exactly as before, exact match still returns every field, and one pinning that the fallback returns the same fully-populated record the exact path does (full JSON equality), so a later refactor cannot quietly return a half-built row.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets --all-features -D warnings` (clean), full Rust lib suite (5918 passed, 0 failed), and the pre-push gate including cargo-deny and cargo-machete.


saeedkolivand:

Issue #1226, opened 2026-09-21
extension: Fill's first-time confirmation dialog claims it fills the résumé file, but Fill never touches file inputs


saeedkolivand:
## Summary
The first-time Fill confirmation dialog's copy says "This will fill: name, email, phone, résumé file." — but Fill's `FILLABLE_TYPES` only ever fills `text | email | tel | url` inputs and explicitly excludes `type=file`. Fill can never attach a résumé file; that's what the separate Documents-tab "Attach" action is for. The confirmation the user is asked to approve overstates what the action will actually do.

## Environment
Extension 0.154.0 (store build), desktop app 0.154.0, Windows 11, Chrome. Reproduced on a real Greenhouse application form.

## Steps to reproduce
1. Open the side panel on a job application page with a résumé file-upload field, on a host that has never been confirmed.
2. Click "Fill this form" in the Job tab.
3. Read the confirmation dialog copy, then click Fill and check the file input afterward.

## Expected
The dialog only promises what Fill actually does (name/email/phone-type text fields).

## Actual
Dialog text: `"This will fill: name, email, phone, résumé file. Nothing is submitted — you review and press the site's own button."`
After confirming and filling, the page's `resume` and `cover_letter` file inputs were both still empty (`files.length === 0`) — Fill filled 6 text/email/tel fields (first name, last name, email, phone, location, LinkedIn) and never touched either file input, as expected from the code. The dialog's "résumé file" claim never happens via this action.

## Root cause
- `apps/extension/src/lib/site-memory.ts` — `DEFAULT_FILL_CONFIRM_COPY` includes "…résumé file."
- `apps/extension/src/lib/autofill.ts:111` — `FILLABLE_TYPES` (`text | email | tel | url`) has no `file` entry; file inputs are never candidates for Fill.

## Suggested fix direction
Drop "résumé file" from `DEFAULT_FILL_CONFIRM_COPY`, or split it into two distinct sentences/dialogs if the intent is eventually to also mention that a separate Attach action exists for the résumé file.

_Found during an end-to-end test of the published 0.154.0 extension against the 0.154.0 desktop app._

saeedkolivand:
🎉 This is included in version 0.155.0.

Issue #1219, opened 2026-09-21
extension: check-fit's reload advice is wrong on Web Store and PDF-viewer pages


saeedkolivand:
## Summary
Clicking "Check fit" on a page Chrome permanently blocks from content-script access (e.g. the Chrome Web Store) or whose rendered content the desktop app can never turn into job text (Chrome's built-in PDF viewer) shows an error message telling the user to reload the page. Reloading never helps in either case — the failure is structural, not transient — so the message gives actionable-sounding advice that cannot work.

## Environment
extension 0.154.0 store build, desktop app 0.154.0, Windows 11, Chrome

## Steps to reproduce
1. Open the extension popup on a Chrome Web Store page (`chromewebstore.google.com/...`) and click "Check fit".
2. Separately, open a PDF that Chrome renders in its built-in viewer and click "Check fit" in the popup.

## Expected
Since Chrome permanently disallows content-script injection on the Web Store, and its native PDF viewer exposes no extractable job-posting text, the failure message should not suggest that reloading will fix it.

## Actual
- Chrome Web Store: "Could not read this page. Reload the job page and try again."
- PDF viewer: "Could not read this job posting. Reload the page and try again."

Two different fixed strings for what is functionally the same class of failure (a page that can never be scripted or parsed), both implying reload is a fix. Reproduced consistently on both page types.

## Evidence
- "Could not read this page. Reload the job page and try again." — shown when the content-script capture itself fails.
- "Could not read this job posting. Reload the page and try again." — shown when capture succeeds but no job text can be extracted server-side; forwarded to the popup unchanged.

## Root cause
- `apps/extension/src/background.ts:1406-1423` (`runMatchLive`) always returns the fixed "reload the job page" text on any `captureTabHtml` failure, with no check for hosts Chrome permanently excludes from scripting.
- `apps/desktop/src-tauri/src/extension_bridge/match_live.rs:89-95,286-297` (`NO_JOB_TEXT_MESSAGE`) always suggests reload when no job text can be extracted, even for a page (PDF viewer) that will produce the same empty result on every reload.
- `apps/extension/src/job-tools/job-tools.ts:239-245` (`resolveMatchLiveResponse`) forwards both server- and client-side messages to the popup verbatim, with no distinction for permanently-restricted page types.

## Suggested fix direction
[...]

PR #1243, opened 2026-09-22
fix(extension): stop offering x/twitter fields as questions, correct two misleading messages

### LINKED ISSUE #1218: extension: X/Twitter link fields are captured as AI-draftable Answers questions
## Summary

The Answers-tab field matcher only recognizes LinkedIn, GitHub, and portfolio/website links as identity/link fields. Any other social-link field — for example a bare "X" label for a Twitter/X profile handle — falls through every exclusion gate unrecognized and is captured as a genuine, AI-draftable application question, with the full tone/length/regenerate "Draft this answer" UI, on a field that is really just a handle or URL.

## Environment

extension 0.154.0 (store build), desktop app 0.154.0, Windows 11, Chrome

## Steps to reproduce

1. Open an ATS application form (confirmed on Ashby) that has GitHub, LinkedIn, and X (Twitter) profile-link fields.
2. Open the extension side panel → Answers tab → Rescan.
3. Expand the row whose label is exactly "X".

## Expected

The X field is recognized as an identity/link field and excluded from the Answers question list, the same way the GitHub and LinkedIn fields on the same form already are — or, at minimum, it is not offered the full AI-drafting treatment, since there's nothing to compose.

## Actual

The field labelled "X" is captured as a full question, with TONE (Warmer/Formal/Simpler/…) and LENGTH chips and a "Draft this answer" button — identical to a genuine essay question like "Tell us more". Reproduced on 2/2 scans of the same form (initial scan and after Rescan).

## Root cause

- `apps/extension/src/lib/field-signal.ts`'s `NAMED_KEY_PATTERNS` table (~line 528) maps only `linkedin`, `github`, and `website` (portfolio/personal site) to a recognized identity key — there is no `twitter`/`x` pattern.
- The `AMBIGUOUS` denylist and its siblings (`field-signal.ts`, ~lines 34-223) have no protection for a bare "X"/"twitter" token either; a bare `x` can't safely be added as an unanchored substring (it would match "experience", "text", "next", etc.) — it needs a `\b`-anchored entry, the same technique already used for other short tokens in `AMBIGUOUS_WORDS`.
- `apps/extension/src/lib/answers-capture.ts`'s `isCapturable` (~line 95) only excludes a field via `matchAutocompleteKey`, `isAmbiguousSignal`, or `matchNamedKey`; none of those fire for a bare "X" label, and the ATS renders the field as a plain `type=text` input, so `CAPTURABLE_INPUT_TYPES` (~line 82) doesn't filter it either.
- This isn't a case of an existing mapping being intentionally left out of the capture path — the app's own Contact Profile wire shape (`{ fullName, email, phone, location, linkedin, github, website }`) has no twitter/x field at all, and no test or comment in either file mentions twitter/x, confirming the gap was never considered rather than deliberately scoped out.

## Suggested fix direction

Add a word-boundary-anchored twitter/x row to `NAMED_KEY_PATTERNS` (e.g. matching `x`/`twitter` as a standalone token, careful not to collide with unrelated short labels) so the field resolves to a recognized key and is excluded by `isCapturable` the same way linkedin/github/website already are. This key doesn't need a corresponding Contact Profile field or autofill wiring — it only needs to make the capture-exclusion path treat it as a link field instead of a genuine open-ended question.

saeedkolivand:
Closes #1218, #1219, #1226. Three small extension-side defects from the end-to-end test round, batched because each is a few lines.

## #1218 — X/Twitter fields were offered as AI-draftable questions

The field matcher knew `linkedin`, `github` and `website` as identity links, so those are excluded from captured answer questions. A field labelled `X` matched nothing and was captured as a genuine essay question with the full tone/length/"Draft this answer" UI.

- `field-signal.ts` gains a `twitter` key: `twitter` as a substring, plus a label that is exactly `x` or `x` with a handle-ish qualifier (`x handle`, `x username`, `x profile`, `x url`, `x link`, `x id`, `x-handle`). A stray `x` elsewhere is not matched — `mac os x experience`, `x experience`, `x ray`, `xray` and `do you use x` are all tested negatives.
- The contact profile has no twitter slot, so `valueForKey` returns `''`. Rather than letting the key claim-and-drop the field (which would have **stopped filling X fields that previously filled from a matching profile link**), `twitter` joins `website` in `EXTRA_LINK_FALLBACK_KEYS`, so an extra link labelled X or Twitter still fills it. That regression was caught in review; the test for it was checked to fail without the fix.

## #1219 — "reload the page" advice on pages that can never be read

`Check fit` answered every capture failure with reload advice, including on the Chrome Web Store and Chrome's built-in PDF viewer, where injection is permanently blocked. `Stamp this results page` had a third, differently-worded variant of the same lie.

- `isPermanentlyUnreadablePage` detects browser-internal schemes, the extension origins, the built-in PDF viewer, `.pdf` paths and the Web Store host, and also folds in the empty url Chrome hands back for a restricted tab (the extension deliberately holds no `tabs` permission).
- Both gestures short-circuit those with one shared message, and keep a reload hint only for a transient failure on a page that genuinely can be read. Three strings collapse to two shared constants.
- Both gestures now resolve the active tab exactly once per request, preserving the single-snapshot property from an earlier review finding.

## #1226 — the Fill confirmation promised something Fill cannot do

The first-time confirmation said Fill would fill "name, email, phone, résumé file". Fill only touches `text | email | tel | url` inputs and explicitly skips file inputs; attaching is the Documents tab's separate action. The copy now names what Fill actually fills.

## Verification

`pnpm -F @ajh/extension lint`, `typecheck` and `test` all clean: 35 files, 1054 tests passing (up 19 from the new branches). Pre-push gate passed.


saeedkolivand:

PR #1244, opened 2026-09-22
fix(extension): act on the requesting window's tab, not the last-focused one

### LINKED ISSUE #1215: extension: popup/panel actions can hit another Chrome window's tab (Import writes the wrong job)
## Summary
The side panel's action handlers (Check fit, and by the same code path Import/Fill/Save answers) resolve "the active tab" inside the background service worker via `browser.tabs.query({ active: true, currentWindow: true })`. A service worker has no window of its own, so Chrome's `currentWindow` falls back to whichever window last had OS focus — not necessarily the window the side panel instance is actually docked to. When a second Chrome window has focus, the panel's action buttons read (or fail to read) the wrong window's tab instead of the tab the panel is following.

## Environment

## Steps to reproduce
1. Open the side panel in Chrome window 1 on a job posting; the panel correctly shows it is following that tab.
2. Open a different job posting in a new Chrome window (window 2), which becomes the OS-focused window. Do not interact with window 1 again.
3. Without refocusing window 1, click "Check fit" in window 1's still-open panel.

## Expected
The panel lives in window 1 and is following window 1's tab, so "Check fit" should act on window 1's tab regardless of which window currently has OS focus.

## Actual
The panel's tab caption degrades to a bare hostname instead of the tracked posting's title, a spurious empty "Form" group appears, and the action fails with "Could not read the current tab URL." Refocusing window 1 and retrying succeeds immediately with the correct result. Observed twice in a row while window 2 held focus; not reproduced once window 1 regained focus. This is a multi-window edge case, not something a single-window user will hit.

## Evidence
- Thrown error text: `Could not read the current tab URL.`
- Panel caption after the failure: `Reading: www.linkedin.com` (bare host, not the tracked posting's title)

## Root cause
`apps/extension/src/background.ts` resolves "the active tab" via `browser.tabs.query({ active: true, currentWindow: true })` in 13 separate helpers (e.g. `activeTabUrl` at background.ts:449, `captureActiveTabHtml` at background.ts:460, `activeTabIdAndUrl` at background.ts:492, used by `runMatchLive` at background.ts:1406 for Check-fit). Running in the background service worker, `currentWindow` there resolves to the last OS-focused window, not the side panel's own window. The side panel already tracks its own window id correctly for its display/follow logic (`panelWindowId`, `apps/extension/src/sidepanel/sidepanel.ts:474`, queried with `windowId: panelWindowId` at sidepanel.ts:489) but never sends that id along with action requests (`apps/extension/src/lib/messages.ts`'s `PopupRequest` carries no window id field), so the background handlers have no way to target the right window.

## Suggested fix direction
Add an optional `windowId` to the relevant `PopupRequest` messages and have the side panel pass its own `panelWindowId`. Have the affected `background.ts` helpers accept an optional `windowId` and query `{ active: true, windowId }` when supplied, keeping `currentWindow: true` as the fallback for the popup (which is always opened in the focused window by construction).

saeedkolivand:
Closes #1215.

## The bug

The background is a service worker, so it has no window of its own, and `tabs.query({ active: true, currentWindow: true })` there resolves to whichever window the browser focused **last** — not the window whose popup or side panel sent the request.

With a second window focused, a gesture in the first window's surface acted on the other window's active tab. Reads failed with a confusing "Could not read the current tab URL". **Import was worse**: during the end-to-end test round it imported an unrelated page from the other window — a GitHub repository page — created a real saved application from it, and reported success. A later "I already applied" click then moved that stray row to *applied*. It reproduced 2/2 once it started, after 8 correctly-targeted imports in the same session, which is what makes it dangerous: it passes spot checks and corrupts data occasionally, with no error.

## The fix

Both surfaces already know their own window, so they send it and the background targets it:

- `PopupRequest` is now `SurfaceWindow & (…)`, carrying an optional `windowId`.
- The popup resolves its window once, via a single shared promise awaited by every `send`, so concurrent requests still reach the background in the order they were issued. A per-send lookup would let a later request overtake an earlier one.
- The panel sends its existing `panelWindowId`.
- The background gets one seam, `activeTabIn(windowId?)`, replacing all 13 scattered `tabs.query` calls, and the id is threaded explicitly through the handlers to the helpers.

**No id means no change**: the previous last-focused-window behaviour is kept, so context-menu and auto-track paths, which carry their own tab or none, are untouched.

The id is threaded through parameters rather than held in a module-level "current request" variable on purpose: requests from two windows are concurrent in a worker, and shared state would mistarget under exactly the conditions this fixes.

## Two bugs found while verifying this change

- **A temporal-dead-zone error I introduced**: `panelWindowId` was declared well below `send`, but the panel's auto-save notice sends at module load, so reading the binding threw — and the notice's own `catch` swallowed it, silently killing the banner. Caught by its tests; the declaration moved above `send` with a comment saying why.
- **An order-coupled test**: entering `connected` fires `appliedCheck`, `fieldsProbe` and `answerScan` together, and the stale-response test queued mock responses by call order, so one extra microtask handed check B's response to a sibling request. It now answers by request kind, which is what the test was actually about. The popup's own staleness guard is generation-based and was never affected.

## Tests

New `#1215` block in `background.test.ts` with a fake `tabs.query` that answers `{ active: true, windowId }` and `{ active: true, currentWindow: true }` **differently** — without that distinction the bug could not be exercised at all. It covers import (asserting the url on the wire, the data-corruption path), fill, matchLive, the no-`windowId` fallback, and the submit watcher.

Mutation-checked: removing the threading from the import path alone makes the import test fail.

## Verification

`pnpm -F @ajh/extension lint`, `typecheck`, `test` (35 files, 1059 tests) and the repo-wide `eslint . --max-warnings 0`, plus the pre-push gate.


saeedkolivand:

PR #1245, opened 2026-09-23
fix(extension): escalate bridge backoff after a failed handshake

### LINKED ISSUE #1216: extension: bridge reconnect never escalates backoff after a connect-then-fail handshake
## Summary

After the extension's bridge client opens a transport (native messaging or the loopback WebSocket) but then fails the v2 HMAC handshake, every reconnect attempt retries at the backoff ladder's floor (500ms) forever instead of escalating up to the documented 10s cap. The exponential backoff exists in code (`BACKOFF_MS = [500, 1000, 2000, 5000, 10000]`) but never actually escalates for this failure class, because the retry counter is reset back to zero at the start of every connection attempt, before that attempt's outcome is known.

## Environment

## Steps to reproduce

1. Pair the extension with the desktop app (a valid pairing token is stored).
2. Regenerate the pairing token in the desktop app (Settings, or a factory reset) while the extension's browser is closed or its service worker is asleep, so the live `token.revoked` frame is never delivered to it (this is the already-documented "offline at rotation" case).
3. Reopen the browser / let the service worker wake. The extension reconnects with the now-stale token: the desktop bridge accepts the transport connection, rejects the HMAC proof, and closes the socket without a reply (the deliberately silent, ambiguous close the protocol uses for a bad proof).
4. Observe the reconnect timer interval and/or the desktop app's `extension_bridge` warning logs over the next several minutes.

## Expected

Each reconnect attempt after a connect-then-fail handshake should back off further up the ladder (500ms -> 1s -> 2s -> 5s -> 10s) and settle at roughly one attempt every 10 seconds.

## Actual (with frequency)

Every reconnect after a connect-then-fail handshake retries at the ladder's floor, ~500ms, indefinitely -- it never advances to the higher rungs. This reproduces on the very first retry cycle and on every one after it (deterministic in code). It is consistent with real desktop-log bursts showing dozens of connection/handshake warning lines within a single minute, sustained for roughly an hour at a time, rather than the sparse, decreasing-frequency pattern a working backoff would produce.

## Evidence

Desktop `extension_bridge` warning log pattern during an affected window (sanitized, no tokens/paths):

```
[WARN] [extension_bridge] read error: WebSocket protocol error: Connection reset without closing handshake
[WARN] [extension_bridge] handshake: client proof failed constant-time verification
```

These repeat at intervals consistent with a fixed ~500ms retry rate rather than an escalating one, including runs of the client-proof warning only about a second apart -- the ladder's floor, never its higher rungs.

## Root cause

`apps/extension/src/lib/bridge.ts`: `attach()` (~line 2101) unconditionally resets `this.backoffIndex = 0` at line 2103, on every connection attempt, before that attempt's handshake result is known. Any handshake failure that reaches `finishHandshake('app_not_running')` (the timeout/closed/ambiguous-silence branches inside `performHandshake()`, ~lines 2200-2262) then calls `scheduleReconnect()` (~lines 2695-2703), which reads `backoffIndex` -- already reset to 0 moments earlier by that same `attach()` call -- so it always resolves to `BACKOFF_MS[0]` (500ms, line 89). `scheduleReconnect()` does increment the index afterward, but the next retry cycle calls `attach()` again, which resets it to 0 before the next `scheduleReconnect()` call runs, discarding the increment every time. (The sibling failure path in `doConnect()`, where no transport can even be opened at all, calls `scheduleReconnect()` directly without going through `attach()` and backs off correctly -- only the "transport opened, handshake then failed" class is affected.)

## Suggested fix direction

Move the `backoffIndex` reset out of `attach()`'s entry and into the handshake's actual success path only: inside `performHandshake()`'s success branch (alongside setting `authenticated = true`) and in `attach()`'s no-token early return (which also reaches `connected` without a handshake). That lets `scheduleReconnect()`'s increment persist across repeated connect-then-fail cycles so the existing `BACKOFF_MS` ladder escalates as designed. Add a test that drives two consecutive connect-then-fail handshake cycles and asserts the second reconnect waits longer than the first, since the current suite only ever exercises a single cycle.

## Related UX effect

During such a storm the popup shows "App not running" (with a Retry button) for as long as the loop lasts, although the desktop app is running. The silent close on a bad proof is deliberately ambiguous (so it cannot act as a token oracle), and the client maps it to `app_not_running`. With a working backoff this would be a short blip; with the stuck 500 ms loop it becomes the steady state. A bounded-retry hint ("re-pair?") after N consecutive connect-then-fail cycles, already a recorded follow-up under "Revocation reach" in `docs/knowledge/extension-domain.md`, would give the user a way out.

saeedkolivand:
Closes #1216.

## Problem

The desktop bridge accepts the transport **before** it judges our HMAC proof, and the `Unauthorized` path closes the socket without a reply. So a stale pairing token produces an endless connect-then-silent-close cycle — and `attach()` reset `backoffIndex = 0` at the top, before that attempt's outcome was known. The ladder re-armed at its floor every time: one probe every 500ms, forever, instead of settling at the documented 10s cap.

## Fix

Reset the ladder only where the connection is genuinely usable:

- the no-token attach that reaches `connected` (a working, unpaired socket), and
- a completed mutual handshake (`authenticated = true`).

The `revoked` branch keeps its own deliberate reset — that path re-probes exactly once by design.

## Verification

- New regression test drives two consecutive connect-then-fail cycles and asserts the second waits at `BACKOFF_MS[1]`, not the floor.
- **Mutation-checked**: restoring `backoffIndex = 0` in `attach()` fails the new test.
- `pnpm vitest run` in apps/extension: 1060 passed (35 files).
- `pnpm typecheck` clean; `pnpm lint:strict --max-warnings 0` clean.


saeedkolivand:

PR #1246, opened 2026-09-23
fix(extension): tell a dead api key apart from a retriable draft failure

### LINKED ISSUE #1217: desktop: draft-answer failure always says "please retry" even for a dead API key
## Summary

Drafting an AI answer in the extension's Answers tab always shows the same fixed message, "Could not draft an answer. Please retry.", for every compose failure -- including a provider auth/config error (e.g. an invalid API key returning 401) that retrying can never fix. The desktop backend already classifies the underlying error as non-retriable (`AppError::Config`, for which `AppError::retriable()` returns `false`) before it reaches the compose path, but `answer_assist`'s failure handling discards that classification and always emits the same generic sentinel. The user is repeatedly told to do something guaranteed to fail, and -- because the provider's daily budget is charged before each attempt -- every retry the message invites also spends one unit of that limited quota on a call that cannot succeed.

## Environment

extension 0.154.0 store build, desktop app 0.154.0, Windows 11, Chrome

## Steps to reproduce

1. Configure the AI provider with an invalid/unauthorized API key (any call returns 401/403).
2. Open the extension side panel, go to the Answers tab on an application form, and click "Draft this answer" on any row.

## Expected

A non-retriable configuration/auth failure (bad API key, expired credentials) should be told apart from a transient generation hiccup -- e.g. pointing the user at Settings -> AI Provider -- rather than showing text that literally invites a retry guaranteed to fail identically.

## Actual

The row shows the exact same generic text every attempt: "Could not draft an answer. Please retry." The real, actionable reason (a provider-reported "invalid or unauthorized API key" message) only reaches the desktop app's log at WARN level and never reaches the extension UI. Reproduced twice, byte-identical failure and message both times, roughly 40 seconds apart.

## Root cause

- A provider 401/403 is mapped to `AppError::Config(...)` (`apps/desktop/src-tauri/src/commands/ai_provider/mod.rs:1291`), which the shared error type already classifies as not retriable (`AppError::retriable()`, `apps/desktop/src-tauri/src/error.rs:82-85`, true only for `Network`/`RateLimited`).
- That typed error survives unchanged through the compose stream (`apps/desktop/src-tauri/src/extension_bridge/stream.rs:730-745` returns the underlying `Err(e)` as-is).
- `compose_attempts` (`apps/desktop/src-tauri/src/extension_bridge/answer_assist.rs:1027-1067`) passes every non-length-cut compose failure through `to_draft_failed` (lines 1039, 1065), which (lines 170-173) logs the real cause and unconditionally replaces it with the fixed `DRAFT_FAILED_MESSAGE` (line 154) -- with no branch on error kind.
- `charge_compose_budget` (lines 905-909) charges the provider's daily ceiling before each attempt, so every doomed retry the message invites also burns quota.

## Suggested fix direction

Before collapsing in `to_draft_failed`, branch on the error's retriability (`AppError::retriable()`/kind) and return a second fixed sentinel for non-retriable compose failures (e.g. pointing at Settings -> AI) instead of the generic "please retry" text -- mirroring the existing `NO_PROVIDER_MESSAGE` vs `AI_ASSIST_OFF_MESSAGE` split (ADR-044 decision 8). This reuses classification the codebase already computes (`AppError::retriable()`/`code()`) and keeps both messages as fixed, no-dynamic-content strings, so the "nothing dynamic reaches the wire" invariant is untouched.

saeedkolivand:
Closes #1217.

## Problem

`to_draft_failed` collapsed **every** downstream compose error into one sentinel ending "Please retry." — including a provider 401/403, which `commands::ai_provider::friendly_api_error` maps to `AppError::Config` and which `AppError::retriable()` already classifies as false. The user was told to do the one thing guaranteed to fail, and since the daily provider budget is charged before each attempt, every invited retry also burned a unit of that quota on a doomed call.

## Fix

Branch on the typed error inside `to_draft_failed` and return a second fixed sentinel that points at Settings → AI. The match is narrow on purpose — `AppError::Config` only, not `!retriable()`, which would sweep in `Parse`/`Provider`/`Timeout`/`Storage` failures that are genuinely worth retrying and are not a settings problem.

Still a fixed string, so the wire-error discipline is untouched: the provider's raw error text never reaches the extension.

## Verification

- New test asserts a `Config` error maps to the new sentinel, is **not** the generic one, and still carries no provider name.
- `cargo test --lib answer_assist`: 81 passed, 0 failed.
- `cargo fmt --check` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean.


saeedkolivand:

PR #1247, opened 2026-09-23
fix(extension): clear a stale fit badge, read trailing-currency salaries, drop junk chips

### LINKED ISSUE #1221: On-page fit badge goes stale after an in-page SPA job-to-job navigation
## Summary

The on-page fit badge ("AI Job Hunter fit: N% — ...") that "Check fit" paints in the bottom-right corner of a job posting is never cleared or refreshed when the user moves to a DIFFERENT posting via a client-side (SPA) navigation, such as selecting another card in LinkedIn's job-search results without a full page reload. The badge keeps showing the FIRST posting's score, résumé match, and missing-keyword chips while the visible page has completely changed to a different job.

## Environment

Extension 0.154.0 (store build == repo main), desktop app 0.154.0, Windows 11, Chrome. LinkedIn job-search results (`linkedin.com/jobs/search/...`), which routes card selection client-side without a full navigation.

## Steps to reproduce

1. On a LinkedIn job-search results page, open the popup on the first result and click "Check fit". The on-page badge appears bottom-right with that posting's score (e.g. "27% low match").
2. Without reloading, click a DIFFERENT job card in the results list on the left. LinkedIn updates the URL's `currentJobId`, the page's `<h1>`/company/description, and the whole right-hand detail pane to the new posting — client-side, no navigation event the extension listens for.
3. Look at the on-page badge again.

## Expected

The badge should disappear (or at least clearly indicate it no longer matches what's on screen) once the user is looking at a different posting than the one it was computed for.

## Actual

The badge is still present, unchanged, still showing the FIRST posting's exact score/band/résumé-name/missing-keywords, now overlaid on a completely different job's page. Nothing in the UI hints that the number is stale. Reproduced 2/2 tries (tested Harvey → Suno, and again after re-checking Harvey).

## Evidence

- Screenshots: before navigation, `<h1>` = "Staff Software Engineer, Frontend" (Harvey), badge = "27% low match"; after clicking the Suno card (no reload), `<h1>` = "Senior / Staff Software Engineer, Frontend" (Suno), URL `currentJobId` changed from `4432888947` to `4463942585`, badge STILL reads "27% low match" (Harvey's number).
- Confirmed via `document.getElementById('ajh-fit-badge')` still present with the same content both before and after the SPA navigation — the host node is never removed or repainted.

- `apps/extension/src/background.ts:1341-1361` (`injectFitBadge`) only guards against painting the badge for the WRONG posting on arrival (`if (location.href !== expectedUrl) return;` at `background.ts:1346`) — it never runs again to re-check or clear an EXISTING badge once the tab's own URL changes later.
- `apps/extension/src/background.ts:2432-2437` (`browser.tabs.onUpdated` listener) only reacts `if (changeInfo.status !== 'loading')`, i.e. to real navigations. LinkedIn's job-to-job card selection is a same-document `history.pushState`/`replaceState` update, which Chrome does not reliably surface through `tabs.onUpdated` the same way — so nothing in the background script ever learns the user moved to a different posting and nothing calls back into the content script to remove `#ajh-fit-badge`.
- The badge is only ever cleared as a side effect of a NEW successful "Check fit" call repainting it (`apps/extension/src/lib/fit-badge.ts:74`, "always clears any prior badge first") — there is no independent navigation-driven cleanup path.

## Suggested fix direction

Have the content-script side that owns the badge also watch for same-document navigation (e.g. wrap `history.pushState`/`replaceState`, or poll `location.href` on a short interval / `MutationObserver` on the title) and remove `#ajh-fit-badge` as soon as the URL changes out from under it, independent of whether a new "Check fit" ever gets triggered for the new posting.

### LINKED ISSUE #1222: Check fit: German trailing-currency-symbol salary ("60.000 € – 75.000 €") never detected
## Summary

"Check fit"'s salary-fact extraction (`salary.posting`, shown as "Posting says …" in the popup/panel/on-page badge) requires a currency symbol or ISO code to come BEFORE the first number of a range. German job postings very commonly write amounts with the currency symbol AFTER each number instead ("60.000 € – 75.000 €"), which never matches. The UI then shows no salary line at all — identical to a posting that genuinely lists no salary — even though the page states one clearly.

## Environment

Extension 0.154.0 (store build == repo main), desktop app 0.154.0, Windows 11, Chrome.

## Steps to reproduce

1. Open (or view) a German job posting that states its salary as "NN.NNN € – NN.NNN €" (symbol trailing each number, which is the common convention on German job boards, vs. the leading-symbol convention on English-language/US boards).
2. Click "Check fit" and expand "why?" in the result card.

## Expected

The card should show a "Posting says 60.000 € – 75.000 €" line, the same way it shows "Posting says $231,000 - $340,000" for a US posting with symbol-first phrasing.

## Actual

No salary line appears anywhere in the result (popup, panel, or on-page badge) — the UI is indistinguishable from a posting that has no salary information at all. Reproduced twice with a minimal, realistic posting page containing exactly the line "Je nach Erfahrung bieten wir ein Jahresgehalt von 60.000 € – 75.000 € brutto." (a 50%-match posting, so the fit card rendered normally, just with no salary fact).

## Evidence

- Screenshot: the expanded "why?" panel for this posting shows missing-keyword chips ("entwickler", "m/w/d", "testjobsde") and "Résumé used: …" but no "Posting says …" line, despite the page's own text containing an unambiguous, correctly-formatted salary range.
- For comparison, the SAME session correctly extracted and displayed "Posting says $231,000 - $340,000" for a US LinkedIn posting phrased with the symbol first ("$231,000 - $340,000"), confirming the regex itself works — only the trailing-symbol phrasing is missed.

- `apps/desktop/src-tauri/src/extraction/salary.rs:28` (`CURRENCY`) and `:42-47` (`SALARY_RANGE_RE`) build the pattern as `{CURRENCY}\s*{NUMBER}\s*{SEPARATOR}\s*{CURRENCY}?\s*{NUMBER}` — the currency marker is REQUIRED immediately before the FIRST number (only the second occurrence is optional). A string where the currency symbol instead follows each number ("60.000 € – 75.000 €") never satisfies the leading `{CURRENCY}` before the first `{NUMBER}`, so the whole match fails and `extract_salary_range` (`salary.rs:94-96`) returns `None`.
- The UI has no distinct "salary present but unparsed" state — `salary.posting` being absent renders identically whether the posting truly has no salary or the extractor simply missed a real one, so there is no way for a user to tell these apart from the fit card alone.

## Suggested fix direction

Add a second pattern (or make the marker's position symmetric) that also matches a NUMBER immediately followed by the CURRENCY marker, so `NN.NNN €` phrasing matches on either side, while keeping the "always require a currency marker somewhere" discipline that already excludes bare number ranges / dates / percentages.

### LINKED ISSUE #1223: Check fit: missing-keyword chips can include non-job-content junk (nav/chart text), score shifted once on an unchanged posting
## Summary

"Check fit"'s "missing keywords" chips (the "why?" disclosure) can be populated with tokens that are clearly not job-posting vocabulary — URL path fragments and job-board chart-widget date labels — rather than genuine skills/requirements the résumé lacks. On the same, completely unchanged posting, the fit score itself also shifted once (27% → 21%, then stayed stable at 21% across 6 further repeated clicks), which is consistent with the extracted job text not being fully stable across capture attempts.

## Environment

Extension 0.154.0 (store build == repo main), desktop app 0.154.0, Windows 11, Chrome. LinkedIn job posting (`linkedin.com/jobs/view/...` and the equivalent job-search detail pane).

## Steps to reproduce

1. Open a LinkedIn job posting (one with a company "Hiring insights" panel — employee-growth chart, applicant-education/seniority stats — which LinkedIn shows on most postings).
2. Click "Check fit" immediately after the page/pane finishes loading, and expand "why?".
3. Click "Check fit" again a few more times on the same, unreloaded tab/posting.

## Expected

Missing-keyword chips should be job-relevant terms genuinely present in the posting text but absent from the résumé. The score should be identical across repeated calls on the same, unchanged posting and résumé.

## Actual

- Chips observed for one real posting (Harvey, "Staff Software Engineer, Frontend"): `/company/harvey`, `1sep`, `2025mar`, `2026sep`, `accommodations` — the first four are clearly not job-posting keywords (a link path fragment and what look like abbreviated chart axis date labels from the "Hiring insights" employee-growth chart on the same page), not skills or requirements text.
- The FIRST "Check fit" call (via the popup, right after the page/pane finished loading) scored this posting 27%. Every subsequent call on the exact same tab/posting/résumé (6 more clicks, via both popup and panel) consistently scored it 21% — a real, reproducible one-time shift, not click-to-click noise (it never moved again after the first call).

## Evidence

- Screenshots: expanded "why?" chip list showing `/company/harvey`, `1sep`, `2025mar`, `2026sep`, `accommodations`, immediately followed by the correct "Posting says $231,000 - $340,000" line (so the surrounding extraction IS finding the real posting — the junk is additive, not a wrong-page capture).
- Score sequence recorded across 7 consecutive "Check fit" clicks on the one tab (no reload between any of them): 27%, then 21%, 21%, 21%, 21%, 21%, 21%.

## Root cause (repo-relative path:line refs, best available from static reading — not fully confirmed against the LinkedIn DOM itself)

- `apps/desktop/src-tauri/src/commands/match_resume.rs:366-399` computes `job_keywords`/`gaps` directly from `job_text`, with no filtering beyond stopwords/stemming — whatever text arrives in `job_text` becomes candidate "missing keyword" material verbatim.
- `job_text` traces back to the `description` field `apps/desktop/src-tauri/src/scraping/scrape_url/mod.rs:466-548` (`parse_from_html`) assembles for the extension's Scan-mode capture, via JSON-LD → `data-ajh-job-root` hint → `readability_content_text`/`main_content_text` fallback (`mod.rs:540-548`). The specific chip content (a chart's date-axis labels, a nav-link path) is consistent with page chrome outside the actual job-description block leaking into whichever of those extraction paths ran for this capture — LinkedIn's SPA is a plausible case where the hint/JSON-LD may not scope as tightly as a fully server-rendered page.

## Suggested fix direction

Filter/validate candidate keyword tokens before they can surface as "missing keyword" chips — e.g. reject tokens that are pure digit/month-abbreviation runs or contain path-separator characters — as a defensive floor independent of fixing the upstream extraction scoping. Separately, worth double-checking whether `job_text` capture/extraction is fully deterministic for an unchanged LinkedIn posting, since the score itself moved once in this session.

_Found during an end-to-end test of the published 0.154.0 extension against the 0.154.0 desktop app. Lower confidence than the other findings from this track — flagging for the owning domain expert to confirm the exact leak point._

saeedkolivand:
Closes #1221, closes #1222, closes #1223.

## #1221 — the on-page badge went stale after an SPA job-to-job navigation

LinkedIn's job→job move is a `history.pushState`, which fires **no event in any world** — and the isolated world cannot observe the page world's own `pushState` calls. No event-based scheme can catch it, so the rendered badge now watches its captured url: an immediate first check, `popstate`/`hashchange` as same-tick accelerators, and a 1s poll of `location.href` as the only mechanism that never misses a pushState.

On a mismatch the badge is **invalidated**, never re-checked — a re-check would mean a desktop round trip for a posting the user already navigated away from. The watcher tears down on dismiss and on re-render, so it can never outlive its badge or hold a detached subtree.

## #1222 — German trailing-currency salary never detected

The pattern required the currency marker *before* the first number, so `"60.000 € – 75.000 €"` never matched. Adds a second, trailing-currency branch. A magnitude floor keeps that new branch from claiming bare number ranges (`"2024 – 2025"`, `"10.000 Schritte"`, `"3 € – 5 €"`).

The floor is scoped to the **new branch only**. Round 1 applied it to both, which silently dropped ordinary sub-10k *monthly* ranges whose period words `PERIOD_SUFFIX` doesn't recognise — `"£2,500 - £3,500 a month"` and `"€4.500 - €6.000 pro Monat"` both regressed from a match to `None`. That's now guarded by a dedicated no-regression test asserting the pre-change output byte-for-byte.

## #1223 — junk missing-keyword chips

Chips could carry scraped page chrome — url path crumbs (`/company/harvey`) and chart axis date labels (`1sep`, `2025mar`) — as if they were skills. The keyword kernel drops both shapes. The filter sits **after** synonym canonicalisation, so `ci/cd → cicd` and `c/c++ → cpp` survive; `es2015`, `oauth2` and `react17` survive because their cores aren't month abbreviations.

**Half B (the one-time 27% → 21% score shift) has no fix here, deliberately.** That path never invokes an LLM (`semantic_enabled` is hardcoded `0`), rounding can't move 6 points, and the cache is keyed on a job-text hash so it can't serve a stale score. Every step is a pure function of (résumé, captured job text), so the shift requires different *input* — LinkedIn lazy-renders panels, so a capture seconds earlier or later contains different text. That variable is in the page, not this repo, and the original captures aren't replayable. Adding a cache or a rounding stabiliser would only hide it.

## Verification

Every claim below was re-run independently of the implementer.

- `cargo test --lib salary`: 119 passed · `keywords`: 53 passed
- `pnpm vitest run` (apps/extension): **1067 passed** (35 files)
- `pnpm typecheck` clean · `pnpm lint:strict --max-warnings 0` clean

**Mutation checks, all run independently:**

| Guard removed | Result |
|---|---|
| badge watcher never armed | 5 fit-badge tests fail |
| floor widened back to both branches | `leading_currency_ranges_are_never_floored` fails |
| chip filters bypassed | both keyword tests fail |


saeedkolivand:

PR #1248, opened 2026-09-23
fix(extension): make the panel's trust and confirmation states honest

### LINKED ISSUE #1224: extension: Attach's first-time confirmation dialog is invisible when clicked from the Documents tab (silent hang)
## Summary
The Documents tab's "Attach résumé to this page" button uses the SAME first-time-per-host confirmation dialog as the Job tab's "Fill this form" (`mountFirstFillConfirm`), but that dialog is DOM-mounted inside the Job tab's `<section>`. Since the tab bar hides inactive tab panels with the `hidden` attribute, clicking Attach while the Documents tab is the active tab renders the confirmation dialog inside a `hidden` ancestor — fully built (correct copy, buttons, checkbox) but completely invisible and unclickable. The `await fillConfirm.confirm(...)` call the click handler is waiting on has no way to ever resolve, so the whole Attach action just hangs forever with zero feedback.

## Environment
Extension 0.154.0 (store build), desktop app 0.154.0, Windows 11, Chrome. Reproduced on a real Greenhouse application page (`job-boards.greenhouse.io`).

## Steps to reproduce
1. Open the side panel on a job page whose host has never been confirmed for Fill/Attach (`fillConfirmDontAskHosts` has no entry for it).
2. Switch to the Documents tab (never visit the Job tab first).
3. Click "Attach résumé to this page".

## Expected
Either the same "First Attach on `<host>`" confirmation dialog the Job tab shows, or at minimum a visible response of some kind.

## Actual
Nothing happens. The button just gets focus; no dialog, no status message, no console error, no network/bridge activity. It looks completely broken. Reproduced 2/2 tries. Switching to the Job tab afterward reveals the pending "FIRST ATTACH ON `<HOST>`" dialog that was silently waiting there the whole time — proving the click was accepted and is hung, not ignored.

## Evidence
`evaluate_script` on the side panel while the Documents tab was active:
```json
{"jobPanelHidden": true, "docPanelHidden": false, "jobPanelChildCount": 4,
 "jobPanelInnerHTMLTail": "...<div class=\"inset\"><p class=\"inset-label\">First Attach on <host></p><p class=\"inset-copy\">This will attach your résumé file — nothing is submitted.</p><div class=\"action-row\"><button class=\"btn btn--primary\" type=\"button\">Attach</button>..."}
```
The dialog markup exists, fully rendered with the right host and copy, inside `jobPanel` while `jobPanelHidden: true`.

## Root cause
- `apps/extension/src/sidepanel/sidepanel.ts:185-187`: `fillConfirmHost` is created and appended to `jobPanel` (the Job tab's section), and `mountFirstFillConfirm` is mounted there once.
- `apps/extension/src/sidepanel/sidepanel.ts:241`: the Documents tab's `confirmAttach` reuses that exact same `fillConfirm.confirm(...)` instance/DOM host — there is no separate mount for Documents.
- `apps/extension/src/tabs/tabs.ts:82-89` (`setActive`): `entry.panel.hidden = !active` — the Job tab's `<section>` gets `hidden = true` whenever any other tab is selected, and `hidden` cascades to all descendants regardless of their own `hidden` state, so the confirmation dialog is unreachable by the user.

## Suggested fix direction
Mount `fillConfirmHost` outside the tab bar's hidden `<section>` panels entirely (e.g. as a sibling overlay at the top level of the side panel), so it stays visible/interactive no matter which tab is currently selected.

_Found during an end-to-end test of the published 0.154.0 extension's Documents tab against the 0.154.0 desktop app._

### LINKED ISSUE #1225: extension: Documents tab shows permanent "Loading…" instead of the Job tab's "grant access" message when the page isn't trusted
## Summary
When the followed tab is not "trusted" (`isPageTrusted(state)` false — the same condition that makes the Job tab show its clear "Click the toolbar icon to grant access to this page, then use these tools." line), the side panel calls `documents.reset()` instead of `documents.refresh()`. `reset()` clears the candidate list and sets an EMPTY status text, and the Documents tab's render function falls back to the literal placeholder string `"Loading…"` whenever there are no candidates and no explicit status. The result: the Documents tab shows "Loading…" forever — not because anything is in flight, but because the page isn't trusted — with no error, no explanation, and no visible way to recover, while the Job tab (driven by the exact same state) shows a clear, actionable message for the identical condition.

## Environment

## Steps to reproduce
1. Open the side panel already following some tab.
2. Navigate that same tab to a different site (a plain in-tab navigation, no toolbar click).
3. Open/stay on the Documents tab.

## Expected
Either the doc list loads, or — matching the Job tab's behaviour for the identical untrusted state — a clear message telling the user to click the toolbar icon.

## Actual
The Documents tab is stuck on "Loading…" indefinitely (observed for 2+ minutes with zero backend activity and zero console output) while the Job tab, showing the SAME follow() state, correctly renders "Click the toolbar icon to grant access to this page, then use these tools." Reproduced on 2 separate hosts (an Ashby direct-application page and a happyhotel/Ashby iframe-embedded page) after an ordinary same-tab navigation — no rapid tab-cycling needed.

Confirmed the message channel itself is fine: calling the exact same request manually from the panel's own context resolves instantly —
```js
await chrome.runtime.sendMessage({kind:'documentsList'})
// => {"ok":true,"kind":"documentsList","result":{"ok":false,"resource":"documents","error":"url is required"},"url":""}
```
— so a real (if itself imperfect — see below) response exists in under a second; the Documents tab component just never gets a chance to render it because `refresh()` was never called for this state in the first place.

## Root cause
- `apps/extension/src/sidepanel/sidepanel.ts:403-417` (inside `follow()`'s `subscribeAnswerState` callback): `documents.refresh()` is only called at line 408, inside the `if (state && isPageTrusted(state))` branch. The `else` branch (lines 410-414) calls `documents.reset()` instead, which never triggers a fetch.
- `apps/extension/src/documents/documents.ts:557-565` (`reset()`): sets `setStatus('', 'muted')` — an empty status — then renders.
- `apps/extension/src/documents/documents.ts` (render, ~line 216): `host.append(el('p', 'msg msg--muted', statusText || 'Loading…'))` — an empty status text falls back to the literal string "Loading…", indistinguishable from a genuine in-flight fetch.

As a secondary, related issue: `runDocumentsList()` (`apps/extension/src/background.ts:795-799`) resolves the active tab's URL via `browser.tabs.query({active:true, currentWindow:true}).catch(() => '')`; in the panel context this can itself resolve to an empty string (`tab?.url` empty), producing the Rust side's `"url is required"` error shown above — but that response is moot here since the panel never calls `refresh()` for the untrusted state to begin with.

## Suggested fix direction
Give the Documents tab (and Prep, which has the identical `else` branch at sidepanel.ts:414) a distinct "page not trusted / click the toolbar icon" status instead of a silent `reset()`, mirroring the Job tab's gated line — e.g. have `reset()` accept an explicit reason/status string, or have `follow()` pass the untrusted state through so `documents.render(state)` can show it.

_Found during an end-to-end test of the published 0.154.0 extension's Documents tab against the 0.154.0 desktop app._

### LINKED ISSUE #1227: extension: popup's Fill has no first-time confirmation at all, unlike the identical side-panel action
## Summary
"Fill this form" exists in both the popup (`popup.html`) and the side panel's Job tab, calling the same underlying `fill` action — but only the side panel wires a `confirmFill` dependency into `mountJobTools`. The popup passes no `confirmFill` at all, so clicking Fill from the popup fires immediately with zero confirmation, on any host, even the very first time. The side panel shows the "First Fill on `<host>`" dialog for the identical action on the identical host.

## Environment

## Steps to reproduce
1. Pick a job application form on a host that has never been confirmed (`fillConfirmDontAskHosts` empty for it).
2. Trigger the extension's toolbar button to open the popup, and click "Fill this form".
3. Separately (fresh host), open the side panel's Job tab and click "Fill this form" there.

## Expected
Consistent trust gating for the same destructive-ish action (it writes into a live application form) regardless of which surface triggered it.

## Actual
- Popup: click → immediately "Filled 6 fields — review them on the page (name split is a guess — verify)." — no dialog shown at all.
- Panel, same host reset to unconfirmed: click → "FIRST FILL ON `<HOST>`" dialog appears first, requiring an explicit "Fill" click (or "Don't ask again") before anything is filled.

## Root cause
- `apps/extension/src/popup/popup.ts:169` — `mountJobTools(els.jobToolsHost, { send, hideSaveAnswers: true })` — no `confirmFill`.
- `apps/extension/src/sidepanel/sidepanel.ts:215-233` — `mountJobTools(jobToolsHost, { send, confirmFill: async () => { ... } })` — wires the dialog.

## Suggested fix direction
Either give the popup the same `confirmFill` wiring (with its own small inline confirm, since the popup has no persistent side-panel-style inset to reuse), or make the first-Fill trust decision live somewhere both surfaces read from before calling `fill`, so the guarantee "you'll be asked before your saved details are written into a form for the first time on this site" holds regardless of entry point.

### LINKED ISSUE #1234: Prep tab gets stuck on unlabeled "Loading…" (draft buttons still enabled) when the active tab isn't a readable page
## Summary

The side panel's **Prep** tab has three visually-identical-but-different states that all render as bare `Loading…`: (1) genuinely loading, (2) the page hasn't granted the extension `activeTab` access yet, and (3) the window's active tab is a non-content page (e.g. a pending `ajh://` deep-link tab). In cases 2 and 3 it never resolves on its own, the `Draft company brief` / `Draft salary answer` buttons stay visibly enabled the whole time, and there's no indication anything is wrong — unlike the **Job** tab, which correctly shows "Click the toolbar icon to grant access to this page" for case 2.

## Environment

extension 0.154.0 store build, desktop app 0.154.0, Windows 11, Chrome.

## Repro A — fresh navigation before activeTab is (re-)granted

1. Open the side panel, go to **Prep**.
2. `navigate_page`/type a new URL into the same tab (no toolbar click).
3. Prep tab: `Loading…` + both draft buttons enabled, indefinitely.
4. Switch to the **Job** tab on the same panel: correctly shows "Click the toolbar icon to grant access to this page, then use these tools."
5. Click the toolbar icon to grant access, switch back to Prep: now correctly shows `Nothing yet for this job.` + `Prepare in the app`.

## Repro B — "Prepare in the app" opens a pending tab that becomes active

1. On a job with no Prep content, click **Prepare in the app** (`browser.tabs.create({url:'ajh://prep?url=...'})`).
2. This opens a **new tab in the same window** for the `ajh://` external-protocol handoff, which becomes the window's active tab while Chrome's own external-app confirmation is pending.
3. Prep tab: reverts from `Nothing yet for this job.` back to bare `Loading…` (draft buttons still enabled) for as long as that pending tab is active.
4. Closing that pending tab (`chrome.tabs.remove`) and refocusing the original job tab self-heals Prep back to the correct empty state.

## Expected

Either a state-specific message (mirroring the Job tab's "click the toolbar icon…" message) or, at minimum, `Loading…` shouldn't be indistinguishable from "actually loading" when it's actually "no valid page to read right now" — and it shouldn't require the user to know to switch tabs/panel-tabs to recover.

## Actual

Bare, permanent `Loading…` with fully-clickable Draft buttons in both cases; self-heals only once the active tab becomes a valid job page again, with no hint to the user about what to do.

## Suspected cause

`prep/prep.ts`'s render logic (`renders statusText||'Loading…'`, per the existing code-map notes for this area) falls back to the generic `'Loading…'` whenever `statusText` is unset — which happens whenever `isPageTrusted(state)` never resolves to a definite trusted/untrusted value (state stays `null`) rather than explicitly reaching the "no page" case the Job tab already handles separately. `sidepanel/sidepanel.ts`'s population gate (`prep.refresh()`/`prep.reset()` around lines 405-414) is the likely place a "no valid active tab" branch is missing relative to what the Job tab's own component already does.

## Note re: automation vs. real behavior

In Repro B, the deep link itself never reached the desktop app during the automated test (no matching log line) — that part is very likely an automation artifact (Chrome's native external-protocol confirmation dialog needs a real user click, which an automated test can't perform) and is **not** part of this finding. The Prep-tab-freezes-while-that-pending-tab-is-active behavior is independent of whether the user ultimately accepts or declines that dialog, and would affect a real user identically either way.

saeedkolivand:
Closes #1224, closes #1225, closes #1227, closes #1234.

Four defects that share a shape: the surface knew the truth and showed the user nothing usable.

## #1224 — Attach's confirmation was invisible from the Documents tab

The confirm host was mounted inside the **Job tab's** `<section>`. The tab bar hides inactive panels with `hidden`, which cascades to descendants — so clicking Attach from the Documents tab built the dialog correctly, invisibly, under a hidden ancestor, and `await confirm(...)` could never resolve. The action hung forever with no feedback.

It now mounts into `#view-connected`, a sibling of the tab content, outside every `[data-section]`.

## #1225 + #1234 — permanent "Loading…" (one root cause, two issues)

`follow()` called `reset()` on an untrusted page, which set an **empty** status, and both render functions fell back to the literal `"Loading…"` whenever status was empty. Indistinguishable from a real fetch, forever — while the Job tab, reading the same state, showed an actionable line.

Both modules now carry a real in-flight flag, cleared in **every** terminal branch (reply, desktop refusal, kind mismatch, thrown error), so `"Loading…"` only ever means "in flight". Untrusted pages get the Job tab's existing `JOB_TOOLS_GATED_LINE` — one shared string, not three. Per #1234, Prep's Draft buttons are also disabled when there's no readable page; an enabled button that cannot work is part of the same lie.

## #1227 — the popup's Fill had no confirmation at all

The panel wired `confirmFill`; the popup passed none, so it wrote saved details into a live form with zero confirmation on any host.

**The first fix for this was rejected.** It fed the confirm from the pushed answer-state origin, which is `null` for any page the user hasn't scanned — and `confirm()` early-returns `true` on a null host. Probed:

```
expected { insetShown: false, filled: true }
      to deeply equal { insetShown: true, filled: false }
```

— a silent fill with no prompt, on the ordinary path. The Fill button isn't gated behind that state either; `job-tools.ts` deliberately leaves these buttons enabled.

It now resolves the origin **on demand at click time** via `tabs.query` (the same read `background.ts`'s `activeTabOriginAtGesture` already performs under the toolbar-click `activeTab` grant), re-validates after the dialog settles so an approval can't transfer to a page the user navigated to, and **refuses visibly** when the page can't be identified rather than treating unknown as approved.

Measured, not assumed: in a headless Chrome with the unpacked build, `tabs.query` without a grant returns `[{"id":273536669}]` — no `url`. So the refusal branch is a live path, not theoretical. The granted case could not be measured headlessly (`chrome.action` is absent there, so no toolbar click can be synthesised) and is **not** claimed as measured; production precedent is the existing `activeTabOriginAtGesture` read.

## Verification

Re-run independently of the implementer.

- `pnpm vitest run` (apps/extension): **1073 passed** (35 files)
- `pnpm typecheck` clean · `pnpm lint:strict --max-warnings 0` clean
- **Mutation check:** replacing the unknown-origin refusal with "approve" fails `refuses visibly — and never fills — when the tab url cannot be read`.

## Known, not fixed here

The side panel has the same `currentOrigin = state?.origin ?? null` shape and is briefly exposed **before its first state delivery**. Unlike the popup it cannot be fixed the same way — the panel follows arbitrary tabs and `activeTab` is gesture-scoped, so there's no grant to read a followed tab's url on demand. Deliberately not widened into this PR; filed separately.


saeedkolivand:

PR #1250, opened 2026-09-23
fix(extension): stop stale stamps, fences and settings reaching the user

### LINKED ISSUE #1220: Stamp results page: duplicate Saved/Applied pills accumulate on repeated Stamp clicks
## Summary

"Stamp this results page" places a "Saved"/"Applied" pill next to each matched job card by appending a new host element after the card's anchor. The de-dup guard that is supposed to remove a card's previous stamp before placing a new one lives in module-level `WeakMap`s inside `results-stamp.ts` — but that file is injected as a fresh classic script on every single "Stamp" click, so the guard's memory is empty every time even though the DOM nodes from earlier clicks are still sitting on the page. Result: every repeated "Stamp this results page" click on an unchanged results page adds one more duplicate pill to the same card instead of replacing the existing one.

## Environment

Extension 0.154.0 (store build == repo main), desktop app 0.154.0, Windows 11, Chrome.

## Steps to reproduce

1. Open a job board's listing page that has at least one card the desktop app already has tracked (saved/applied) — e.g. a Greenhouse company board (`job-boards.greenhouse.io/<company>`) with one already-imported posting.
2. Open the extension popup (or panel) and click "Stamp this results page". Confirm the status reads "Stamped 1 card." and exactly one "Saved ×" pill appears next to the tracked card.
3. Without reloading or navigating, click "Stamp this results page" again on the same page.

## Expected

The second click should leave exactly one pill on the card (replace, not add), since nothing about the page or the tracked posting changed between clicks.

## Actual

The second click adds a SECOND identical "Saved ×" pill next to the same card, both fully interactive (each with its own working dismiss "×"). Reproduced 2/2 tries in a row; each further click would add another.

## Evidence

- Screenshots: `apps/extension` evidence captured during this test — one card ("Engineering Manager — Premium Cards (Frontend)" on `job-boards.greenhouse.io/dkbcodefactory`) with exactly one "Saved ×" pill after the first Stamp click, and the same card with two "Saved ×" pills side by side after the second Stamp click on the same, unreloaded page. No other card on the board (including the adjacent, similarly-named "Premium Cards (Mobile)" card) ever gets a pill — targeting itself is correct, only the re-run dedup is broken.

- `apps/extension/src/lib/results-stamp.ts:64,69` — `stampNodes`/`stampShadows` are plain module-level `WeakMap`s, the ONLY bookkeeping `placeStamp` (`results-stamp.ts:137-182`) has for "did I already stamp this anchor".
- `apps/extension/src/lib/results-stamp.ts:143` (`placeStamp` calls `clearStamp(anchor)` first) and `results-stamp.ts:121-125` (`clearStamp` looks the anchor up in `stampNodes`/`stampShadows` before removing anything) — this only works if the maps still hold the entry from the PREVIOUS run.
- `apps/extension/src/background.ts:974` — `runStampResults` injects `results-stamp.js` via `browser.scripting.executeScript({ target: { tabId }, files: ['results-stamp.js'] })` on every single "Stamp" click. Chrome re-executes a classic script's top-level code on each injection, so `stampNodes`/`stampShadows` are re-declared as brand-new EMPTY maps every time — `clearStamp` can never find a stamp placed by an earlier injection, even though that stamp's actual DOM node (`results-stamp.ts:145-179`, `anchor.insertAdjacentElement('afterend', host)`) is still physically in the page.

## Suggested fix direction

Make the dedup check DOM-based instead of (or in addition to) memory-based: before inserting a new host, look for an existing `[data-ajh-stamp]` sibling right after the anchor (`anchor.nextElementSibling?.matches('[data-ajh-stamp="true"]')`) and remove/replace it, since that check survives a fresh script injection the same way the DOM itself does.

### LINKED ISSUE #1229: extension: side panel trust line renders the agent `job` resource's literal `<job_posting>` prompt-fence wrapper instead of stripping it for display
## Summary

The extension's side panel "Reading: …" trust line, and the `ajh-tauri agent job <url>` CLI output, both show the literal `<job_posting>…</job_posting>` prompt-injection-defense wrapper around a tracked job's `title`/`company` (and, on the CLI, `location`/`description`) instead of plain text. The wrapper is intentional and transient — applied only when a value crosses the agent/MCP read boundary (ADR-010) — but the extension's own human-facing UI reuses that exact boundary for its own display and never strips the markup back off, so a security artifact meant only for an LLM prompt leaks into text a person reads.

This does **not** indicate any corruption of the desktop's stored job data — see "What this is not," below.

## Environment

extension 0.154.0, desktop app 0.154.0, Windows 11, Chrome.

## Steps to reproduce

1. Have autopilot track at least one job (any board).
2. Open the extension's side panel on that job's own tab, or run `ajh-tauri agent job "<a tracked job's url>"` from the CLI.
3. Read the panel's trust line, or the CLI's `title`/`company`/`location`/`description` fields.

## Expected

The `job` agent-read resource's fence wrapper exists for an AI/MCP consumer building a model prompt (`apps/desktop/src-tauri/src/extension_bridge/agent_read.rs`'s own doc: "a consumer whose entire purpose is 'an AI agent reads this'"). A human-facing renderer — the side panel's trust line — should strip that wrapper (or the extension should have its own small unwrap helper) before display.

## Actual

Both the CLI reply and the side panel's trust line show the literal `<job_posting>`/`</job_posting>` tags wrapped around `title`/`company` (CLI also shows them on `location`/`description`).

## Root cause

- `apps/desktop/src-tauri/src/extension_bridge/agent_read.rs:605-611` (`fence_description`) and `apps/desktop/src-tauri/src/extension_bridge/agent_read/best_matches.rs:256-264` (`fence_posting_display_fields`) wrap `description` and `title`/`company`/`location` respectively, unconditionally, on every `job` resource reply — by design (ADR-010), and this never mutates the underlying store.
- `apps/extension/src/background.ts:700-706` (`readJobTitleCompany`) and `apps/extension/src/background.ts:716-723` (`runTrustLineJob`) call `agentQuery('job', {url})` and copy `title`/`company` out of the reply as-is, with no fence-stripping.
- `apps/extension/src/sidepanel/sidepanel.ts:302-312` (`refreshTrustLineJob`) then renders `Reading: ${res.title} · ${res.company}` with the wrapper markup still attached.
- `rg -n "job_posting" apps/extension/src` (excluding tests) returns no hits — the extension's TypeScript has no code anywhere that knows about or strips this fence-tag convention.

## Suggested fix direction

Add a small client-side unwrap helper in the extension (mirroring `prompt_fence::strip_fence_wrapper`'s exact-shape match: strip a `<job_posting>\n…\n</job_posting>` wrapper only when the string matches that precise shape) and apply it wherever the extension displays a `title`/`company`/`location`/`description` value read through `agent.query`'s `job` (and likely `best-matches`/`found-jobs`) resource, starting with `readJobTitleCompany` in `apps/extension/src/background.ts`.

## Note: what this is not

An earlier version of this issue attributed the same visible symptom to the desktop *persisting* fence tags into the found-jobs/autopilot store via a write path that supposedly lacked `unfence_job_field`'s protection (`apps/desktop/src-tauri/src/prompt_fence.rs:478`, applied in `apps/desktop/src-tauri/src/commands/scrape.rs:532-548`). That is not what happens: the only writer of `FoundJob`'s text fields, `build_found_job` (`apps/desktop/src-tauri/src/commands/autopilot.rs:900-918`), copies them straight from the raw scraped posting, and no agent-reachable write command exposes an individual found job's fields for write at all (`autopilot_update`'s schema, `packages/shared/src/schemas/index.ts:981-983`, only edits autopilot-level config). The wrapper is applied fresh, transiently, on every `job` resource read — exactly as designed — and the only real defect is that the extension's own display code doesn't undo it before showing the text to a person.

### LINKED ISSUE #1233: Side panel left open across an auto-tracked submit shows stale (pre-apply) status
## Summary

When the side panel is already open on a job's page *before* auto-track's submit-detection marks that job "applied" (a real form submit, or the fake-form technique used for E2E testing), the panel's **Job** tab keeps showing the old, pre-apply state (`Import this job` primary button, no stage tracker, `Answers` tab with no count) even well after the write has landed. A brand-new popup or a freshly-reopened panel shows the correct "Applied" state immediately — only the *already-open* surface is stale.

## Environment

extension 0.154.0 store build, desktop app 0.154.0, Windows 11, Chrome.

## Steps to reproduce

1. Import a fresh, untracked posting; keep the side panel open on that tab throughout.
2. Arm the watcher (`Check fit`) and submit the page's application form (a 3-field fake form was used for testing, but this is about the panel's own state, not the watcher itself).
3. Confirm the desktop side actually applied it: `ajh-tauri agent job "<url>"` → `"applied": true`.
4. Look at the still-open side panel's **Job** tab (no interaction other than switching its own internal tabs) — waited 17s+, tried switching Job/Answers/Job tabs internally.
5. Compare against a **freshly opened** popup/panel on the same tab.

## Expected

The already-open panel converges on the true state shortly after the write (it doesn't need to be instant, but it shouldn't stay wrong indefinitely with no user action available to fix it short of closing and reopening the panel).

## Actual

- Stale panel (open since before the submit): `Import this job` button, no stage tracker, `Answers` tab (no count) — indistinguishable from "never imported."
- Same tab, freshly triggered popup: `Applied Sep 21` status pill, correct.
- Same tab, side panel closed and reopened: `Applied Sep 21` + full `Saved → Applied → Interview → Offer` stage tracker, correct.

So the data is correct from the moment of the write; only the continuously-mounted panel component fails to pick it up. A user who leaves the panel open while filling out and submitting an application (the exact intended save-answers-on-submit + auto-track workflow) never sees confirmation that anything happened unless they close and reopen the panel.

## Suspected cause

The panel's Job-tab status likely isn't re-fetched on `submitDetected`/`status.update` success — only on its own mount (`start()`) or an explicit `getStatus`-triggering user gesture (per `connection-status.ts`). Auto-track's background flow (`background.ts`'s `handleSubmitDetected` → `lib/auto-track.ts`) doesn't appear to broadcast a refresh to already-open surfaces after a successful `updateStatus`. Worth checking whether the Job tab's data source subscribes to the same broadcast channel the Notification Center / badge nudge use, or only ever polls on mount.

### LINKED ISSUE #1236: Appearance settings (theme, default panel tab) don't live-sync to an already-open panel
## Summary

Two Appearance settings (Theme, Default panel tab) don't reliably reach a side panel that is already open in the same window — confirmed live, twice, with before/after screenshots. Neither is data loss and neither claims false success, but both are silent: nothing tells the user the change didn't apply to what's currently on screen.

## Environment

extension 0.154.0 store build, desktop app 0.154.0, Windows 11, Chrome. Found during an end-to-end test of the Settings page.

### Steps to reproduce

1. Open the side panel in a window (any tab).
2. In a separate Settings tab, change Theme from Dark to Light.

### Expected / Actual

Settings itself updates immediately (confirmed via screenshot: `Theme: Light` selected, cream background). The already-open panel stays on the old theme (screenshot: still the dark background) — no repaint. Closing and reopening the panel picks up the new theme correctly, so the setting itself is fine; only the live surface is stale.

### Root cause

`apps/extension/src/lib/theme.ts` — `bootTheme()` runs once per surface load; there's no `chrome.storage.onChanged` listener repainting an already-mounted surface.

### Steps to reproduce (directly reproduced)

1. Window A: open the panel (Default panel tab = Job at the time) — it opens on Job, as expected. Manually click over to the Answers tab.
2. Close the panel and reopen it in the **same** window A (Default panel tab setting unchanged).
3. Separately, Window B (freshly opened, panel never used in it before): open the panel — it opens on Job, matching the default. Change Default panel tab to Answers in Settings. Reopen the panel in the **same** window B.

### Expected / Actual

Step 2: reopening in window A shows **Answers** — the tab the user last clicked to, correctly remembered per-window across a close/reopen.

Step 3: reopening in window B shows **Answers** too, but here that's because it now matches the live default — window B never had an explicit tab switch, so it falls through to `getDefaultPanelTab()` and correctly reflects the change.

Not directly re-tested, but a direct reading of the code below: once a window's session memory holds a tab (from an explicit switch, as in window A), the fallback to `getDefaultPanelTab()` is skipped unconditionally — nothing re-checks whether a *later* Default-panel-tab change should override it. So a window like A, after a manual switch, would keep showing its remembered tab on reopen even if the default is changed afterward, until that window is closed. This half of the claim should be verified directly (switch tabs in a window, then change the default, then reopen in that same window) before being treated as confirmed rather than inferred.

### Root cause

`apps/extension/src/sidepanel/sidepanel.ts` — `loadActiveTab` only falls back to `getDefaultPanelTab()` when `storage.session` has no persisted tab for that window; a window where the user switched tabs at least once keeps its own memory until the window closes.

## Suggested fix direction

Add a `chrome.storage.onChanged` listener in the panel (and any other already-mounted surface) for the `theme` and `defaultPanelTab` keys, at minimum repainting the theme live; for the default-tab case, either clear the per-window session entry when the setting changes, or accept the current behavior as intentional ("remembers where you were") and say so in the Settings copy so it doesn't read as unconditional.

saeedkolivand:
Closes #1220, closes #1229, closes #1233, closes #1236.

Four surfaces that kept showing the user something the app already knew was wrong.

## #1220 — duplicate Saved/Applied pills

Dedup lived in module-level `WeakMap`s, but `results-stamp.js` is injected as a **fresh classic script on every Stamp click** — so its memory was empty each time while the previous click's pills were still in the DOM. Every repeat click appended another.

Dedup is now DOM-based: a bounded forward-sibling scan (max 8) from the anchor removing nodes matching the stamp's exact signature and stopping at the first non-match. That survives injection *and* heals pills already stacked from before the fix. The in-memory maps stay — they still earn their keep within a single injection — but they're no longer the only guard. A page node carrying the marker is only touched if it sits in that bounded run **and** matches the full signature.

## #1229 — the `<job_posting>` prompt fence leaked into the trust line

The fence is applied by design on every agent-resource read (ADR-010); the extension just never undid it before showing the text to a person.

New `fence-strip.ts` mirrors `prompt_fence::strip_fence_wrapper` exactly — including the subtlety that the suffix is tested against the **remainder**, so the empty-body overlap `<tag>\n</tag>` correctly does *not* match. No greedy tag-stripping, so a title legitimately containing angle brackets passes through byte-for-byte. Rust side untouched.

## #1233 — an already-open panel showed stale pre-apply status

Auto-track's write was correct from the moment it landed; nothing told a continuously-mounted surface to re-read.

Reuses the **existing** `runtime.sendMessage` broadcast channel (the one `broadcastStatus` already uses) with a new `jobStatusChanged` push, fired only on a confirmed `result.ok`. The panel's single module-scope listener refreshes only when the url's origin matches the followed page. No polling, no timers; a malformed url is dropped rather than compared.

## #1236 — Appearance settings didn't reach an open panel

**Theme:** `bootTheme()` ran once per surface load. A `storage.onChanged` listener — filtered to area `local` and key `theme` — now repaints an open surface, treating removal as `system`, and returns an unsubscribe for teardown.

**Default panel tab:** the issue flagged this half as inferred. It reproduces, and is worse than described: `storage.session` **outlives window close** and window ids are reused, so a dead window's remembered tab could mask a newer default indefinitely. A `windows.onRemoved` listener clears that window's key. Precedence is preserved — the user's explicit choice still wins for the rest of that window's life; only the cross-window masking goes away.

## Verification

Re-run independently of the implementer.

- `pnpm vitest run` (apps/extension): **1121 passed** (36 files)
- `tsc --noEmit` run **directly**, bypassing the turbo cache: clean
- `pnpm lint:strict --max-warnings 0`: clean

**Mutation checks, run independently:**

| Guard removed | Result |
|---|---|
| `jobStatusChanged` origin filter made unconditional | origin-scoping + malformed-url tests fail |
| `isStampHost` stubbed to `false` | all three #1220 dedup tests fail |

The merge with main resolves cleanly by construction: both sides only **added** tests to `popup.test.ts` and `sidepanel.test.ts` (zero deleted lines from the base on either side), so both sets are kept and both were verified present and passing afterwards.


saeedkolivand:

PR #1251, opened 2026-09-23
fix(extension): keep accepted answers writable, honour instructions, stop cancelled spend

### LINKED ISSUE #1230: extension: Accept into field works once per row, then fails and orphans the draft on the next Rescan
## Summary

In the extension's Answers tab, "Accept into field" only works ONCE per row. After the first successful accept fills a previously-empty field, every subsequent Accept on that same row (e.g. after clicking Regenerate to produce a new version) fails immediately with "Could not find this field — the page may have changed." — even though the field is untouched, still visible, and holds exactly the text the first Accept wrote. If the user then clicks Rescan to recover, the real-world field is split into TWO panel rows: a fresh one with no draft history, and an orphaned one that keeps the AI-drafted versions but permanently loses its "Accept into field" button, stranding that work.

## Environment

extension 0.154.0 store build, desktop app 0.154.0, Windows 11, Chrome, Ollama Cloud (`gpt-oss:20b`) as the AI provider.

## Steps to reproduce

1. Open an application form with an empty text/textarea question (confirmed on Ashby, e.g. jobs.ashbyhq.com/linear/.../application, "What country are you based in?").
2. Open the side panel → Answers tab → expand that row → "Draft this answer" → "Accept into field". Confirm the page field now shows the drafted text.
3. Click "Regenerate" on the same row (produces a new version, v2/v3/…).
4. Click "Accept into field" again, without Rescanning in between.

## Expected

The new version replaces the old text in the (now non-empty) field, the same way "Accept" worked the first time.

## Actual

Step 4 fails instantly: `"Could not find this field — the page may have changed."` The field itself is confirmed present and unchanged (verified by reading the live DOM value directly: still holds exactly what the first Accept wrote). Reproduced 1/1 cleanly; the message is actively misleading since nothing about the page changed.

If the user then clicks **Rescan** to try to recover: the Answers tab count inflates (4 questions → 5 in my repro) and the SAME real field now shows as two separate rows — one labelled "…Filled" with no version history (Accept works fine on this one), and the original row (still showing its full v1..vN draft history) with NO "Accept into field" button at all — only Copy/Regenerate remain, so the user's edited draft can never be written to the page again without manually copy-pasting it themselves.

## Root cause

- `writeRowText` (`apps/extension/src/background.ts:1954-1988`) branches on `row.field.kind` ('empty' vs 'filled') to pick `injectAnswerFill` (empty-field path) vs `injectAnswerReplace` (filled-field path). On a successful write, its state update (lines 1979-1986) only refreshes `field.currentText` — it never flips `field.kind` from `'empty'` to `'filled'`.
- So a second Accept on the same row still takes the `injectAnswerFill` branch, which calls `locateQuestionField` (`apps/extension/src/lib/answers-capture.ts:247-256`) — this only searches `emptyCandidateFields`. Since the field is no longer empty, it is absent from that candidate set, the occurrence count no longer matches, and the function returns `null` → the fixed `NOT_FOUND` result (`apps/extension/src/lib/answer-fill.ts:44-47`), surfaced verbatim as "Could not find this field — the page may have changed."
- The Rescan-time row-split is a direct consequence: `fieldRowId(kind, question, index)` (`apps/extension/src/lib/answer-state.ts:197-199`) bakes `kind` into the row's own id. A fresh scan of the now-genuinely-filled field produces a `filled:…` id, which is different from the original `empty:…` id the row was created under, so `buildRows`'s carry-over-by-id logic (`answer-state.ts:254-332`) cannot reunite them — it creates a brand-new "Filled" row for the live field and leaves the old row to survive only because it "holds versions" (per that function's own doc comment), now with `field` undefined and therefore no Accept action (`canAccept`, gating `apps/extension/src/answer-tools/answer-tools.ts:459`).

## Suggested fix direction

When `writeRowText` succeeds against an `'empty'`-kind field, also flip `field.kind` to `'filled'` in the same state update (alongside `currentText`) so a second Accept correctly takes the `injectAnswerReplace` path without needing a Rescan. Separately, `buildRows`'s carry-over might consider matching a prior row by `(question, index)` ignoring `kind` when the exact-id match misses (or the field ref could migrate to `filled:...`'s row rather than orphaning the original), so a Rescan doesn't strand drafted versions were the kind-transition to happen anyway.

### LINKED ISSUE #1231: extension: Answers-tab free-text instruction is ignored by chips and dropped server-side on Regenerate
## Summary

The Answers tab's free-text "Instruction for '<question>'" box has no effect on the drafted/rewritten text, in either of the two ways a user could plausibly use it: clicking a TONE/LENGTH chip while it's filled in ignores it completely (chips carry their own hardcoded preset/instruction), and using "Regenerate"/"Draft this answer" — the one control that actually transmits the typed instruction to the desktop app — also has no effect, because the desktop's draft-mode compose path never reads the `instruction` field at all.

## Environment

extension 0.154.0 store build, desktop app 0.154.0, Windows 11, Chrome, Ollama Cloud (`gpt-oss:20b`).

## Steps to reproduce

1. Open the Answers tab on a drafted row with an existing version (e.g. "Tell us more").
2. Type a specific instruction into "Instruction for '…'", e.g. "Focus only on the websocket migration story, keep it under 120 characters total."
3a. Click a TONE or LENGTH chip (e.g. "Shorter"), OR
3b. Click "Regenerate" (the button that is documented — via the UI hint "Regenerate rethinks it from your résumé" — to be the one that uses free text).

## Expected

At least path (b) — Regenerate — incorporates the typed instruction (topic focus and/or the 120-character limit) into the new version, since it is the control that sends `instruction` in its request payload.

## Actual

Neither path reflects the instruction at all:
- (a) Chip click: result is close in length to the pre-chip text and still about React/Next.js/Tauri/etc — nothing about the requested "websocket migration" topic.
- (b) Regenerate click: result (v6, ~650 characters) still restates unrelated highlights from the résumé — no mention of "websocket" anywhere, and nowhere close to the requested 120-character limit. The desktop AI request log confirms a real, successful call was made (`req=21`, `status=200`, `duration=5663ms`) — this is not a silent failure, the instruction is simply never reaching the model.

## Root cause

- **Chips never read the free-text box.** `apps/extension/src/answer-tools/answer-tools.ts:165-190` — every `TONE_CHIPS`/`LENGTH_CHIPS` entry carries its OWN fixed `preset` or `instruction` string (e.g. `Warmer` → `"Make this warmer and more personal…"`). The click handler (`answer-tools.ts:392-401`) sends only `chip.preset`/`chip.instruction` — the view-local `instructions` map (what the user typed) is never referenced there at all.
- **Regenerate DOES send it, but the backend drops it.** `answer-tools.ts:488-503`, the Regenerate/"Draft this answer" handler, sends `mode: 'draft'` plus `instruction: instructions.get(row.id)` when non-empty (line 500) — this part is correct. But server-side, `apps/desktop/src-tauri/src/extension_bridge/answer_assist.rs`:
  - `parse_instruction`/`resolve_rewrite_instruction` (lines 388-417) are documented as "rewrite mode only" and are only invoked via `validate_rewrite_fields`, which is itself only called `match mode { AssistMode::Rewrite => Some(validate_rewrite_fields(payload)?), AssistMode::Draft => None }` (lines 720-723).
  - `build_user_message`, the draft-mode prompt builder (lines 551-592), takes `question, resume, job_description, company_brief, web_notes, salary_range` — there is no parameter for a free-text instruction anywhere in its signature or body.
  - So a `mode:'draft'` request's `instruction` field is parsed nowhere in the whole compose path; it is present on the wire and then simply never read.

## Suggested fix direction

Either wire the free-text instruction into the draft-mode prompt (extend `build_user_message` with an optional "additional instruction from the candidate" parameter, fenced the same way `question` is, and have `resolve_answer_assist`'s `AssistMode::Draft` arm read `payload.get("instruction")` and pass it through), or — if draft mode is intentionally never meant to take a free-text instruction — remove/hide the instruction box for the draft/Regenerate action (or gray it out with a note) so it isn't presented as something that shapes the next draft when it silently does nothing. The current combination (a visible, always-enabled input, a UI hint implying Regenerate is exactly the control that uses it, and total silence about the fact that nothing happens) is the misleading part.

### LINKED ISSUE #1232: Prep tab's Cancel does not actually stop the in-flight AI request (still billed in full)
## Summary

Clicking **Cancel** on a Prep-tab on-demand draft (`Draft company brief` / `Draft salary answer`) correctly reverts the extension UI back to the draft button, but the underlying AI-provider request is **not aborted** — it runs to completion server-side and consumes real tokens/cost, exactly as if Cancel had never been clicked.

## Environment

extension 0.154.0 store build, desktop app 0.154.0, Windows 11, Chrome, provider `ollama-cloud` / `gpt-oss:20b`.

## Steps to reproduce

1. Open the side panel's **Prep** tab on any job with no existing company-brief/salary-answer content.
2. Click **Draft salary answer**.
3. Immediately click **Cancel** (verified via script: the Cancel button is clicked at the first opportunity, effectively 0ms after the draft starts, well before the ~3.5s the request actually takes).
4. Compare the extension's own click timestamp against the desktop app log (`AI Job Hunter.log`, UTC).

## Expected

Cancelling stops the outbound AI request; no further tokens are spent for that draft once Cancel is clicked.

## Actual

The app log shows the request completing in full, at the same duration/token cost as an un-cancelled draft:

```
[21:32:34] [ai] → req=29 provider=ollama-cloud model=gpt-oss:20b endpoint=/chat/completions streaming=true
[21:32:37] [ai] stream start job=job-6a6f0064-...
[21:32:37] [ai] stream end provider=ollama-cloud model=gpt-oss:20b answerLen=112 thinkingLen=43 finishReason=Some(End) usageIn=2413 usageOut=43 empty=false
[21:32:37] [ai] ← req=29 ... status=200 duration=3467ms ok=true
```

The extension UI reverted cleanly to `Draft salary answer` (no stray result, no error) — the bug is invisible from the UI; it only shows up as wasted spend in the app log.

## Suspected cause

`apps/extension/src/lib/bridge.ts`'s `cancelCurrent()` → `supersedeAnyPendingAssist()` does send an `assist.cancel` frame (`cancelAssist()`, `apps/extension/src/lib/bridge.ts:2014-2022`) and correctly stops relaying further chunks/settles the promise client-side — but nothing downstream on the desktop side appears to abort the actual provider HTTP stream for this code path. `apps/desktop/src-tauri/src/extension_bridge/answer_assist_topic.rs` (the Prep tab's on-demand `topic` extension to `answer.assist`, PR4) has no references to `cancel`/`register`/`AbortHandle`/`spawn` at all, despite its own doc comment claiming it "grounds it through the EXACT SAME draft pipeline every other `answer.assist` call uses — same gate, same `ai_research` bucket/limiter, **same registry**, same streaming frames." The real cancellable-job registry lives in `apps/desktop/src-tauri/src/extension_bridge/assist_registry.rs` (`cancel`/`cancel_all`, backed by a `JobCanceller`/`AbortHandle`) — worth confirming whether a topic-mode draft's job actually gets `register`ed into it, or whether it silently bypasses registration.

## Impact

"AI spend is real money" — a user who reflexively cancels a draft (wrong topic clicked, changed their mind, etc.) is still billed for the full generation every time, with zero indication anything went wrong.

saeedkolivand:
Closes #1230, closes #1231, closes #1232.

## #1230 — "Accept into field" worked exactly once per row

A successful write into an empty field never flipped `field.kind`, so the second Accept still took the fill path — whose locator only searches **empty** candidates. It couldn't find the field it had just filled and returned the fixed NOT_FOUND, claiming the page had changed when the field was sitting right there holding our own text. A Rescan then split the field into two rows and stranded the drafted versions on the one with no Accept button.

The kind now flips on a **confirmed** write (`result.filled` guards it, so a refused write still takes the fill path next time), and `buildRows` migrates drafts by `(question, index)` onto the row that owns the field, with a `claimed` set so one scan can't adopt the same prior row twice.

## #1231 — the free-text instruction was collected and discarded

Two separate drops: chips carried their own hardcoded directive and never read the typed box, and the draft compose path never read the `instruction` field at all — it was on the wire and parsed nowhere.

Both now reach the model. The instruction is **user-typed and untrusted**, so it is byte-capped at 500 (mirroring the existing rewrite-mode sibling), fenced into its own `<candidate_instruction>` block, and — the part that matters — **the tag is registered in `FENCE_TAG_PATTERNS`**, so a crafted `question` or résumé cannot forge a sibling instruction block. The shared forged-boundary test grew from 6 cases to 7.

## #1232 — Cancel reverted the UI but still paid for the request

**The issue's stated root cause was wrong**, and verifying it first is what found the real one. Topic mode does *not* bypass the registry: every `answer.assist` frame goes through `spawn_answer_assist` → `begin_or_reject_duplicate`.

The actual leak: three grounding steps run **before** `compose_draft_stream`'s own `start_and_register`, and only the **company-brief** one checked for a cancel that raced ahead. The salary-market lookup and the web-search notes did not — and the report's repro was "Draft salary answer", which is exactly the unguarded path. Cancel stopped the compose while the grounding call was already billed in full.

All three now share one named guard, `abort_if_cancelled_early`, tested directly against a real registry.

## Verification

- `cargo test --lib`: **5939 passed**, 0 failed (full suite, not filtered)
- `cargo fmt --check` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean
- `pnpm vitest run` (apps/extension): 1085 passed
- `tsc --noEmit` run directly (no turbo cache); `pnpm lint:strict` clean

**Mutation checks:**

| Guard removed | Result |
|---|---|
| `kind: 'filled'` flip | new background test fails — falls back to `answer-fill.js` |
| `candidate_instruction` unregistered | cross-tag forgery test fails |
| `abort_if_cancelled_early` body | cancelled-generation test fails |

## Two things the full suite caught that a filtered run did not

- `every_registered_fence_tag_is_load_bearing` is a deliberate drift guard holding an independent copy of the fence vocabulary. Adding a tag without listing it fails loudly — by design. Now listed, so the shared loop probes the new tag with all three forgery shapes.
- R8's LOC cap: the additions pushed `answer_assist.rs` to 1471 against a hard 1400. Rather than take an allowlist exception, the pure `AppHandle`-free half (clamping, parsers, the mode enum, rewrite validation) is split into `answer_assist_parse.rs`. 1471 → 1282, no behaviour change.


saeedkolivand:

PR #1252, opened 2026-09-23
fix(extension): stop import claiming success it did not earn

### LINKED ISSUE #1238: extension: Import silently captures the embedding page for iframe-embedded ATS boards
## Summary

When a job board is embedded on a company's own careers page via a cross-origin `<iframe>` (a common pattern for Ashby, and potentially other ATS platforms with embeddable widgets), clicking **Import** on the wrapper page silently imports the WRAPPER page's own title/heading instead of the job posting inside the iframe. There is no error, no partial-capture warning — only a very low match-fit percentage in the success toast, which is easy to miss or not recognize as a signal that anything went wrong.

## Environment

- Extension 0.154.0 (store build)
- Desktop app 0.154.0
- Windows 11, Chrome

## Steps to reproduce

1. Open a careers page that embeds an ATS job board (e.g. Ashby) in a cross-origin `<iframe>`, with the specific job selected via a query parameter in the outer page's URL.
2. Click the extension's toolbar icon to open the popup on that tab.
3. Click **Import this job**.

## Expected

Either the real job posting inside the iframe is captured and imported, or the user is clearly told the capture may not reflect the actual posting (e.g. the existing "couldn't read the description — open AI Job Hunter to paste it" partial-import warning), so they know to check/complete the record manually.

## Actual

The import succeeds with a confident-looking message ("Imported \"<wrapper page's own heading>\". Open AI Job Hunter → Applications to view it.") built from the OUTER page's own `<title>`/`<h1>`, not the job. The only anomaly visible to the user is a very low percent-fit suffix on that same message — there is no explicit warning, and the resulting Application record has the wrong title/company and no real job description. Frequency: deterministic, every time, for any cross-origin-iframe-embedded ATS whose wrapper page happens to render a non-empty title/heading (not timing-dependent).

## Evidence

Reproduced against a live Ashby-embedded careers page; the popup's success message named the wrapper page's own "Careers at <company>"-style heading rather than the actual job title, at a single-digit percent fit. (Sanitized — no personal data; the specific company/job-posting URL used for the repro is omitted here as it is not needed to understand or fix the defect.)

## Root cause

- `apps/extension/src/background.ts:474-477` (`captureTabHtml`) injects the DOM-capture script via `browser.scripting.executeScript({ target: { tabId }, files: ['content.js'] })` with no `allFrames` option, so `content.ts`'s capture (`document.documentElement.outerHTML`) can only ever see the top-level document — a cross-origin iframe's content is unreachable without `host_permissions` this extension does not request. (The README already documents this exact no-allFrames/activeTab-only trade-off for the Fill feature — `apps/extension/README.md`'s "Known limit — assisted autofill fills the TOP-LEVEL document only" note — but not for Import.)
- `apps/desktop/src-tauri/src/scraping/scrape_url/mod.rs:466-548` (`parse_from_html`) falls through to the whole-document `<title>`/first-`<h1>` fallback when the extension's job-root hint finds nothing in the outer page (which it never can here), picking up the wrapper page's own heading.
- `apps/desktop/src-tauri/src/scraping/scrape_url/mod.rs:129-191` (`canonical_job_url`) and `apps/desktop/src-tauri/src/scraping/ats_ref.rs:131-138` (`ashby_slug`) only recognize a URL whose OWN host is the ATS's host — nothing detects/resolves an embedded ATS iframe on an unrelated host.
- `apps/desktop/src-tauri/src/extension_bridge/import_flow.rs:103-105` (`usable()`) treats ANY non-empty title as usable, so the wrapper page's own heading (never empty) suppresses the existing "couldn't read the description" partial-import warning (`apps/extension/src/job-tools/job-tools.ts:80-127`) that would otherwise flag this case.

## Suggested fix direction

Either direction addresses the silent-wrong-data problem without requiring new cross-origin permissions:

1. **Cheap, in line with the existing Fill precedent**: extend the README's documented Fill limitation to Import, and have the extension/desktop side detect "no job-root hint was found in the outer document AND the page contains a same-page ATS iframe" (a `document.querySelector('iframe[src*="ashbyhq.com"], iframe[src*="greenhouse.io"], ...')` check in `content.ts` is enough signal) to surface the iframe's own URL to the user (e.g. "This job is embedded — open it directly at <iframe src> to import").
2. **Fuller fix**: teach `usable()`/`import_flow.rs` that a title sourced only from the whole-document fallback (never from JSON-LD, `__NEXT_DATA__`, or the extension's own job-root hint) is a weaker signal, and route that case through the existing `partial` warning path instead of treating it as a clean success.

### LINKED ISSUE #1239: desktop: LinkedIn import via DOM capture keeps site-branded title and falls back to hostname as company
## Summary
When the extension imports a LinkedIn job posting using its DOM-capture ("Scan mode") path, the desktop backend's generic HTML fallback produces a title still carrying the page's site-branding suffix and a company name that is just the URL hostname, even though the real employer name is present elsewhere in the captured markup.

## Environment

## Steps to reproduce
1. As a logged-in LinkedIn user, open a job view page (`https://www.linkedin.com/jobs/view/<id>`).
2. Trigger the extension popup or side panel and click Import.

## Expected
- Title: the job title alone (e.g. "Javascript Developer").
- Company: the real employer name (e.g. "Digital Waffle").

## Actual
- Title is saved as the whole `<title>` tag verbatim, including site branding, e.g. `"Javascript Developer | Digital Waffle | LinkedIn"`.
- Company is saved as the URL host, e.g. `"www.linkedin.com"`.
- DOM capture itself succeeded (the stored job description was ~8,000 characters of real content, including the text "Company logo for, Digital Waffle" from an ``) — the extraction heuristic just never looks at that signal.
- Frequency: reproduced on the one LinkedIn posting tested; deterministic given the code path (not timing-dependent).

## Evidence
Imported application record: `title` = `"Javascript Developer | Digital Waffle | LinkedIn"`, `company` = `"www.linkedin.com"`, with the correct job URL and a full-length job description containing the real company name as incidental text (not extracted as a field).

## Root cause
Scan-mode Import (`apps/desktop/src-tauri/src/extension_bridge/import_flow.rs:252`) parses the extension's captured DOM directly via `scraping::scrape_url::parse_from_html`, bypassing the dedicated LinkedIn board handler (which needs the desktop's own authenticated session and isn't needed when the DOM was already captured from the owner's logged-in browser). That generic path has two gaps:
- `parse_generic_html` (`apps/desktop/src-tauri/src/scraping/scrape_url/mod.rs:918-931`, called at `mod.rs:468`) takes the first `<title>`/`<h1>` element's full text with no trimming of a trailing " | SiteName" suffix.
- `parse_generic_company` (`apps/desktop/src-tauri/src/scraping/scrape_url/mod.rs:939-969`) only checks JSON-LD `hiringOrganization.name` and `meta[property="og:site_name"]`; neither is populated by LinkedIn's authenticated, client-rendered job page. Its caller (`mod.rs:549`) then falls back to the bare URL host (`parse_generic_company(html).unwrap_or(host)`).

## Suggested fix direction
In the generic fallback: strip a trailing `" | <site>"`/`" - <site>"` suffix from the title when it matches a known site name or `og:site_name`; and add a company fallback that looks for a distinctive `Company logo for, X`-style pattern (or similar DOM signal already present in LinkedIn's captured markup) before giving up and using the host.

saeedkolivand:
Closes #1238, closes #1239.

Two ways a capture produced a confident-looking record that described the wrong thing.

## #1238 — an iframe-embedded ATS board imported the wrapper page

A company careers page that embeds an ATS board in a **cross-origin `<iframe>`** is unreachable from the extension's top-level-only capture — deliberately, since reaching it would need `host_permissions` for every ATS origin, far beyond this extension's `activeTab` scope. The generic parser then fell back to the wrapper's own `<title>`/`<h1>`, and `usable()` (a non-empty title) was happily true, so the user got *"Imported \"Careers at X\""* with the only hint being a single-digit percent-fit.

**Not fixed by widening permissions.** Fixed by honesty: the import is flagged `partial` — reusing the existing "couldn't read the description, paste it" path rather than adding a third message — when **both** of these hold:

1. no description was extracted, and
2. the captured document embeds a cross-origin frame.

Both halves are required on purpose. Analytics, consent and video frames are cross-origin on nearly every page, so signal 2 alone would fire constantly; signal 1 is the discriminator, and a page that parsed a real description is never flagged.

The existing no-`allFrames` note in `apps/extension/README.md` now covers Import as well as Fill.

## #1239 — LinkedIn scan-mode kept site branding and stored the hostname as the employer

`"Javascript Developer | Digital Waffle | LinkedIn"` as the title, `"www.linkedin.com"` as the company.

**Title:** one trailing `" | <site>"` / `" - <site>"` suffix is dropped, and **only when that segment actually matches** the `og:site_name` or the host's own label. A blanket "drop everything after the last separator" would mangle `"Engineer | Payments"`, so the negative cases are tested explicitly — an internal pipe, an internal dash, and a title that *is* the site name (never stripped to empty).

**Company:** after JSON-LD and `og:site_name` come up empty, a named heuristic reads a `Company logo for, <name>`-style ``. It's documented as a heuristic, it's consulted last, and it returns `None` on any page without that shape so the host fallback still works exactly as before.

Both extracted fields are byte-capped before they can reach a stored Application row — the markup is attacker-controlled.

## Verification

- `cargo test --lib`: **5939 passed**, 0 failed (full suite)
- `cargo fmt --check` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean

**Mutation checks:**

| Guard removed | Result |
|---|---|
| title suffix strip | end-to-end LinkedIn-shaped test fails |
| logo-alt company heuristic | 2 tests fail |
| cross-origin comparison inverted | both iframe tests fail |

## Module split

The additions pushed `scrape_url/mod.rs` to 1534 against R8's hard cap of 1400. The generic-extraction half moves to `scrape_url/generic.rs` — the natural seam, since nothing in it makes a request or knows any board's shape. 1534 → 1251, no behaviour change.


saeedkolivand:

PR #1254, opened 2026-09-23
fix(extension): attach to the résumé field when a cover letter sits beside it

### LINKED ISSUE #1228: extension: Attach always reports "more than one résumé field" on Greenhouse-style forms with a separate cover-letter upload
## Summary
`isResumeFileInput`'s fallback rule matches a file input as "résumé-like" when its `accept` attribute lists document types (`pdf|\.docx?|msword|wordprocessingml`), with no requirement that the name/id/label also mention résumé/CV. On a very common real-world ATS pattern — a separate, optional "Cover Letter" file upload sitting next to the "Resume/CV" upload, both accepting the same document types — this fallback matches BOTH inputs, so Attach always reports "Found more than one résumé upload field on this page — attach it yourself," even though the correct field is trivially identifiable by its id/name alone.

## Environment
Extension 0.154.0 (store build), desktop app 0.154.0, Windows 11, Chrome. Reproduced on a real Greenhouse application form (`job-boards.greenhouse.io`).

## Steps to reproduce
1. Open a Greenhouse job application page that has both a résumé upload and a separate cover-letter upload.
2. Open the side panel's Documents tab, pick a résumé source, and click "Attach résumé to this page".

## Expected
Attach locates the résumé-specific field and attaches to it (its `id="resume"` is unambiguous).

## Actual
`"Found more than one résumé upload field on this page — attach it yourself."` — reproduced 3/3 tries (deterministic, not flaky). DOM inspection of the two file inputs:
```json
[{"id":"resume","accept":".pdf,.doc,.docx,.txt,.rtf"},
 {"id":"cover_letter","accept":".pdf,.doc,.docx,.txt,.rtf"}]
```
Both inputs share the exact same `accept` list, so both match `RESUME_FILE_ACCEPT_RE`, and neither `id` needed to say "resume" for the false match — the `cover_letter` field matched purely on `accept`.

## Root cause
`apps/extension/src/lib/submit-watch.ts:110-131`:
```ts
const RESUME_FILE_TEXT_RE = /resume|curriculum|lebenslauf|(?:^|[^a-z])cv/;
const RESUME_FILE_ACCEPT_RE = /pdf|\.docx?|msword|wordprocessingml/;
...
return RESUME_FILE_TEXT_RE.test(textSignal(el)) || RESUME_FILE_ACCEPT_RE.test(accept);
```
The two signals are OR'd with no way for one to disambiguate the other, and the `accept`-only match is too generic — "accepts a PDF or Word doc" is true of nearly every document upload field on a job application form, résumé or not.

## Suggested fix direction
When the accept-based match alone would produce more than one candidate on the page, prefer any candidate whose name/id/label/aria-label matches `RESUME_FILE_TEXT_RE`, and/or explicitly exclude a field whose text signal matches a cover-letter pattern (`cover.?letter|anschreiben|motivation`) even if its `accept` list matches.

_Found during an end-to-end test of the published 0.154.0 extension's Documents tab against the 0.154.0 desktop app._

saeedkolivand:
Closes #1228.

## Problem

Greenhouse ships both uploads with byte-identical accept lists:

```json
[{"id":"resume","accept":".pdf,.doc,.docx,.txt,.rtf"},
 {"id":"cover_letter","accept":".pdf,.doc,.docx,.txt,.rtf"}]
```

`isResumeFileInput` matches on name/label **or** accept list, so the accept-only signal matched both — and Attach refused with *"Found more than one résumé upload field on this page — attach it yourself"* on a page where the correct field is labelled `resume`. Deterministic, 3/3 in the report.

## Why the fix is in the caller, not the shared predicate

`isResumeFileInput` has two consumers that want **opposite safe directions**:

- `submit-watch.ts`'s `looksLikeApplicationForm` only asks "does this form take a document upload". A cover-letter field answers that just as well as a résumé one, so its permissiveness is correct — tightening the shared rule would make the watcher miss real application forms it currently detects.
- `attach-file.ts` has to pick exactly **one** field, and that's where the ambiguity hurts.

So the narrowing is a separate `isNamedResumeFileInput` applied only by the caller that needs it. The permissive predicate is unchanged.

## Still fails closed

The tie is broken **only** when exactly one candidate names itself a résumé. Two résumé-named fields, or none, refuse exactly as before — it never guesses which of several was meant, and never silently attaches to whichever came first.

## Verification

- `pnpm vitest run` (apps/extension): **1129 passed** (36 files)
- `tsc --noEmit` clean; `pnpm lint:strict --max-warnings 0` clean
- **Mutation check:** disabling the narrowing fails the new cover-letter test.

Three tests: the résumé field wins (and the cover-letter field is asserted to have received nothing), several résumé-named fields still refuse, and no-named-candidate still refuses.


saeedkolivand:

PR #1255, opened 2026-09-23
fix(extension): never treat an unnamed page as consent to fill

### LINKED ISSUE #1249: extension: side panel's Fill/Attach can skip the first-time confirmation before the first state delivery
## Summary

The side panel's `currentOrigin` is fed only from `state?.origin ?? null` in `follow()`'s `subscribeAnswerState` callback. Before the **first** async state delivery, `currentOrigin` is `null`, `hostOf(null)` is `null`, and `mountFirstFillConfirm`'s `confirm()` early-returns `true` on a null host — so a Fill (or Attach) clicked in that window writes to the page with no first-time confirmation shown.

This is the same defect class as #1227, which was fixed for the popup. It is **not** fixed for the panel, and deliberately so — see below.

## Why the popup's fix doesn't transfer

The popup now resolves the origin on demand via `tabs.query` at click time, which works because opening the popup **is** the toolbar-click gesture that grants `activeTab`.

The panel follows arbitrary tabs via `tabs.onActivated`/`onUpdated`. `activeTab` is gesture-scoped, so for a followed tab the user never clicked there is no grant and no readable url. The panel needs a different design.

## Scope of the exposure

Narrower than the popup's was:

- `job-tools.ts` starts `let trusted = true` and only flips on a delivered `render(state)`, and `follow()` calls `jobTools.render(state)` only inside the subscription callback — so the controls are clickable while connected and before the first delivery.
- **After** the first delivery the window closes: an untrusted/absent state makes `isPageTrusted` false and the gated line replaces the controls, and any *trusted* state is an `AnswerState` whose `origin` is a required string (`answer-state.ts`, `isAnswerState` checks `typeof s.origin === 'string'`). So a reachable panel Fill implies a non-null origin.

So this is a pre-first-delivery race, not a permanent absence.

## Suggested fix direction

Treat a null origin as **untrusted** rather than as "no host to confirm against" — i.e. the panel's controls should not be actionable until an origin is known, mirroring the popup's new fail-safe. Worth checking whether `mountFirstFillConfirm`'s `if (!siteHost) return true;` early return should also become a refusal, but that needs an audit of every caller first: it exists for callers that legitimately have no host.

_Found while reviewing the fix for #1227; reported rather than widened into that PR._

saeedkolivand:
Closes #1249.

## Problem

The side panel's `currentOrigin` is fed **only** by `follow()`'s answer-state subscription, so it is `null` until the first push lands. `mountFirstFillConfirm`'s `confirm()` resolved `true` for a null host — which turned *"we cannot name this page"* into *"approved"*. A Fill or Attach clicked in that window wrote into the page with **no dialog shown at all**.

This is the same defect class as #1227, found while reviewing that fix and filed rather than silently widened into it.

## Fix — two latches, deliberately

**The primitive now fails closed.** `confirm(null)` returns `false`. That is the right default for a consent gate and fixes both surfaces at the root. I audited every caller first: there are exactly three call sites (one in the popup, two in the panel), and all three want the refusal.

**The panel also refuses before it calls.** Not redundancy for its own sake — a consent gate should not depend on a primitive that two surfaces share staying fail-closed on its behalf. It also makes the panel's own intent legible at the call site.

The popup is unaffected: since #1227 it resolves its origin on demand and refuses *visibly* before it ever reaches the shared confirmation.

## Why not the popup's approach

The panel follows arbitrary tabs via `tabs.onActivated`/`onUpdated`, and `activeTab` is gesture-scoped — there is no grant to read a followed tab's url on demand. So the panel cannot resolve an origin it was never handed; it can only refuse to act without one.

## Verification

- `pnpm vitest run` (apps/extension): **1133 passed** (36 files)
- `tsc --noEmit` clean; `pnpm lint:strict --max-warnings 0` clean

**Mutation checks, run separately per latch:**

| Guard removed | Result |
|---|---|
| panel Fill latch + panel Attach latch | both new panel tests fail |
| primitive's unknown-host refusal | both new site-memory tests fail |

The panel tests establish the precondition explicitly (follow a fresh tab, deliver `null`) rather than assuming it, because module state persists across tests in that file — my first draft passed vacuously for exactly that reason. A third test pins the other direction: with a real origin the host still resolves normally, so the guard has not quietly become "never confirm".

Note the panel suite **mocks** `site-memory` with an always-true `confirm`, so the panel-level latch is what its tests actually exercise — the primitive's behaviour is pinned by its own unit tests.


saeedkolivand:

PR #1256, opened 2026-09-23
fix(extension): recognise the embedded board, not just any third-party frame

saeedkolivand:
Follow-up to #1238. **Found by end-to-end testing against a real careers page — the original fix never fired.**

## What the live run showed

Driving the real extension against a live company careers page that embeds an Ashby board:

```
cross-origin frames: ["https://jobs.ashbyhq.com/happyhotel?embed=js",
                      "https://consentcdn.cookiebot.com/sdk/bc-v4.min.html"]
import result: "Imported “Karriere bei happyhotel”. Open AI Job Hunter → Applications
                to view it. — 6% fit."
```

That is exactly the reported symptom, unchanged: the **wrapper page's own heading** imported as a confident success, the only hint being a 6% fit.

## Why the first fix missed it

It required **both** "no description was extracted" **and** "the page embeds a cross-origin frame". A careers page has plenty of prose, so a description *was* extracted — just not the job's — and the gate never closed.

The unit test passed because I wrote its fixture to match that assumption (an empty-bodied page) rather than a real careers page. A test written from the same wrong premise as the code cannot catch the code being wrong.

## The fix

Key on **recognising the board**, via the existing `ats_ref` registry (Greenhouse, Lever, Ashby, SmartRecruiters, Personio, Workable, Recruitee, Breezy, BambooHR, Pinpoint, Rippling), and drop the description condition entirely.

Recognising the board is what actually separates a wrapper careers page from the analytics, consent and video frames that are cross-origin on nearly every page — which was the job the description condition was hired for. It wasn't earning its place; it was suppressing the finding.

A frame pointing at the page's own host is still ignored, so an ATS page that embeds itself is never flagged.

## Tests, rebuilt from the real shape

- the **actual** captured frame set (Ashby board + Cookiebot consent frame) → flagged
- Greenhouse, Lever, SmartRecruiters embeds → flagged
- Cookiebot, YouTube, GTM, a relative `src`, `about:blank`, no frames at all → **not** flagged
- an Ashby page embedding its own host → not flagged
- unparseable page url → **warns**, deliberately: a needless "couldn't read the description" costs a second look, while staying silent writes a wrong record

**Mutation check:** replacing the registry lookup with "any cross-origin frame counts" fails the ordinary-frames test.

## Verification

- `cargo test --lib`: **5950 passed**, 0 failed


saeedkolivand:

PR #1257, opened 2026-09-23
fix(extension): abandon a cancelled grounding step instead of paying for it

saeedkolivand:
Follow-up to #1232. **Found by end-to-end testing — Cancel still billed.**

## What the live run showed

```
cancel sent at +260ms -> {"ok":true,"kind":"assistCancel"}
[ai] → req=4 /api/web_search      ... ok=true (847ms)
[ai] → req=5 /chat/completions    ... ok=true (2584ms)   ← STARTED after the cancel
```

Two provider calls completed after the cancel was acknowledged, and the second one *began* afterwards.

## Why the first fix wasn't enough

It added a **pre-flight peek** before each grounding step. That only fires when the cancel beats the step's *start*. Each grounding helper makes several provider calls internally — the salary lookup does a web search and then a completion — so a cancel landing mid-step passed the peek and paid for everything after it.

## The fix

Await the grounding future against the cancel marker and **drop it** when one lands. Dropping an in-flight `reqwest` future is what actually cancels the HTTP request; no flag check can do that once the call is out.

Deliberately contained in `answer_assist.rs` rather than threading a cancellation signal through `salary_research` and `commands::ai` — those are shared by callers that have nothing to do with the extension's cancel, and widening them for this would be the wrong trade.

The pre-flight peeks stay: they still avoid starting a step that is already doomed, which is cheaper than starting and abandoning one.

## Verification

- `cargo test --lib`: **5952 passed**, 0 failed

Two new tests: one asserts the value still comes back when no cancel arrives, the other asserts a long step is abandoned **and** that its future never ran to completion (an atomic flag it would have set).

**Mutation check:** making the poll never abandon fails the abandon test.


saeedkolivand:

PR #1264, opened 2026-09-23
fix(settings): stop reporting a sleeping extension as not connected

### LINKED ISSUE #1258: settings: Browser extension shows "Not connected" whenever the MV3 service worker is asleep
## Summary

Settings → Browser extension almost always reads **"Not connected"**, even when the extension is paired and demonstrably working. The indicator reports whether an authenticated WebSocket is **live right now**, but an MV3 service worker sleeps after ~30s idle and drops that socket. So the steady state for a perfectly healthy pairing is "Not connected".

## Environment

Extension 0.154.0, desktop app 0.154.0, Windows 11, Chrome 153.

## Steps to reproduce

1. Pair the extension. Use it (import a job, check fit) — it works.
2. Leave the browser idle ~30s so the MV3 worker unloads.
3. Look at Settings → Browser extension.

## Expected

Something that reflects reality: the pairing is fine and the extension reconnects on demand. "Not connected" reads like a fault the user should fix, and the Regenerate-token button sits right below it, inviting exactly the wrong action.

## Actual

`× Not connected`, essentially permanently. Observed while the extension was actively driving the app through the bridge all session.

## Root cause

- `commands/extension_bridge.rs::extension_bridge_status` returns `connected: state.is_connected()`.
- `extension_bridge/mod.rs::is_connected` is `self.connected.load(..) > 0` — a count of **live authenticated sockets**, incremented in `inc_connected` on reaching `Authenticated` and decremented in `dec_connected` on teardown.
- An MV3 service worker is evicted when idle, closing the socket, so the count returns to 0 within seconds of the last use. Nothing is wrong; there is simply no socket while nothing is happening.

Confirmed during an end-to-end run: with the worker evicted the service-worker target is absent and the count is 0, while the same extension answered every request immediately when woken.

## Suggested fix direction

Report **last-seen**, not just live. Track the timestamp of the most recent authenticated connection alongside the count, return it from `extension_bridge_status`, and have the UI distinguish three states rather than two:

- a live socket → "Connected"
- no socket but authenticated recently → paired/idle, which is the normal MV3 steady state
- never authenticated, or not for a long time → "Not connected"

The token/Regenerate affordance should only read as remedial in the third case.

_Found during the end-to-end verification round for the #1213–#1239 fixes._

saeedkolivand:
Closes #1258.

## Problem

Settings → Browser extension read **"Not connected"** essentially permanently, even while the extension was actively driving the app.

`is_connected()` counts **live authenticated sockets**. Chrome evicts an MV3 service worker after ~30s idle and the socket goes with it, so the steady state for a *healthy* pairing is zero sockets. The indicator was measuring the wrong thing for an MV3 client — and the **Regenerate token** button sits directly beneath it, inviting the one action that would actually break the pairing.

Confirmed during the end-to-end run: with the worker evicted the service-worker target is absent and the count is 0, while the same extension answered every request instantly once woken.

## Fix

Record when a socket last reached `Authenticated` and report it alongside the live count, so the surface can distinguish **three** states instead of two:

| State | Meaning |
|---|---|
| live socket | Connected |
| no socket, seen recently | **Paired** — the normal MV3 idle state |
| never seen | Not connected |

Only the last reads as a fault.

The count itself is unchanged — it is still correct for what it measures, and `EXTENSION_BRIDGE_CHANGED`, which rides its 0↔1 transitions, is untouched.

## Notes

- The timestamp is wall-clock and used **only** for this display. Never for ordering or security, so a clock adjustment can at worst make it look stale — never unsafe. Said so at the definition.
- The contract now documents that `connected` alone is a poor health signal, so the next reader doesn't repeat the mistake.
- Both locales updated; `docs/API.md` regenerated from the contract rather than hand-edited.

## Verification

- `pnpm -F @ajh/desktop test`: **4356 passed** (325 files)
- `cargo test --lib`: **5952 passed**, 0 failed
- `pnpm typecheck` clean · `pnpm lint:strict --max-warnings 0` clean · `cargo clippy -- -D warnings` clean

Two new tests: idle-but-recent shows "Paired" and **not** "Not connected"; never-paired still shows "Not connected" and not "Paired".

**Mutation check:** forcing the idle state off fails the idle test.


saeedkolivand:

PR #1266, opened 2026-09-23
ci(charts): summarise the repo-charts run as a table

saeedkolivand:
The 📈 Repo Charts run's numbers were only readable by opening the step log. Now the job summary carries them.

## What it renders

Live output from an actual run (local, hence no store credentials — which is exactly the case worth showing):

### 📈 Repo charts — 2026-09-23

Badge: **installs 936**

| Source | Count |
| --- | ---: |
| GitHub installers | 904 |
| Microsoft Store | _unavailable_ |
| Snap Store | _unavailable_ |
| Chrome Web Store | 30 |
| Firefox AMO | 2 |
| **Total** | **936** |

| Chart | This run |
| --- | --- |
| `downloads.svg` | rebuilt — 32 daily points (2026-07-21 → 2026-09-23) |
| `stars.svg` | rebuilt — 41 dated points from 57 stargazers |

…plus one line from the publish step: either the short SHA pushed to `badges`, or "nothing to do — the published tree is byte-identical".

## Two choices worth flagging

**Written by the build script, not a separate summary step.** The script already holds every value; re-reading `badge-out/` would create a second place to keep in step the next time an output changes — which is the exact failure mode the publish step's own comment warns about (`downloads-by-platform.json` shipped missing once that way).

**An unreachable store renders `_unavailable_`, never `0`.** `collectStoreCounts` deliberately preserves `null` through to `store-counts.json` so an outage stays visible. Flattening it to a zero in the one place people actually look would undo that, and would quietly under-report the total as if a store had no users.

## Verification

- Ran the real script with `GITHUB_STEP_SUMMARY` set — the table above is its genuine output, not a mock-up.
- Ran it again **without** the variable and confirmed no file is written, so a local run is unaffected.
- Workflow YAML parses; all 7 steps intact.
- `pnpm lint:strict --max-warnings 0` clean.


saeedkolivand:
🎉 This is included in version 0.155.1.

PR #1268, opened 2026-09-23
fix(extension): run the pairing handshake when a token is saved on an open socket

### LINKED ISSUE #1267: extension: first pairing shows connected but never authenticates until the socket drops
## Observed (store builds, app 0.155.0)
Pasting the pairing token into the extension: extension pill says **● Connected**, desktop Settings → Browser extension says **Not connected** — and Settings' Refresh does not change it. Only after a wait (MV3 worker idles, socket drops, reconnect) does the desktop flip to connected.

## Cause
1. Before a token exists, `attach()` opens the socket and sets phase `'connected'` with NO handshake (`apps/extension/src/lib/bridge.ts`, the `if (!token)` branch after `onClose` wiring).
2. `setToken` (`apps/extension/src/background.ts`) stores the token, calls `resetForNewToken()` + `ensureConnected()` — but `ensureConnected` returns early on `isOpen()`, so the v2 handshake never runs on the already-open, unauthenticated socket.
3. `computeStatus()` maps `bridge.phase === 'connected' && hasToken` → `connected`, though `phase === 'connected'` is documented as reachable without auth (see the `authenticated` field's doc).
4. Desktop only counts a socket after `AuthOk`, so it correctly reports not connected.

## Fix direction
- After a new token is saved, drop an open-but-unauthenticated transport so the next connect performs the handshake.
- `computeStatus` must key `connected` on the authenticated session, not the raw phase.

saeedkolivand:
Fixes #1267

## Problem
On a first pairing the popup said **Connected** while the desktop said **Not connected** (Refresh didn't help) until the MV3 worker idled and the socket reconnected. A socket opened with no token reaches phase `connected` without a handshake; saving a token left that socket open and unauthenticated (`ensureConnected` no-ops on `isOpen()`), and `computeStatus` treated `connected + hasToken` as authenticated.

## Fix
- `BridgeClient.resetForNewToken()` drops an open **unauthenticated** transport so the next `ensureConnected()` runs the v2 handshake with the new token. Authenticated sessions are untouched.
- `attach()`'s `onClose` ignores closes from a transport that is no longer current (same identity check `performHandshake` uses) — a replaced socket can't null the new one, flip to `app_not_running`, or arm a reconnect.
- `BridgeStatus` gains `authenticated`; `computeStatus` reports `connected` only when it is true (otherwise `searching`).

## Tests
- 3 bridge tests (re-handshake on saved token, stale close is a no-op, authenticated session not dropped) + `computeStatus` folding test.
- Mutation-checked: removing each of the three changes independently fails its test.
- Extension typecheck, `vitest run` (1138 passed), eslint `--max-warnings 0` all green.


saeedkolivand:
🎉 This is included in version 0.155.1.

